Issue #5, opened 2018-10-04
[Feature Request] Logging


sco1:
Logging events and errors while an application is running is extremely useful for tasks like diagnosing issues and analyzing usage patterns. 

Python's [`logging` library](https://docs.python.org/3/library/logging.html) provides a great built-in tool to enable application-level logging. Python's documentation also offers a [comprehensive series of tutorials](https://docs.python.org/3/howto/logging.html#logging-basic-tutorial) aimed at users with varying levels of experience.

The proposed feature set includes:

- [ ] Log to a discrete `/log/` folder, whose contents are not tracked by Git
- [ ] Single `logger` configuration and initialization
- [ ] Log bot events (e.g. login)
- [ ] Log events in all cogs and helper utilities at the appropriate level (`info`, `debug`, etc.)
- [ ] Log errors & warnings

ninexball:
I am interested in doing this

heavysaturn:
it's all yours!

heavysaturn:
Solved in #22

PR #22, opened 2018-10-06
Logging Feature

ninexball:
This is what I have so far thanks
Issue: #5


heavysaturn:
this is a good start. I love the use of rotatingfilehandler. maybe you could add some actual logging in the bot as well? it'd be useful to log whenever we load in a cog, and the new userinfo thing cog we merged earlier could probably use a few log events.

heavysaturn:
Excellent! I'm gonna merge this.

Issue #9, opened 2018-10-05
[Feature Request] Random scary movie generator


jodth07:
The goal of this tasks is to access the OMDB API, and find movies that are scary / spooky, and make them searchable/accessible via the bot.
- [ ] generate a random movie tag scary and/or spooky
- [ ] should contain, title, main actors and image, and year
- [ ] should also link to OMDB
- [ ] Perhaps also a spookiness/scary rating

akashsara:
Can I take this?

rgables:
I'd like to take a crack at this

jodth07:
have at it @rgables

heavysaturn:
solved in #23

PR #23, opened 2018-10-07
Added cog for random scary movie

rgables:
For issue #9: Random scary movie generator.

Currently picks a random "Halloween" movie.  We still need to figure out some kind of list of scary movies to randomly select from.


jodth07:
I am not sure if [omdb](http://www.omdbapi.com/) have by genre in their api.
[OMDB](https://www.themoviedb.org/documentation/api?language=en-US) seems to though

heavysaturn:
I think that for the purposes of this hacktoberfest, retrieving a random halloween movie will be good enough.

bot/cogs/movie.py
```diff
@@ -0,0 +1,65 @@
...
+
+        await ctx.send(movie_details)
+
+    @staticmethod
+    async def select_movie():
+        """
+        Selects a random movie and returns a json of movie details from omdb
+        """
+
+        # TODO: Come up w/ a scary movie list to select from. Currently returns random Halloween movie
```
heavysaturn: This line can be removed - a random halloween movie will be fine.

bot/cogs/movie.py
```diff
@@ -0,0 +1,65 @@
...
+        omdb_params = {
+            'apikey': OMDB_API_KEY,
+            'i': selection
+        }
+        response = requests.get('http://www.omdbapi.com/', omdb_params)
+
+        return response.json()
+
+    @staticmethod
+    async def format_metadata(movie):
```
heavysaturn: I'd like to see this method return an embed instead.

On our PyDis bot, we have a snake movie command, which builds an embed like this with the data it gets from OMDB:

image

Would you be able to do something like that?

rgables: Yeah, that sounds do-able.  Thanks for the feedback.

rgables:
> I am not sure if [omdb](http://www.omdbapi.com/) have by genre in their api.
> [OMDB](https://www.themoviedb.org/documentation/api?language=en-US) seems to though

I wasn't aware of themoviedb, but their api is way fancier.  Thanks for the tip

rgables:
Switched from omdb to tmdb's api for the advanced search filters.  !movie also now embeds the movie poster + info instead of just linking to an outside website.
screen shot 2018-10-09 at 1 05 32 am

heavysaturn:
That looks good!

Issue #17, opened 2018-10-05
[Feature Request] Bot configuration


GhostofGoes:
Being able to configure the bot without having to modify the source code would enhance ease of use, and enable users to easily update the bot without worrying about their settings being overwritten. Having the settings in a single location improves discoverability and makes changes to the configuration quicker and easier.

Discussion items
* File format to use: INI, JSON, YAML, or other
* Should logging should be configurable as well?
* If the configuration file is modified at run-time, should the bot's current configuration be updated (e.g Django).

Proposed features:
* [ ] Loading of configuration from a file
* [ ] Mechanism to override default values
* [ ] Set of default configuration values
* [ ] Changing the bot to use the values pulled from this file
* [ ] Ability to specify the location of the configuration file via a command line argument

Primary use case is configuration of cogs, such as the update rate for querying an API, roles/permissions for an administration cog, location of a file with terms to use for a quote generator, etc.

Some other settings:
* Command message prefix (e.g. `!`)
* name of environment variable to use for the token

If you have any others please sound off in the comments. I'm not terribly experienced with Discord bots, so my knowledge of typical examples is rather limited.

sco1:
An example of the types of things we're looking to be configurable would be useful.

GhostofGoes:
@sco1 Updated with some examples

xedre:
I'd like to work on this if approved

heavysaturn:
Let's wait a little bit with this one until some of the others have been merged, so we actually have some stuff to configure first. But it does seem reasonable apart from that.

heavysaturn:
yep!

xedre:
Okay working on it

sco1:
Being tackled now as part of #64

PR #25, opened 2018-10-07
Make use of constants.py

LewdNeko:
This PR changes it so that all constants previously found in the main `bot.py` are now found in the correct place - `constants.py`. It also adds support for setting the bot's prefix through the environment var `HACKTOBERBOT_PREFIX`, but will default to `.`, `👻 `, 👻 if it is not found.

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -1,14 +1,10 @@
-from os import environ
 from pathlib import Path
 from sys import stderr
 from traceback import print_exc
 
 from discord.ext import commands
 
-HACKTOBERBOT_TOKEN = environ.get('HACKTOBERBOT_TOKEN')
-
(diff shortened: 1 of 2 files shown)


heavysaturn:
we will not be accepting any of these five minute PRs to make minor changes that there are no staff created issues for. a `constants.py` should be implemented as part of a larger PR when and if the problem that is being solved calls for it. as this is a temporary project that will only run on a single server, there is no reason to believe we will ever change those prefixes.

LewdNeko:
I've already completed hacktoberfest, if you're concerned about me just using your repo as a farm. I'm really more interested in making good contributions to your bot :) Oh well.

sco1:
This will probably fit in with #17 once there are a bit more things to configure with the bot.

PR #34, opened 2018-10-09
Implement Halloweenify feature.

markylon:
I scraped some images and nicknames from the web and implemented the halloweenify feature.

### DIFF
--- a/bot/cogs/halloweenify.py
+++ b/bot/cogs/halloweenify.py
@@ -35,7 +35,7 @@ class Halloweenify:
         embed.description = (
             f'**{ctx.author.display_name}** wasn\'t spooky enough for you? That\'s understandable, '
             f'{ctx.author.display_name} isn\'t scary at all! Let me think of something better. Hmm... I got it!\n Your '
-            f'new nickname will be: \n :jack_o_lantern: **{nickname}** :ghost:'
+            f'new nickname will be: \n :ghost: **{nickname}** :jack_o_lantern:'
         )
         embed.set_image(url=image)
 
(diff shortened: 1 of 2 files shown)


heavysaturn:
this looks like it does what it should do - but the grammar needs some work. `wasn't spooky enough for you that you have decided to change it?` is not a grammatically correct way to put that.

my suggestion would be something like

**Lucy** wasn't spooky enough for you? That's understandable, **Lucy** isn't scary at all! Let me think of something better. Hmm... I got it!

Your new nickname will be: 
:ghost: **Hannibal Lecter** :jack_o_lantern:

markylon:
I agree, couldn't think of a fine message ;/ Good to merge?

bot/cogs/halloweenify.py
```diff
@@ -33,8 +33,9 @@ def __init__(self, bot):
         embed.colour = discord.Colour.dark_orange()
         embed.title = 'Wow!'
         embed.description = (
-            f'Your previous nickname, **{ctx.author.display_name}**, wasn\'t spooky enough for you that you have '
-            f'decided to change it?! Okay, your new nickname will be **{nickname}**.\n\n'
+            f'**{ctx.author.display_name}** wasn\'t spooky enough for you? That\'s understandable, '
+            f'{ctx.author.display_name} isn\'t scary at all! Let me think of something better. Hmm... I got it!\n Your '
+            f'new nickname will be: \n 👻 **{nickname}** 🎃'
```
heavysaturn: It's best not to put actual emojis in the code, as some systems will have trouble running the code when you do that. Instead, it'd be better to use `:jack_o_lantern:` or `:ghost:`

bot/cogs/halloweenify.py
```diff
@@ -33,8 +33,9 @@ def __init__(self, bot):
         embed.colour = discord.Colour.dark_orange()
         embed.title = 'Wow!'
```
heavysaturn: The title could probably be improved a bit, too. Maybe in staying with the theme below, it could be something like `Not spooky enough?`

PR #45, opened 2018-10-12
Add regex & command prefix filters to reaction cog

sco1:
- Add a command prefix check to prevent commands from firing the spooky reaction cog
- Translate the spooky reaction check into a regular expression
  - Prevents matching of substrings (e.g. `"halloween"` matching `"halloweenify"`
  - Allows for flexible matching of certain reactions (e.g. `"spoooooky"` fires the `"spooky"` reaction)

### DIFF
--- a/bot/cogs/spookyreact.py
+++ b/bot/cogs/spookyreact.py
@@ -1,11 +1,13 @@
+import re
+
 SPOOKY_TRIGGERS = {
-    'spooky': "\U0001F47B",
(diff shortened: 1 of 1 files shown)


jos-b:
At the moment the bot is responding to its own messages, so if a command response contains a trigger word it drops a reaction on it. This seems like a fitting PR to address that in.

sco1:
Short-circuit added

PR #48, opened 2018-10-12
Vote monster refactor

dfitzpatrick:
Hi all. Here is our PR for adding functions to vote for your favorite monster.
image

### Commands
`.monster`
Display a help menu

`.monster vote`    
Display list of monsters to vote for.

`.monster leaderboard`    
Get the leaderboard rankings of each monster.

`.monster show _monster_`
Get datailed information about a monster.

`.monster vote _monster_`
Vote for a particular monster


bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    Vote for your favorite monster!
+    This command allows users to vote for their favorite listed monster.
+    Users may change their vote, but only their current vote will be counted.
+    """
+
+    def __init__(self, bot: Bot):
+        """Initializes values for the bot to use within the voting commands."""
+        self.bot = bot
+        self.registry_location = os.path.join(os.getcwd(), 'resources', 'monstersurvey', 'monstersurvey.json')
```
sco1: Needs to use the correct path:

`os.path.join(os.getcwd(), 'bot', 'resources', 'monstersurvey', 'monstersurvey.json')`

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,161 @@
+from discord.ext.commands import Bot, Context
+from asyncio import sleep as asleep
+from typing import Optional
+from discord.ext import commands
+from discord import Embed
+import logging
+import json
+import os
+
+log = logging.getLogger(__name__)
```
sco1: Nothing is done with this, please add some appropriate logging statements to the rest of the cog.

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,161 @@
...
+            inline=False
+        )
+        vote_embed.set_thumbnail(url=self.voter_registry[name]['image'])
+        self.json_write()
+        return await ctx.send(embed=vote_embed)
+
+    @monster_group.command(
+        name='show'
+    )
+    async def monster_show(self, ctx: Context, name: str):
```
sco1: Need a default argument for `name`, otherwise an exception will be thrown if no monster name is provided.

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Shows the named monster. If one is not named, it sends the default voting embed instead.
+        :param ctx:
+        :param name:
+        :return:
+        """
+        m = self.voter_registry.get(name)
```
sco1: A more robust search needs to be made here:

- Current search is case sensitive. e.g. `.ms show Dracula` doesn't match anything
- It's not intuitive or clear to the user how to show a specific monster until the blank vote embed is shown. Because you're matching on keys here, monster names don't align with the `full_name` being shown on the leaderboard. e.g. `.ms show Count Dracula` doesn't match anything.
- Might be worth also accepting an integer input to show the corresponding monster from the leaderboard

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,161 @@
...
+                color=0xFF6800,
+                description='Vote for your favorite monster!'
+            )
+            default_embed.add_field(
+                name='.monster show monster_name(optional)',
+                value='Show a specific monster. If none is listed, show a brief of all.',
+                inline=False)
+            default_embed.add_field(
+                name='.monster vote monster_name',
+                value='Vote for a specific monster. You can vote more than once, but you can only vote for one monster'
```
sco1: This behavior does not align with the feature request.

Users should only be able to vote for a single monster, but are able to change their vote at any time.

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if ctx.invoked_subcommand is None:
+            default_embed = Embed(
+                title='Monster Voting',
+                color=0xFF6800,
+                description='Vote for your favorite monster!'
+            )
+            default_embed.add_field(
+                name='.monster show monster_name(optional)',
+                value='Show a specific monster. If none is listed, show a brief of all.',
```
sco1: The alternative behavior described in this help statement doesn't happen. If no monster is provided, all the user is shown is the blank voting embed.

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,161 @@
+from discord.ext.commands import Bot, Context
```
sco1: Please sort out the imports according to PEP8: https://www.python.org/dev/peps/pep-0008/#imports

3rd party libraries should be after built-ins, with a blank line in between. Most of the modern IDEs should have a command to do this for you.

```python
import json
import logging
import os
from asyncio import sleep as asleep
from typing import Optional

from discord import Embed
from discord.ext.commands import Bot, Context
```

bot/cogs/monstersurvey.py
dfitzpatrick: For clarification, do you want a path such as ../resources/monstersurvey/monstersurvey.json or are your equesting that monstersurvey.json live in the resources directly instead of its own subfolder?

sco1: Neither. The bot is invoked as a python module, `os.getcwd()` is going to return the root of this repository. `./resources` doesn't exist, `./bot/resources` does.

sco1:
Hey guys, any progress on the requested changes?

dfitzpatrick:
Tried to implement but I broke the bot. I did a git pull from the upstream (was trying to just make sure I had most recent commits) and now no commands are registering. Not sure what to do-- posted in hacktoberfest channel but couldn't get resolution.

sco1:
There haven't been any pushed changes to this branch since the last review, nor is there anything on it that would conflict with the upstream master, since you're creating two new files and not doing anything to the rest of the bot.

I would suggest resetting to the last commit on this branch and making the requested adjustments. This branch functioned just fine for testing.

bot/cogs/monstersurvey.py
dfitzpatrick: This was just an error in the embed. Feature worked as drafted. Updated embed to show correctly.

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+    async def monster_vote(self, ctx: Context, name: Optional[Union[int, str]] = None):
+        """Casts a vote for a particular monster, or displays a list of monsters that can be voted for
+        if one is not given."""
+        vote_embed = Embed(
+            name='Monster Voting',
+            color=0xFF6800
+        )
+        if isinstance(name, int):
+            name = list(self.voter_registry.keys())[name]
```
sco1: - This is off by one, since Python indexing starts at 0
- Needs to handle numbers that are out of range & provide feedback
- Should index by the leaderboard, not by the order of the source monster JSON dictionary. The user has no way to know which number corresponds to which monster otherwise.
- User needs some guidance in the voting embed that an integer vote is possible

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,178 @@
...
+                        name='Vote unsuccessful.',
+                        value='You already voted for this monster. '
+                              'If you want to change your vote, use another monster.',
+                        inline=False
+                    )
+                    log.info(f"{ctx.author.name} Tried to vote for the same monster.")
+                    await ctx.send(embed=vote_embed)
+                    await asleep(.5)
+                    return await ctx.invoke(self.monster_vote)
+        self.voter_registry[name]['votes'].append(ctx.author.id)
```
sco1: While you're checking `name.lower()` above, this will still throw an error. It would be easiest to lowercase `name` at the beginning of the function to avoid capitalization issues.

bot/cogs/monstersurvey.py
```diff
@@ -0,0 +1,178 @@
...
+        vote_embed.add_field(
+            name='Vote successful!',
+            value=f'You have successfully voted for {self.voter_registry[name]["full_name"]}!',
```
sco1: Since we're using one generic successful vote feedback embed, should include a line indicating that any previous votes have been removed.

Issue #28, opened 2018-10-08
[Feature Request] Spooky voice channel sounds


heavysaturn:
When the `.spookysound` command is triggered, the bot should connect to the hacktoberfest voice channel and play a spooky sound, such as ghosts oooing, chains rattling, or a bat fluttering its wings.

## Implementation details

- [ ] Find some spooky sound effects we can use for this. These **must** be licensed for free use.
- [ ] When the command is triggered, connect to the channel, play a random spooky sound effect, and disconnect again.
- [ ] Implement a short cooldown for the command (maybe 5 minutes), so people don't spam it.

sco1:
As a starting point in your search for free sounds with APIs:

- [Freesound](https://freesound.org/docs/api/)
- [Google Sound Library](https://developers.google.com/actions/tools/sound-library/)
- [BBC Sound Effects](http://bbcsfx.acropolis.org.uk) (advanced)

sco1:
Another possible option as a learning exercise for the Spotify API:

* [Stranger Things Halloween Sounds From The Upside Down](https://open.spotify.com/album/75CmkjqYUefIuXg649pKKJ)

meooow25:
I'd like to take this up :D

sco1:
Merged in #51

PR #51, opened 2018-10-14
Add cog to play spooky sounds

meooow25:
For issue #28 

A SpookySound cog plays a spooky sound when the command `.spookysound` is used.
The bot connects to a voice channel, plays the sound, and disconnects. The command cannot be used more than once per two minutes.


jodth07:
We should have the spooky sounds attached to a command, or listening to a set of words. See the implementation of spookey react. You can still keep it limited to once, every 2 mins though.

meooow25:
I'm not sure what you mean, since the behavior is exactly as described in issue #28. Perhaps I should have used a better PR description, updated now.

bot/cogs/spookysound.py
```diff
@@ -0,0 +1,39 @@
+import random
+from pathlib import Path
+
+import discord
+from discord.ext import commands
+
+HACKTOBERBOT_VOICE_CHANNEL_ID = 101010  # Replace with actual channel ID
```
heavysaturn: before we can merge this, this should actually be replaced.

```suggestion
HACKTOBERBOT_VOICE_CHANNEL_ID = 498804789287714816
```

bot/cogs/spookysound.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class SpookySound:
```
heavysaturn: Can we get a class docstring?

bot/cogs/spookysound.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        Connect to the Hacktoberbot voice channel, play a random spooky sound, then disconnect. Cannot be used more than
+        once in 2 minutes.
+        """
+        await ctx.send("Initiating spooky sound...")
+        voice = await self.channel.connect()
+        file_path = random.choice(self.sound_files)
+        src = discord.FFmpegPCMAudio(str(file_path.resolve()))
+
+        async def disconnect():
```
heavysaturn: I think this should just be a helper method that takes the voice object as a parameter. It's not ideal to define helper functions inside of methods.

bot/cogs/spookysound.py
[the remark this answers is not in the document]
meooow25: Updated.

meooow25: Updated.

bot/cogs/spookysound.py
meooow25: Done.

bot/cogs/spookysound.py
meooow25: Thanks, I was not aware of that. I have changed it to a static method.

meooow25:
Thanks, I have made the requested changes.

Pipfile
```diff
@@ -5,6 +5,7 @@ name = "pypi"
 
 [packages]
 "discord.py" = {ref = "rewrite", git = "https://github.com/Rapptz/discord.py"}
```
sco1: The `discord.py` requirement will need to be changed to:

```python
discord-py = {extras = ["voice"], ref = "rewrite", git = "https://github.com/Rapptz/discord.py"}
```

`PyNaCl` is already a requirement for this so it doesn't need to be added to our Pipfile

ByteCommander: @meooow25 Could you please make that requested change so that your PR is ready to be merged?

meooow25: Apologies for the delay, I am having trouble with locking the pipenv due to incompatible dependencies. I asked about it in the Discord server but received no solution.

`Could not find a version that matches pycodestyle<2.4.0,<2.5.0,<3.0.0,>=2.0.0,>=
2.4.0`

`pycodestyle < 2.4.0` and `>= 2.4.0` is of course impossible to satisfy. Some of the `flake8` packages are probably causing this. What should be done here?

sco1: This is some other issue with your environment, I just downloaded your branch and it installs & locks just fine.

We'll have to do some more research to figure out what's going on.

PR #52, opened 2018-10-14
Fixed doc string inaccuracy

skilldeliver:
Simply made new doc string description because the wrong one is pasted from bot\cogs\template - line 16.


jodth07:
unfortunately, this PR is too low effort to count.

PR #64, opened 2018-11-18
Allows you to create Seasons.

heavysaturn:
These are objects which change the bots behavior within specific date ranges. For example, a season can determine things like the avatar, the nickname, and which cogs are loaded. Season automatically changes according to the date range you specify when you create it.


heavysaturn:
Testing if the webhook is working now.

bot/seasons/season.py
```diff
@@ -0,0 +1,43 @@
...
+        await bot_member.edit(nick=self.bot_name)
+        await self.bot.user.edit(avatar=self.bot_avatar)
+
+        # Loads all the cogs for that season, and then the evergreen ones.
+        cogs = []
+        for cog_folder in (self.name, "evergreen"):
+            if cog_folder:
+                log.info(f'Start loading extensions from bot/cogs/{cog_folder}/')
+                for file in Path('bot', 'cogs', cog_folder).glob('*.py'):
+                    if not file.name.startswith("__"):
```
MarkKoz: You can exclude these in the glob: `"[!__]*.py"`.

bot/seasons/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    "halloween": Halloween,
+    "christmas": Christmas
+}
+
+
+def get_season(bot, season_name: str = None, date: datetime.date = None):
+    """
+    Returns a Season object based on either a string or a date.
+    """
+    assert season_name or date, "This function requires either a season or a date in order to run."
```
MarkKoz: Put a space after the docstring.

bot/seasons/season.py
heavysaturn: good suggestion!

bot/seasons/__init__.py
heavysaturn: yep, done.

scragly:
I feel the structure and model could do with some refinement. 

I have created a branch with recommended changes [here on a fork](https://github.com/scragly/seasonalbot/tree/season_systems_alteration) as I currently don't have branch push permissions for this repo.

The proposed structure would look like this instead of having both a cogs directory and a seasons one:
```
└── bot
    ├── __init__.py
    ├── __main__.py
    ├── bot.py
    ├── constants.py
    └── seasons
        ├── __init__.py
        ├── season.py
        ├── template.py
        ├── christmas
        │   └── __init__.py
        ├── evergreen
        │   ├── __init__.py
        │   └── uptime.py
        └── halloween
            ├── __init__.py
            ├── hacktoberstats.py
            ├── halloween_facts.py
            ├── halloweenify.py
            ├── movie.py
            └── spookyreact.py
```

As a general idea on how this would work, here's a run through on what happens on bot launch:

- `bot/__main__.py` loads the `bot.seasons` package as an extension.  

- `bot/seasons/__init__.py` has a `setup(bot)` function called on ext load, adding the cog class `SeasonManager` (renamed from the previous `Season` cog class)

- `bot/seasons/season.py` contains `SeasonManager` cog which runs the function `get_season` contained in the same module.

- `get_season` now dynamically generates the list of available seasons going by the available packages in the `bot/seasons/` directory.

- Season classes that inherit the `SeasonBase` class, with their details such as dates and new bot name are defined in the `__init__.py` of their respective season package. E.g. `bot/seasons/halloween/__init__.py:Halloween`.

- `SeasonManager` starts the `load_seasons` task (which used to be the `on_ready` cog event method).

- `SeasonBase.load` method automatically detects extensions as before, but now will load both module-type extensions and package-type extensions. This will enable better organisation for larger extension features if needed.

The primary benefits of this is the reduced maintenance when adding, removing or changing seasons and their extensions. You only need to make a new directory with a `__init__.py` that has the season class defined in it to create a new valid season. No code changes in `season.py` needed required.

I tried to keep changes in the fork's branch within the scope of this PR, but there was some adjustments made during the setting up of my own development instance of the bot as a matter of convenience, such as the constants and bot changes. Apologies.

scragly:
I have created a PR at #65 directed to this feature branch if you wish to use any of the alterations I've done.

heavysaturn:
Excellent work, @scragly. I will be using more or less all of that. It is a better design, and you still maintained all the important core design ideas.

scragly:
On re-review, I noticed some minor but possible issues:

- The load_seasons task won't ever end in the case the SeasonManager cog is unloaded or reloaded, and reloading will add a new task on top of the existing one. (my fault, I forgot about this when I changed it to a task)

- The editing of the user account's avatar and username is heavily ratelimited and the limit will be encountered often in development, and while less likely, still possible to encounter in production.

- If you change the seasons during development using the `season` command instead of launching into the season you want immediately, you will usually end up editing the user to the evergreen settings and then back to the season you want to test in. I felt this could be avoided with a season override setting.

- There's no debug mode for the bot to adjust logging levels.

- The builtin object is shadowed in `get_seasons`

- The path to the avatars could be set in the SeasonBase class for consistency and ease.

I've submitted another PR to this feature branch for consideration that address these issues, and I've separated the commits this time so they can be selectively dealt with if necessary.

scragly:
Another thing I noticed was from when a name isn't set, the current date had to be checked against season class attributes in order to find if any of their date ranges matches now. This is fine, but the date attributes were instance variables, so an instance of the season classes were required.

To avoid creating an instance, as these season classes may have other uses in future such as creating special tasks on load and such, I've moved the attributes to be class-level variables. 

In the `SeasonBase` class, I've moved the `start` and `end` instance variables to classmethods to match this.

The `current_year` attribute is now a staticmethod to ensure the year is generated on call when needed as previously it would not be accurate if the value was set before New Years, the bot was online until after New Years, task `load_seasons` was checking for the possible updated season only to not see the new year value.

Since these class attributes are now available at class level, there's no need for `SeasonBase` to have an `__init__` which brings the pleasant side-effect of being able to remove the `super().__init__()` calls in the season classes.

bot/constants.py
```diff
@@ -1,5 +1,70 @@
...
+    reddit = 458224812528238616
+    staff_lounge = 464905259261755392
+    verification = 352442727016693763
+
+
+class Client(NamedTuple):
+    guild = int(environ.get('SEASONALBOT_GUILD', 267624335836053506))
+    prefix = "."
+    token = environ.get('SEASONALBOT_TOKEN')
+    giphy_token = environ.get("GIPHY_TOKEN")
```
jos-b: In future may we not have tokens for services such as advent of code? Should these be contained in a seperate NamedTuple called `Tokens` or something? or is this just a me thing.

bot/seasons/halloween/hacktoberstats.py
```diff
@@ -10,16 +10,18 @@
...
 
+log = logging.getLogger(__name__)
 
-class Stats:
+
+class HacktoberStats:
     def __init__(self, bot):
         self.bot = bot
-        self.link_json = Path('./bot/resources', 'github_links.json')
+        self.link_json = Path("bot", "/resources", "github_links.json")
```
jos-b: I don't really understand how Path works at all but this feels wrong.
```py
>>> Path("bot", "resources", "github_links.json")
PosixPath('bot/resources/github_links.json')
>>> Path("bot", "/resources", "github_links.json")
PosixPath('/resources/github_links.json')
```
This looks like there is a `resources` dir at root, which I'm not aware of.

bot/seasons/halloween/halloween_facts.py
```diff
@@ -30,37 +32,25 @@ def __init__(self, bot):
...
-        facts = list(enumerate(self.halloween_facts))
-        while True:
-            # Avoid choosing each fact at random to reduce chances of facts being reposted soon.
-            random.shuffle(facts)
-            for index, fact in facts:
-                embed = self._build_embed(index, fact)
-                await self.channel.send("Your regular serving of random Halloween facts", embed=embed)
-                self.last_fact = (index, fact)
-                await asyncio.sleep(INTERVAL)
+    @property
```
jos-b: This being a property feels weird, any specific reason why it is?

bot/seasons/halloween/halloween_facts.py
```diff
@@ -30,37 +32,25 @@ def __init__(self, bot):
...
+    def random_fact(self):
+        return random.choice(self.facts)
 
-    @commands.command(name="hallofact", aliases=["hallofacts"], brief="Get the most recent Halloween fact")
-    async def get_last_fact(self, ctx):
+    @commands.command(name="spookyfact", aliases=["halloweenfact"], brief="Get the most recent Halloween fact")
```
jos-b: aliases should be a tuple

bot/seasons/halloween/scarymovie.py
```diff
@@ -18,7 +21,7 @@ class ScaryMovie:
     def __init__(self, bot):
         self.bot = bot
 
-    @commands.command(name='movie', alias=['tmdb'])
+    @commands.command(name='scarymovie', alias=['smovie'])
```
jos-b: aliases should be a tuple

jos-b: Actually, isn't the kwarg aliases?

jos-b: It is indeed aliases https://github.com/Rapptz/discord.py/blob/rewrite/discord/ext/commands/core.py#L169

bot/seasons/halloween/hacktoberstats.py
MarkKoz: It's a remnant of when it was on the main bot, which has does have a [resources directory](https://github.com/python-discord/bot/tree/master/bot/resources).

bot/seasons/halloween/scarymovie.py
scragly: A single item in brackets will not result in a tuple. You will want a trailing comma. 
However, as a note, the aliases kwarg accepts both tuples and lists, and the type error raised explicitly mentions list in it's output:
```python
        if not isinstance(self.aliases, (list, tuple)):
            raise TypeError("Aliases of a command must be a list of strings.")
```
https://github.com/Rapptz/discord.py/blob/rewrite/discord/ext/commands/core.py#L172

bot/seasons/halloween/halloween_facts.py
scragly: Same as https://github.com/python-discord/seasonalbot/pull/64#discussion_r236063406

jchristgit: The suggested change isn't a tuple though, you're missing a comma, it should be
```suggestion
    @commands.command(name="spookyfact", aliases=("halloweenfact",), brief="Get the most recent Halloween fact")
```

jchristgit: Oh, Scragly already referenced it

bot/seasons/halloween/halloween_facts.py
heavysaturn: I did not write this cooode. I guess I'll fix it if this is gonna be a review of all the hacktoberbot PRs.

bot/seasons/halloween/scarymovie.py
heavysaturn: solved

bot/seasons/halloween/spookyavatar.py
```diff
@@ -26,7 +29,7 @@ def __init__(self, bot):
             async with session.get(url) as resp:
                 return await resp.read()
 
-    @commands.command(name='savatar', aliases=['spookyavatar', 'spookify'],
+    @commands.command(name='savatar', aliases=)'spookyavatar', 'spookify'),
```
scragly:
```suggestion
    @commands.command(name='savatar', aliases=('spookyavatar', 'spookify'),
```
flippy bracket

PR #70, opened 2018-11-21
Refine Season Systems, with focus on ratelimit considerations and debug mode.

scragly:
Contributing to address points made in [this comment](https://github.com/python-discord/seasonalbot/pull/64#issuecomment-440475439) for the pending #64 PR.

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -42,6 +42,8 @@ class Client(NamedTuple):
     prefix = "."
     token = environ.get('SEASONALBOT_TOKEN')
     giphy_token = environ.get("GIPHY_TOKEN")
+    debug = environ.get('SEASONALBOT_DEBUG', '').lower() == 'true'
+    season_override = environ.get('SEASON_OVERRIDE')
 
 
 class Hacktoberfest(NamedTuple):
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -52,7 +52,7 @@ class Hacktoberfest(NamedTuple):
 
 
 class Roles(NamedTuple):
-    admin = 351758043643510788
+    admin = 267628507062992896
     announcements = 463658397560995840
     champion = 430492892331769857
     contributor = 295488872404484098
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -52,7 +52,7 @@ class Hacktoberfest(NamedTuple):
 
 
 class Roles(NamedTuple):
-    admin = 267628507062992896
+    admin = 351758043643510788
     announcements = 463658397560995840
     champion = 430492892331769857
     contributor = 295488872404484098
--- a/bot/seasons/season.py
+++ b/bot/seasons/season.py
@@ -7,7 +7,7 @@ from pathlib import Path
 
 from discord.ext import commands
 
-from bot.constants import Roles
+from bot.constants import Client, Roles
 from bot.decorators import with_role
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


bot/constants.py
```diff
@@ -52,7 +52,7 @@ class Hacktoberfest(NamedTuple):
 
 
 class Roles(NamedTuple):
-    admin = 267628507062992896
+    admin = 351758043643510788
```
MarkKoz: Was this a mistake or intentional?

bot/seasons/season.py
```diff
@@ -38,12 +38,19 @@ def get_season(bot, season_name: str = None, date: datetime.date = None):
...
+    # Check if a valid season name was provided
     if season_name:
         season_name = season_name.lower()
         if season_name not in seasons:
             season_name = 'evergreen'
 
+    # If there's a season name, we can just grab the correct class.
+    if season_name:
         season_lib = importlib.import_module(f'bot.seasons.{season_name}')
         season_class = getattr(season_lib, season_name.capitalize())
```
MarkKoz: Should this be put in a utility function? Same thing is done three times in this function.

bot/seasons/season.py
```diff
@@ -38,12 +38,19 @@ def get_season(bot, season_name: str = None, date: datetime.date = None):
...
+        season_name = Client.season_override
+
+    # Check if a valid season name was provided
     if season_name:
         season_name = season_name.lower()
         if season_name not in seasons:
             season_name = 'evergreen'
 
+    # If there's a season name, we can just grab the correct class.
+    if season_name:
```
MarkKoz: Can this not just be moved under the previous if statement?

bot/constants.py
scragly: This was a mistake. I'll fix it up.

bot/seasons/season.py
scragly: Good idea, I'd edited this part so much that I didn't even think of it. Thanks.

bot/seasons/season.py
scragly: It can, again a remnant of some deleted logic. Thanks.

scragly:
Some further adjustments were added in the last commit that address the previous recommendations, and also change the behaviour of season class date checks, avoiding unnecessary instance creation for classes that don't match the current date range.

Additional details are explained [here](https://github.com/python-discord/seasonalbot/pull/64#issuecomment-440498358)

bot/seasons/season.py
```diff
@@ -38,48 +43,79 @@ def get_season(bot, season_name: str = None, date: datetime.date = None):
...
-        self.end = datetime.datetime.strptime(f"{self.end_date}-{current_year}", date_format).date()
+    @classmethod
+    def start(cls):
+        return datetime.datetime.strptime(f"{cls.start_date}-{cls.current_year()}", cls.date_format).date()
+
+    @classmethod
+    def end(cls):
+        return datetime.datetime.strptime(f"{cls.end_date}-{cls.current_year()}", cls.date_format).date()
+
+    @staticmethod
```
MarkKoz: While functionally the same, I do think a static method is more appropriate for these.

bot/seasons/season.py
```diff
@@ -38,48 +43,79 @@ def get_season(bot, season_name: str = None, date: datetime.date = None):
...
-    def __init__(self):
-        current_year = datetime.date.today().year
-        date_format = "%d/%m-%Y"
-        self.start = datetime.datetime.strptime(f"{self.start_date}-{current_year}", date_format).date()
-        self.end = datetime.datetime.strptime(f"{self.end_date}-{current_year}", date_format).date()
+    @classmethod
+    def start(cls):
+        return datetime.datetime.strptime(f"{cls.start_date}-{cls.current_year()}", cls.date_format).date()
+
+    @classmethod
```
MarkKoz: While functionally the same, I do think a static method is more appropriate for these.

scragly: It's not functionally the same. A static method does not return the class as an argument.

MarkKoz: I mean that in the sense that they can both do the same thing. You can always access `cls` by naming it yourself: `SeasonBase.whatever`.

scragly: Sorry, I disagree that a static method should be used in place of a class method in cases where the class is actually of use within. Class methods are not required to return an instance of it's own class, that's just it's most common usage. I find the usage of a class method more proper and elegant in style.

Regardless though, in this particular case, a class method is definitely required because the class that is being passed as the `cls` argument is not `SeasonBase`: it is the class that has inherited `SeasonBase`, the specific season classes such as Halloween. If I used `SeasonBase` as the origin, it would only result in the attributes `start_date` and `end_date` not being defined.

heavysaturn: Agree with scragly here, resolving comment.

PR #76, opened 2018-11-24
Fix markdown for further reading links

SnowyLuma:
Hey there, I noticed that the links in the further reading section currently do not work correctly.

I've gone ahead and fixed them while slightly rewording the sentence to make it sound more natural. Please let me know if I should revert the second change!


heavysaturn:
The build only fails due to a fail that this PR doesn't alter. In other words, it's probably fine to merge it, so I'm gonna do that.

PR #80, opened 2018-12-01
Season Improvements

### LINKED ISSUE #71: Season change improvements.
When the bot changes the season, it should make an announcement and change the server icon to match its new avatar. It is important that this **does not** happen during testing, only during production.

## Implementation details

- [ ] When the season changes, the bot should change the server icon.
- [ ] Instead of storing the icons in this repo, we should be getting them out of the [branding](https://github.com/python-discord/branding) repo.
- [ ] When the season changes, the bot should make an announcement in #announcements declaring that a new season has started, that the server icon has changed, and that the following bot commands are now made available. Then a list of the commands that have been made available, ideally with a description (taken from docstring first line, for example). This announcement should tag the `@Announcements` role.
- [ ] Ensure that we make use of the DEBUGMODE implemented by Scragly so that neither avatar changes, server icon changes nor public announcements will happen when we're just testing the bot.

scragly:
Resolves #71

- [x] When the season changes, the bot should change the server icon.  

- [x] Instead of storing the icons in this repo, we should be getting them out of the branding repo.

- [x] When the season changes, the bot should make an announcement in #announcements declaring that a new season has started, that the server icon has changed, ~~and that the following bot commands are now made available. Then a list of the commands that have been made available, ideally with a description (taken from docstring first line, for example).~~ _Changed to showing loaded season cogs and their docstrings, along with instruction to use `.help Category` to see commands._ This announcement should tag the @Announcements role.
image

- [x] Ensure that we make use of the DEBUGMODE implemented by Scragly so that neither avatar changes, server icon changes nor public announcements will happen when we're just testing the bot.

Additionally:
- Season announcements won't be sent if username is not changed on season load, to prevent announcements on bot restart, season reloads and to ensure everyone isn't pinged before things are working.

- Added an `.announce` command to manually send an announcement to the announcement channel. The announcement channel ID can be changed in testing with the environmental variable `CHANNEL_ANNOUNCEMENTS`. The announcement role does not need to be changed, as it will build the mention string manually, avoiding the need for more IDs to be set in testing.

- Added a `.seasons` command to show the available season extensions. Bolded time periods are those that are recognised as matching the current date. The currently active season is underlined.
image

- Added a `.refresh` command group. Let's you individually set or re-set the currently loaded season's avatar (`.refresh avatar`), server icon (`.refresh icon`) or username (`.refresh username`). These are fully actioned, even if in debug mode. The response will say if it changed successfully, or if something wasn't successful, or if only nickname was set with the username one.
image  
image  
image


bot/seasons/christmas/__init__.py
```diff
@@ -2,15 +2,17 @@
...
-    bot_name = "Santabot"
-
-    def __init__(self, bot):
-        self.bot = bot
-
-    @property
-    def bot_avatar(self):
-        with open(self.avatar_path("christmas.png"), "rb") as avatar:
-            return bytearray(avatar.read())
+    bot_name = "Merrybot"
```
jos-b: Could we have an additional line here, I know properties aren't exactly the same as functions and whatnot but it would feel a lot cleaner imo.
```suggestion

    bot_name = "Merrybot"
```

bot/seasons/halloween/__init__.py
```diff
@@ -3,14 +3,8 @@
 
 class Halloween(SeasonBase):
     name = "halloween"
+    greeting = "Happy Halloween!"
     start_date = "01/10"
     end_date = "31/10"
     bot_name = "Spookybot"
```
jos-b: Position of `bot_name` is inconsistent between seasons, with the Christmas cog it is with the avatar but here it is at the top of the class.

bot/seasons/halloween/__init__.py
```diff
@@ -3,14 +3,8 @@
...
-
-    def __init__(self, bot):
-        self.bot = bot
-
-    @property
-    def bot_avatar(self):
-        with open(self.avatar_path("spooky.png"), "rb") as avatar:
-            return bytearray(avatar.read())
+    icon = "/logos/logo_seasonal/halloween/spooky.png"
```
jos-b: Additional line here would be nice.
```suggestion

    icon = "/logos/logo_seasonal/halloween/spooky.png"
```

bot/seasons/christmas/__init__.py
scragly: I don't see the logical grouping you're using in this case.

bot/seasons/halloween/__init__.py
scragly: If we're going to end up changing the ordering of the season dataclasses, we're better off properly establishing proper logical grouping then.

May I suggest:
```py

    # strings
    name = "halloween"
    bot_name = "Spookybot"
    greeting = "Happy Halloween!"

    # dates
    start_date = "01/10"
    end_date = "31/10"

    # resources
    icon = "/logos/logo_seasonal/halloween/spooky.png"
```

bot/seasons/halloween/__init__.py
scragly: My response in the prior comment above has a suggested edit that would add a new line as you want.

bot/constants.py
```diff
@@ -4,14 +4,14 @@
 
 from bot.bot import SeasonalBot
 
-__all__ = ('Client', 'Roles', 'bot')
+__all__ = ("Channels", "Client", "Roles", "bot")
```
heavysaturn: this should probably contain _all_ the classes, right?
```suggestion
__all__ = ("AdventOfCode", "Channels", "Client", "Colours", "Emojis", "Hacktoberfest", "Roles", "Tokens", "bot")
```

These are arranged alphabetically, and the actual classes should probably be moved around to match that ordering.

bot/seasons/halloween/__init__.py
```diff
@@ -3,14 +3,11 @@
...
-    def __init__(self, bot):
-        self.bot = bot
+    start_date = "01/10"
+    end_date = "31/10"
 
-    @property
-    def bot_avatar(self):
-        with open(self.avatar_path("spooky.png"), "rb") as avatar:
-            return bytearray(avatar.read())
+    colour = 0xe67e22
```
heavysaturn: I feel like this should probably be a constant in the Colours dataclass?

bot/seasons/season.py
```diff
@@ -161,20 +312,176 @@ def __init__(self, bot):
...
+            is_current = season.is_between_dates(datetime.datetime.utcnow())
+            pdec = "**" if is_current else ""
+
+            # underline currently active season
+            is_active = current_season == season.name
+            sdec = "__" if is_active else ""
+
+            entries.append(
+                f"**{sdec}{season.__name__}:{sdec}**\n"
+                f"{forced_space*3}{pdec}{period}{pdec}"
```
heavysaturn:
```suggestion
                f"{forced_space*3}{pdec}{period}{pdec}\n"
```

I think this would look better with an empty line between the seasons.

bot/constants.py
scragly: No probs. Will do.

bot/seasons/halloween/__init__.py
scragly: Yep, that makes sense. I've added it in.

bot/seasons/season.py
scragly: Sure thing. Done:
image

heavysaturn: looks great!

scragly:
OK, I squashed the smaller edits together for the last few things I've done and the previous review recommendations. This should be ready for re-review.

bot/constants.py
```diff
@@ -45,13 +58,26 @@ class Client(NamedTuple):
...
+class Colours:
+    soft_red = 0xcd6d6d
+    soft_green = 0x68c290
+    dark_green = 0x1f8b4c
+    orange = 0xe67e22
+
+
+class Emojis:
+    star = "\u2B50"
+    christmas_tree = u"\U0001F384"
```
jchristgit: Hm, do we need `u` prefixes on 3?

bot/constants.py
```diff
@@ -45,13 +58,26 @@ class Client(NamedTuple):
...
+
+
 class Hacktoberfest(NamedTuple):
     channel_id = 498804484324196362
     voice_id = 514420006474219521
 
 
 class Roles(NamedTuple):
-    admin = 267628507062992896
+    admin = int(environ.get('SEASONALBOT_ADMIN', 267628507062992896))
```
jchristgit: I would prefer a more expressive env var such as `SEASONALBOT_ADMIN_ROLE_ID`.

bot/seasons/season.py
```diff
@@ -1,38 +1,45 @@
...
+
     seasons = []
 
-    for module in pkgutil.iter_modules([Path('bot', 'seasons')]):
+    for module in pkgutil.iter_modules([Path("bot", "seasons")]):
         if module.ispkg:
-            seasons.append(module[1])
-
+            seasons.append(module.name)
     return seasons
```
jchristgit: Couldn't we have this `yield` the modules?

bot/seasons/season.py
```diff
@@ -52,82 +59,289 @@ def get_season(bot, season_name: str = None, date: datetime.date = None):
...
         else:
-            if self.bot.user.name != self.bot_name:
+            if bot.user.name != self.bot_name:
                 # attempt to change user details
                 log.debug(f"Changing username to {self.bot_name}")
-                await self.bot.user.edit(name=self.bot_name, avatar=self.bot_avatar)
+                try:
+                    await bot.user.edit(username=self.bot_name)
+                except discord.HTTPException:
+                    pass
```
jchristgit: Prefer `contextlib.suppress`

bot/seasons/season.py
```diff
@@ -52,82 +59,289 @@ def get_season(bot, season_name: str = None, date: datetime.date = None):
...
+        # track old avatar hash for later comparison
+        old_avatar = bot.user.avatar
+
+        # attempt the change
+        log.debug(f"Changing avatar to {self.icon}")
+        icon = await self.get_icon()
+        try:
+            async with async_timeout.timeout(5):
+                await bot.user.edit(avatar=icon)
+        except (discord.HTTPException, asyncio.TimeoutError):
```
jchristgit: Prefer `contextlib.suppress`

bot/seasons/season.py
```diff
@@ -52,82 +59,289 @@ def get_season(bot, season_name: str = None, date: datetime.date = None):
...
+            pass
+
+        if bot.user.avatar != old_avatar:
+            log.debug(f"Avatar changed to {self.icon}")
+            return True
+
+        log.warning(f"Changing avatar failed: {self.icon}")
```
jchristgit: Why wouldn't we directly check the exceptions, though?

bot/seasons/season.py
```diff
@@ -52,82 +59,289 @@ def get_season(bot, season_name: str = None, date: datetime.date = None):
...
+        except (discord.HTTPException, asyncio.TimeoutError):
+            pass
+
+        new_icon = bot.get_guild(Client.guild).icon
+        if new_icon != old_icon:
+            log.debug(f"Server icon changed to {self.icon}")
+            return True
+
+        log.warning(f"Changing server icon failed: {self.icon}")
+        return False
```
jchristgit: Ditto.

bot/constants.py
scragly: No, I didn't write this though. I only moved the ordering of the constants classes as per lemon's suggestions. Since I'm here anyway, I'll remove it.

bot/constants.py
scragly: Sure.

bot/seasons/season.py
scragly: For what benefit.

scragly: Also this feels unrelated to the current PR.

bot/seasons/season.py
scragly: Because failure can occur without an exception.

bot/seasons/season.py
scragly: Because failure can occur without an exception.

PR #85, opened 2018-12-02
Fix center alignment on avatars

neonsea:
While fixing issue https://github.com/python-discord/seasonalbot/issues/78 I noticed none of the avatars were perfectly centered. 

This commit fixes the vertical and horizontal alignment on all of the avatars in relation to the icon itself (not including the shadows).


scragly:
You're likely going to hate me, but I've got a PR pending that actually deletes these entirely and uses instead the branding repository. Sorry!

neonsea:
Righto, I should probably check the PRs before submitting my own, my bad.

scragly:
Definitely not your fault. That PR is a bit... long to look through.

I should have updated my Issue ticket. Thanks though for making the effort. Feel free to visit the branding repo to address positioning on those resources though, because they're actually in the same situation.

PR #88, opened 2018-12-05
Send devlog notification on season load.

### LINKED ISSUE #83: Deliver a notification when bot has started - like Python.
Right now, there's no easy way to check if the bot has restarted (you'd have to keep doing `.uptime`). Python delivers an embed into #dev-logs when it starts, we should probably be doing that for this bot as well.

scragly:
Resolves #83 

A notification is sent to devlog on the new loading of a season. This means it will effectively send a notification on:
- Season changes to a special season
- Season changes to evergreen
- Bot starts/restarts and loads the current season

These notifications will be sent regardless of debug mode setting.

Notification posted to devlog looks like:
image

Additionally resolves a bug with the `load_seasons` method where on checking for different seasons at midnight, it would always think it's a new one, resulting in the devlog notification being re-sent every midnight.

### DIFF
--- a/bot/seasons/season.py
+++ b/bot/seasons/season.py
@@ -37,7 +37,8 @@ def get_season_class(season_name: str) -> Type["SeasonBase"]:
     """
 
     season_lib = importlib.import_module(f"bot.seasons.{season_name}")
-    return getattr(season_lib, season_name.capitalize())
+    class_name = season_name.replace("_", " ").title().replace(" ", "")
+    return getattr(season_lib, class_name)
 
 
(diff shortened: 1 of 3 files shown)


bot/seasons/season.py
```diff
@@ -326,7 +330,7 @@ def cog_name(cog):
         # Apply seasonal elements after extensions successfully load
         username_changed = await self.apply_username(debug=Client.debug)
 
-        # Avoid heavy API ratelimited elements when debugging
+        # Avoid major changes and annoucements if debug mode
```
jchristgit: Minor typo:
```suggestion
        # Avoid major changes and announcements if debug mode
```

PR #89, opened 2018-12-05
Private Leaderboard Daily Completion Statistics

sco1:
Adds an `.aoc stats` command to display the current completion statistics, by day

image

Also makes a minor fix to the leaderboard parsing regex to account for twitter handles, which start with a leading `@`.


bot/seasons/christmas/adventofcode.py
```diff
@@ -247,9 +247,53 @@ def __init__(self, bot: commands.Bot):
             embed=aoc_embed,
         )
 
+    @adventofcode_group.command(
+        name="stats",
+        aliases=("dailystats", "ds"),
+        brief=("Get daily statistics for the PyDis private leaderboard")
```
jchristgit: Don't need the parens

PR #91, opened 2018-12-10
AoC Join Code Refactor

sco1:
Refactor `.aoc join` to DM the user with the code.

Also updates the `.aoc about` to direct the user to get the current leaderboard code from the `.aoc join` command.

### DIFF
--- a/bot/seasons/christmas/adventofcode.py
+++ b/bot/seasons/christmas/adventofcode.py
@@ -210,9 +210,7 @@ class AdventOfCode:
             await author.send(info_str)
         except discord.errors.Forbidden:
             log.debug(f"{author.name} ({author.id}) has disabled DMs from server members")
-            await ctx.send(
-                f':x: {author.mention}, please (temporarily) enable DMs to receive the join code'
-            )
+            await ctx.send(f":x: {author.mention}, please (temporarily) enable DMs to receive the join code")
 
     @adventofcode_group.command(
         name="leaderboard",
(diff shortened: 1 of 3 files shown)


bot/seasons/christmas/adventofcode.py
```diff
@@ -196,14 +198,37 @@ def __init__(self, bot: commands.Bot):
...
+        """
+        Staff command to update the AoC join code constant locally to allow for the code to be updated
+        on regeneration without having to redeploy the bot
+        """
+
+        author = ctx.message.author
+        log.info(f"{author.name} ({author.id}) has changed the PyDis AoC leaderboard code")
+
+        global AOC_JOIN_CODE  # Necessary (probably?) evil to update the code without redeploying
+        AOC_JOIN_CODE = new_code
```
jos-b: Is there any point in this command? we'll have to update the bot and reboot anyway with the new env so we might as well just do it the proper way.

bot/seasons/christmas/adventofcode.py
```diff
@@ -196,14 +196,23 @@ def __init__(self, bot: commands.Bot):
...
             "Head over to https://adventofcode.com/leaderboard/private "
             f"with code `{AocConfig.leaderboard_join_code}` to join the PyDis private leaderboard!"
         )
-        await ctx.send(info_str)
+        try:
+            await author.send(info_str)
+        except discord.errors.Forbidden:
+            log.debug(f"{author.name} ({author.id}) has disabled DMs from server members")
+            await ctx.send(
+                f':x: {author.mention}, please (temporarily) enable DMs to receive the join code'
```
GhostofGoes: Single quoted string

GhostofGoes: Why is the call split across 3 lines? It doesn't seem like it violates the line length limit.

PR #92, opened 2018-12-26
Fix advent of code countdown

jos-b:
So currently the countdown code is not only running until the 25th but also on the 25th which means it is counting down to a day 26.

This pull request makes the following changes:
- Only run advent of code countdown timers from December 1st to December 25th.
- Don't let the countdown command run when it is not December 1st to December 25th.

### DIFF
--- a/bot/seasons/christmas/adventofcode.py
+++ b/bot/seasons/christmas/adventofcode.py
@@ -180,6 +180,10 @@ class AdventOfCode:
         """
         Return time left until next day
         """
+        if not is_in_advent():
+            await ctx.send(f"Advent of Code is no longer running this year! Come back next year for more puzzles!")
+            return
+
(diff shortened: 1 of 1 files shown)


bot/seasons/christmas/adventofcode.py
```diff
@@ -179,6 +180,10 @@ def __init__(self, bot: commands.Bot):
         """
         Return time left until next day
         """
+        if not is_in_advent():
+            await ctx.send(f"Advent of Code is no longer running this year! Come back next year for more puzzles!")
```
heavysaturn: Maybe it can just return the time until december 1st of next year?

`The Advent of Code event is not currently running. The next event will start in {time_until_next_event}`

jos-b: Added in 049a42d825fe12fe611783dff5531f4fd05af56c

PR #97, opened 2019-02-05
Add markov Poem Generator

martmists:
Note: I have no clue how pipfiles work, so if I did it wrong please let me know.

The code works by first generating a markov of the corpus, then generating 14 lines and ordering these in two quatrains and two triplets, like shakespeare's sonnets.

TODO:

- [ ] Add more sources (requested by Bisk)
- [x] Make it rhyme (requested by Lemon)

### DIFF
--- a/bot/seasons/christmas/adventofcode.py
+++ b/bot/seasons/christmas/adventofcode.py
@@ -13,7 +13,8 @@ from bs4 import BeautifulSoup
 from discord.ext import commands
 from pytz import timezone
 
-from bot.constants import AdventOfCode as AocConfig, Colours, Emojis, Tokens
+from bot.constants import AdventOfCode as AocConfig
+from bot.constants import Colours, Emojis, Tokens
 
 log = logging.getLogger(__name__)
 
--- a/bot/seasons/valentines/poem_generator.py
+++ b/bot/seasons/valentines/poem_generator.py
@@ -1,7 +1,10 @@
+import aiohttp
 import logging
+from typing import List, Dict
 
 from discord.ext import commands
 import markovify
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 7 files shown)


martmists:
If anyone else can provide more corpus data (love poems by well-known poets), feel free to link it here.

AnonGuy:
Could you reset changes to the `Pipfile.lock`? That may be what's causing the Azure build to fail.

martmists:
Then how will I add my dependency (Markovify)?

sco1:
@martmists the pipenv issues should be resolved now, leaving you with some linting issues to address.

bot/seasons/valentines/poem_generator.py
```diff
@@ -0,0 +1,80 @@
...
+        return "\n\n".join([  # Section dividers
+            "\n".join(lines[:4]),  # 4, 4, 3, 3, like a shakespearian sonnet
+            "\n".join(lines[4:8]),
+            "\n".join(lines[8:11]),
+            "\n".join(lines[11:])
+        ])
+
+    @commands.command()
+    async def poem(self, ctx):
+        await ctx.send(await self.generate_poem())
```
scragly: Occasionally the poems may take a while due to it looping continuously and having to wait for two http requests for each line. There's no real fail-break either if it gets stuck in an excessively long loop.

It would be good to put in the ctx.typing context manager around the generate_poem call to ensure users can see some form of feedback that the bot is working, preventing them from thinking it's broken and just trying the command again.

Wrapping with `async_timeout.timeout(10)` too would prevent a bad luck case of infinite/excessive looping, which would inevitably smash the datamuse API.

As it is already, this command during testing made 66 API requests after 5 uses.

It's appreciated that there's caching involved, but it may not resolve the issue if the sample size is too large for it to make a significant impact. That would require much more testing to find out, but it's better to be safe and polite to the API anyway.

A user cooldown perhaps should be added to the command to avoid simultaneous and excessive use.

bot/seasons/christmas/adventofcode.py
```diff
@@ -13,7 +13,8 @@
 from discord.ext import commands
 from pytz import timezone
 
-from bot.constants import AdventOfCode as AocConfig, Colours, Emojis, Tokens
+from bot.constants import AdventOfCode as AocConfig
+from bot.constants import Colours, Emojis, Tokens
```
scragly: This is not in the scope of this PR. Please revert this change.

bot/utils/halloween/spookifications.py
```diff
@@ -1,8 +1,7 @@
 import logging
 from random import choice, randint
 
-from PIL import Image
-from PIL import ImageOps
+from PIL import Image, ImageOps
```
scragly: This is not in the scope of this PR. Please revert this change.

bot/seasons/valentines/poem_generator.py
```diff
@@ -0,0 +1,80 @@
...
+import aiohttp
+import markovify
+from discord.ext import commands
+
+log = logging.getLogger(__name__)
+
+
+class PoemGenerator:
+    SOURCES = [
+        "shakespeare corpus.txt"
```
scragly: Spaces in resource file names are generally discouraged, and other resources already avoid them also, making it not really match the current project style. Could this be renamed without it?

bot/seasons/valentines/poem_generator.py
```diff
@@ -0,0 +1,80 @@
...
+    ]
+
+    def __init__(self):
+        full_corpus = ""
+
+        for file in self.SOURCES:
+            with open(f"bot/resources/valentines/{file}") as f:
```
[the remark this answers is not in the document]
jchristgit: not only recommended, not using that makes it impossible to run the bot on windows

heavysaturn:
When @martmists was approached on the server about looking at this PR, he said he was busy and can look at it in 4 weeks. This has a number of critical unaddressed review comments that must be fixed before it can be merged. I'm also disappointed that it doesn't have a more interesting corpus than Shakespeare. bisk, another user on the server, made several promising suggestions for authors that would make this feature more interesting, and an attempt to extend the corpus with works from these authors should be made.

The authors suggested are:
John Keats, William Wordsworth, Lord Byron, William Blake, Rabby Burnes, Walter Scott.

I am closing this PR, but I'm opening an issue for this feature. This PR can be used as the basis, all review comments must be addressed, and the corpus should be expanded to include at least some of the authors mentioned above. When that's done, we'll merge it.

PR #99, opened 2019-02-05
Add Lovefest configuration, update project documents

sco1:
- Add Lovefest role & channel IDs to bot constants
- Update environment variable name for seasonalbot (was hacktoberbot)
- Update Contributor doc to reference pipenv guide rather than the deprecated Seasonalbot GH wiki

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ All projects evolve over time, and this contribution guide is no different. This
 
 ##  Supplemental Information
 ### Developer Environment
-Seasonalbot utilizes [Pipenv](https://pipenv.readthedocs.io/en/latest/) for installation and dependency management. For users unfamiliar with the Pipenv workflow, a [Pipenv Primer](https://github.com/python-discord/seasonalbot/wiki/Hacktoberbot-and-Pipenv) is provided in Seasonalbot's Wiki.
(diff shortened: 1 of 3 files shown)


scragly:
If someone was using an editor other than PyCharm, there was little to no instructions for them in the current README. Since this PR is updating the README, I've added the following changes to here to keep things together:

As a brief summary:
- An index with links direct to sections has been added to help those who may want to re-reference certain sections and avoids the extra scrolling.
- A requirements list has been added with links on how they can install it.
- Images have been updated to new instructions.
- Instructions and screenshots using `hacktoberbot` as the project name have been replaced.
- Environment variables have been expanded to include new required values and a note about each season possibly having their own requirements.
- New instructions on how to easily add all the environment variables in one step with the paste function.
- PyCharm project script path updated to `bot` directory instead of `bot.py` and adding the `-m` interpreter option.
- Mentioning feature branch workflow instead of committing to fork master to prevent pulling back your own commits from upstream again.
- Noting that people shouldn't attempt to commit directly to the original repo's `master`. They won't have rights anyway but it was recommended to add.
- Merging the `Not PyCharm` listed resources with `Further Reading`.

To quickly see the updated README without diving through the branch files or the diff:
https://gist.github.com/scragly/590f264f06352e195a4b4010dfe06471

Issue #96, opened 2019-02-03
[Feature Request] Love Calculator!


AnonGuy:
Create a bot command that calculates an affinity value based on two users: The author of the message and the user pinged by the message. This could be implemented by adding the two user IDs, and generating a percentage value from this.

![image](https://user-images.githubusercontent.com/25161069/52178931-1162a580-27cc-11e9-82e7-d03aa89ae975.png)

An example of this functionality can be seen here: https://www.lovecalculator.com

sco1:
Merged in #100

PR #100, opened 2019-02-05
Add LoveCalculator as described in #96

darthdelay:
This love calculator uses two user-IDs, adds them together and takes the
modulus of 100 from it.

**Invocation:** .love [mention-user] [mention-user]
**Aliases:** .love_calculator .love_calc

### DIFF
--- a/bot/seasons/valentines/lovecalculator.py
+++ b/bot/seasons/valentines/lovecalculator.py
@@ -1,98 +1,58 @@
+import json
 import logging
-from random import randint
+from pathlib import Path
+from random import choice
 
 import discord
 from discord.ext import commands
 
+from bot.constants import Roles
 
-logger = logging.getLogger(__name__)
-
-
-LOVE_STAT = {
-    '0':
-        [
-            '\U0001F49B When will you two marry? \U0001F49B',
-            '\U0001F49B Now kiss already \U0001F49B'
-        ],
-    '95':
-        [
-            '\U0001f496 Love is in the air \U0001f496',
-            '\U0001f496 Planned your future yet? \U0001f496'
-        ],
-    '80': '\U0001f495 Aww look you two fit so well together \U0001f495',
-    '60': '\u2764 So when will you two go on a date? \u2764',
-    '45': '\U0001F498 You two are really close aren\'t you? \U0001F498',
-    '30': '\U0001F497 You seem like you are good friends \U0001F497',
-    '20': '\U0001F49C You two seem like casual friends \U0001F49C',
-    '5': '\U0001F499 A small acquaintance \U0001F499',
-    '1': '\U0001F494 There\'s no real connection between you two \U0001F494'
-}
-LOVE_TXT = {
-    '0': "You two will most likely have the perfect relationship. But don't think that this means you don't have to "
-         "do anything for it to work. Talking to each other and spending time together is key, even in a seemingly "
-         "perfect relationship.",
-    '95': "Your relationship will most likely work out perfect. This doesn't mean thought that you don't need to put "
-          "effort into it. Talk to each other, spend time together, and you two wont have a hard time.",
-    '80': "Your relationship will most likely work out well. Don't hesitate on making contact with each other though, "
-          "as your relationship might suffer from a lack of time spent together. Talking with each other and spending "
-          "time together is key.",
-    '60': "Your relationship will most likely work out. It won't be perfect and you two need to spend a lot of time "
-          "together, but if you keep on having contact, the good times in your relationship will overweigh the bad "
-          "ones.",
-    '45': "Your relationship has a reasonable amount of working out. But do not overestimate yourself there. Your "
-          "relationship will suffer good and bad times. Make sure to not let the bad times destroy your relationship, "
-          "so do not hesitate to talk to each other, figure problems out together etc.",
-    '30': "The chance of this relationship working is not very high, but its not that low either. If you both want "
-          "this relationship to work, and put time and effort into it, meaning spending time together, talking to "
-          "each other etc., than nothing shall stand in your way.",
-    '20': "The chance of this relationship working is not very high. You both need to put time and effort into this "
-          "relationship, if you want it to work out well for both of you. Talk with each other about everything and "
-          "don't lock yourself up. Spend time together. This will improve the chances of this relationship's survival "
-          "by a lot",
-    '5': "There might be a chance of this relationship working out somewhat well, but it is not very high. With a lot "
-         "of time and effort you'll get it to work eventually, however don't count on it. It might fall apart quicker "
-         "than you'd expect.",
-    '1': "The chance of this relationship working out is really low. You can get it to work, but with high costs and "
-         "no guarantee of working out. Do not sit back, spend as much time together as possible, talk a lot with each "
-         "other to increase the chances of this relationship's survival "
-}
+log = logging.getLogger(__name__)
 
 
 class LoveCalculator:
+    """
+    A cog for calculating the love between two people
+    """
     def __init__(self, bot):
         self.bot = bot
 
     @commands.command(aliases=('love_calculator', 'love_calc'))
-    async def love(self, ctx, name_one: discord.Member, name_two: discord.Member):
-        love_meter = (name_one.id + name_two.id) % 100
-        if love_meter >= 95:
-            love_status = LOVE_STAT['95'][randint(0, len(LOVE_STAT['95']))]
-            love_stat_i = '95'
-        elif 80 <= love_meter < 90:
-            love_status = LOVE_STAT['80']
-            love_stat_i = '80'
-        elif 60 <= love_meter < 80:
-            love_status = LOVE_STAT['60']
-            love_stat_i = '60'
-        elif 45 <= love_meter < 60:
-            love_status = LOVE_STAT['45']
-            love_stat_i = '45'
-        elif 30 <= love_meter < 45:
-            love_status = LOVE_STAT['30']
-            love_stat_i = '30'
-        elif 20 <= love_meter < 30:
-            love_status = LOVE_STAT['20']
-            love_stat_i = '20'
-        elif 5 <= love_meter < 20:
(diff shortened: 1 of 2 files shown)


bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,119 @@
...
+from random import randint
+
+import discord
+from discord.ext import commands
+
+
+logger = logging.getLogger(__name__)
+
+
+LOVE_STAT = {
```
sco1: These constants would probably be better in a JSON. #105 uses a `bot/resources/valentines/` directory, which will work fine for any static resources you need to add.

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,119 @@
...
+            color=discord.Color.dark_magenta()
+        )
+        embed.add_field(
+            name='A letter from Dr. Love:',
+            value=LOVE_TXT[love_stat_i]
+        )
+
+        await ctx.message.channel.send(embed=embed)
+
+    @love.error
```
sco1: These error handlers should provide some feedback to the user

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,119 @@
...
+    async def love_error(self, ctx, error):
+        if isinstance(error, commands.MissingRequiredArgument):
```
sco1: Rather than failing out completely if 2 members aren't provided, how about running it against TenderBot?

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,119 @@
...
+         "no guarantee of working out. Do not sit back, spend as much time together as possible, talk a lot with each "
+         "other to increase the chances of this relationship's survival "
+}
+
+
+class LoveCalculator:
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command(aliases=('love_calculator', 'love_calc'))
```
sco1: Since this command accepts user mentions, this should probably have a [cooldown](https://discordpy.readthedocs.io/en/rewrite/ext/commands/api.html#discord.ext.commands.cooldown) in order to reduce spamming.

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,119 @@
...
+    async def love(self, ctx, name_one: discord.Member, name_two: discord.Member):
+        love_meter = (name_one.id + name_two.id) % 100
+        if love_meter >= 95:
```
scragly: This section is not very DRY and can be simplified. Assuming the data structure is as per my previous recommendations, we can first make a variable once on extension load to cast the available levels to a list of ints after loading the data up at the beginning of the module:
```py
with open(Path("bot", "resources", "valentines", "love_matches.json"), "r") as file:
    LOVE_DATA = json.load(file)
LOVE_LEVELS = [int(x) for x in LOVE_DATA]
```
This avoids the need for us to cast the keys to int multiple times for each usage of the command later.

Then this section can be simplified to something like:
```py
        love_meter = (name_one.id + name_two.id) % 100
        love_idx = str(sorted(x for x in LOVE_LEVELS if x <= love_meter)[-1])
        love_status = choice(LOVE_DATA[love_idx]["titles"])
```

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        embed.add_field(
+            name='A letter from Dr. Love:',
+            value=LOVE_TXT[love_stat_i]
+        )
+
+        await ctx.message.channel.send(embed=embed)
+
+    @love.error
+    async def love_error(self, ctx, error):
```
scragly: This command-specific error handler eats all uncaught errors here. It's better if it was removed, as this block is only logging two exceptions and ignoring the rest, when by default we handle missing args already with user feedback and logging in the default error handler elsewhere.

bot/seasons/valentines/lovecalculator.py
darthdelay: Is a cooldown of 5 seconds per User enough or should I raise that?

bot/seasons/valentines/lovecalculator.py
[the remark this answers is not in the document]
darthdelay: Done.

bot/seasons/valentines/lovecalculator.py
darthdelay: Removed.

bot/seasons/valentines/lovecalculator.py
darthdelay: Implemented as is. I couldn't think of anything better o.o

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,119 @@
...
+         "other to increase the chances of this relationship's survival "
+}
+
+
+class LoveCalculator:
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command(aliases=('love_calculator', 'love_calc'))
+    async def love(self, ctx, name_one: discord.Member, name_two: discord.Member):
```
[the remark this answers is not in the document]
darthdelay: Done. I made it so that name_two is None by default, and added some logic afterwards as wanted c:

bot/seasons/valentines/lovecalculator.py
darthdelay: Error handlers have been removed due to https://github.com/python-discord/seasonalbot/pull/100#discussion_r255285628

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,71 @@
...
+class LoveCalculator:
+    """
+    A cog for calculating the love between two people
+    """
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command(aliases=('love_calculator', 'love_calc'))
+    @commands.cooldown(rate=1, per=5.0, type=commands.BucketType.user)
+    async def love(self, ctx, name_one: discord.Member, name_two=None):
```
scragly: `name_two` should still be a converter, as it'll only convert when an arg is passed to it.

```suggestion
    async def love(self, ctx, name_one: discord.Member, name_two: discord.Member = None):
```

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,71 @@
...
+                .love Ves Zappa Niko Laus -> Will not work
+                .love "Ves Zappa" "Niko Laus" -> Will work
+        """
+
+        if name_two is None:
+            staff = ctx.guild.get_role(Roles().helpers).members
+            name_two = choice(staff)
+        else:
+            name_two = await commands.MemberConverter().convert(ctx, name_two)
+            print(name_two)
```
scragly: With the converter added to the command signature, this `else` block can be removed.
```diff
-        else:
-            name_two = await commands.MemberConverter().convert(ctx, name_two)
-            print(name_two)
```

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,71 @@
...
+
+        with open(Path("bot", "resources", "valentines", "love_matches.json"), "r") as file:
+            LOVE_DATA = json.load(file)
+        LOVE_LEVELS = [int(x) for x in LOVE_DATA]
```
scragly: The open and loading of JSON data should be done once at the load of the extension, so this should be placed at the top of the module after the log definition and before the cog class definition (around line 13).
```py

with open(Path("bot", "resources", "valentines", "love_matches.json"), "r") as file:
    LOVE_DATA = json.load(file)
LOVE_LEVELS = [int(x) for x in LOVE_DATA]


class LoveCalculator:
```

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,71 @@
...
+        Using method 1-4 is highly encouraged, as nobody likes unwanted pings + they'll count as spam.
+        Skipping the second name will lead to something awesome.
+
+        *(1): If the name has any form of spacing, the name must be wrapped inside quotes. Example:
+                .love Ves Zappa Niko Laus -> Will not work
+                .love "Ves Zappa" "Niko Laus" -> Will work
+        """
+
+        if name_two is None:
+            staff = ctx.guild.get_role(Roles().helpers).members
```
scragly: An instance of Roles doesn't need to be made.

```suggestion
            staff = ctx.guild.get_role(Roles.helpers).members
```

bot/seasons/valentines/lovecalculator.py
```diff
@@ -0,0 +1,71 @@
...
+            LOVE_DATA = json.load(file)
+        LOVE_LEVELS = [int(x) for x in LOVE_DATA]
+
+        love_meter = (name_one.id + name_two.id) % 100
+        love_idx = str(sorted(x for x in LOVE_LEVELS if x <= love_meter)[-1])
+        love_status = choice(LOVE_DATA[love_idx]["titles"])
+
+        embed = discord.Embed(
+            title=love_status,
+            description=f'{name_one.display_name} \u2764 {name_two.display_name} scored {love_meter}%!\n\u200b',
```
scragly: We are adjusting the project settings to use 100 character lines in the near future. As just a convenience, could we adjust this down now to avoid someone having to adjust it later?
```suggestion
            description=(
                f'{name_one.display_name} \u2764 {name_two.display_name} '
                f'scored {love_meter}%!\n\u200b'
            ),
```

bot/seasons/valentines/lovecalculator.py
darthdelay: Added. I didn't even know this was possible o.o

bot/seasons/valentines/lovecalculator.py
darthdelay: Sure thing. Done

darthdelay:
Should be all working now. Tested it a bunch the past days.

PR #111, opened 2019-02-12
Readded fun.py file to evergreen and added roll command.

### LINKED ISSUE #93: Write a roll function for the Fun cog
**Originally posted by Gareth Coles:**

Originally submitted by **Automation** to ClickUp:

> Write a bot command for the Fun cog, bot.roll, which returns a random terningkast ie `<:terning3:431249694467948544>`

---

**Volcyy** commented:

> drummeur (not yet on ClickUp) wanted to do this.

astieman:
Closes #93.

Readded fun.py file to evergreen and added roll command that rolls up to 6 dice randomly and uses the terning emotes on the discord.

### DIFF
--- a/bot/seasons/evergreen/fun.py
+++ b/bot/seasons/evergreen/fun.py
@@ -16,6 +16,9 @@ class Fun:
 
     @commands.command()
     async def roll(self, ctx, num_rolls: int = 1):
+        """
+            Outputs a number of random dice emotes (up to 6)
+        """
         output = ""
         if num_rolls > 6:
             num_rolls = 6
--- a/bot/seasons/evergreen/fun.py
+++ b/bot/seasons/evergreen/fun.py
@@ -1,0 +1,40 @@
+import logging
+import random
+
+from discord.ext import commands
+
+log = logging.getLogger(__name__)
+
+
+class Fun:
+    """
+    A collection of general commands for fun.
+    """
+
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command()
+    async def roll(self, ctx, str_input: str = None):
+        output = ""
+
+        if not str_input:
+            output = "To use .roll, format it as such: .roll (number)"
+        else:
+            try:
+                num_rolls = int(str_input)
+                if num_rolls > 6:
+                    num_rolls = 6
+                elif num_rolls < 1:
+                    return
+                for i in range(num_rolls):
+                    output += ":terning%d: " % random.randint(1, 6)
+            except ValueError:
+                return
+        await ctx.send(output)
+
+
+# Required in order to load the cog, use the class name in the add_cog function.
+def setup(bot):
+    bot.add_cog(Fun(bot))
+    log.debug("Fun cog loaded")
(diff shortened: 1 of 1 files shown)


bot/seasons/evergreen/fun.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @commands.command()
+    async def roll(self, ctx, str_input: str = None):
+        output = ""
+
+        if not str_input:
+            output = "To use .roll, format it as such: .roll (number)"
+        else:
+            try:
+                num_rolls = int(str_input)
```
MarkKoz: Discord.py supports type [converters](https://discordpy.readthedocs.io/en/rewrite/ext/commands/commands.html#converters) for command parameters. Instead of taking a string and converting it yourself, at the least, you should be annotating the parameter as an `int`.

However, since you wish to restrict the range to be between 1 and 6, you should write a custom type converter. Implementing it that way allows for it to integrate nicely with other parts of the library (such as the error handling).

bot/seasons/evergreen/fun.py
```diff
@@ -0,0 +1,40 @@
...
+                if num_rolls > 6:
+                    num_rolls = 6
+                elif num_rolls < 1:
+                    return
+                for i in range(num_rolls):
```
scragly: Due to `i` not being used, it should either be named `_` or `_i` to signify it's an intentionally unused variable. Currently the linter is failing on this.

heavysaturn: (we tend to prefer `_`)

```suggestion
                for _ in range(num_rolls):
```

bot/seasons/evergreen/fun.py
```diff
@@ -15,22 +15,14 @@ def __init__(self, bot):
...
-                    num_rolls = 6
-                elif num_rolls < 1:
-                    return
-                for i in range(num_rolls):
-                    output += ":terning%d: " % random.randint(1, 6)
-            except ValueError:
-                return
+        if num_rolls > 6:
+            num_rolls = 6
+        elif num_rolls < 0:
```
MarkKoz: Shouldn't this still be `< 1`?

bot/seasons/evergreen/fun.py
```diff
@@ -0,0 +1,32 @@
...
+class Fun:
+    """
+    A collection of general commands for fun.
+    """
+
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command()
+    async def roll(self, ctx, num_rolls: int):
```
scragly: It would be good if there was a default value of 1. That way uses who don't present an argument get a plain single roll. Feels like something that would happen so it'll improve user experience.

```suggestion
    async def roll(self, ctx, num_rolls: int = 1):
```

bot/seasons/evergreen/fun.py
astieman: I was over engineering it and thought "well maybe they don't want to roll anything" and then realized I'm just being obtuse.

PR #112, opened 2019-02-12
Added command .bemyvalentine

### LINKED ISSUE #75: [Feature Request] Be My Valentine
As is traditional on Valentine's day, users can request that the bot send a Valentine to another user. This can be something cheesy, something sweet, or even a poem! The command should be able to send a user-specified valentine, or pick one randomly from a set if one is not provided. This command should have a cooldown.

Another possible option could allow a user to DM a valentine to be given anonymously.

RohanJnr:
Closes #75 

this pull request will add in a new command ".bemyvalentine".
This command sends valentine to a user if specified or else a random user having a role `lovefest`.( i did use the lovefest role id from the costants file).
There is even a subcommand where the person who will be sending the valentine will be marked anonymous.
There are 3 types of messages :
1) Poems.
2) Compliments (good messages).
3) Custom messages which can be written by the user himself.
I have also added a 30min cool down for the commands to avoid spam.
I have written a detailed help command which can be viewed by ".help bemyvalentine" and ".help bemyvalentine dm".
This is Iceman#6508 in Discord.


RohanJnr:
just another commit before review.

RohanJnr:
some of the changes which i forgot to mention in my commit message :
1) The anonymous bemyvalentine command is only DM specific.
2) The non-anonymous bemyvalentine command can only be used in the server.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
+import logging
+import random
+import typing
+from json import load
+from pathlib import Path
+
+
```
scragly: Import groups should be separated by a single empty line.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    id = Lovefest.role_id
+
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command(name="lovefest")
+    async def add_role(self, ctx):
+        """
+            This command adds people to the lovefest role
```
scragly: This should be the same indent as the quotation marks.
It should probably also end in a period.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+        embed = discord.Embed(
+            title="USER_LOVEFEST list updated!",
+            description=f'''The USER_LOVEFEST has been refreshed,`bemyvalentine` and `bemyvalentine dm` commands can now
+                            be used and there are {USER_LOVEFEST.__len__()} members having the lovefest role.''',
+            color=Colours.pink
+        )
+        await ctx.send(embed=embed)
+
+    @commands.cooldown(1, 1800, BucketType.user)
+    @commands.group(
```
scragly: It's unnecessary for this to be spread across multiple lines; it's not an unreasonably long line if it's formatted as a single one.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+        name='bemyvalentine',
+        invoke_without_command=True
+    )
+    async def send_valentine(self, ctx, user: typing.Optional[discord.Member] = None, *, valentine_type=None):
+        """
+        This command sends valentine to user if specified or a random user having lovefest role.
+
+
+        syntax: .bemyvalentine [user](optional) [p/poem/c/compliment/or you can type your own valentine message]
```
scragly: Avoid double newlines where possible in help output.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+        (optional)
+
+        example: .bemyvalentine (sends valentine as a poem or a compliment to a random user)
+
+        example: .bemyvalentine @Iceman#6508 p (sends a poem to Iceman)
+
+        example: .bemyvalentine @Iceman#6508 Hey I love you, wanna hang around ? (sends the custom message to Iceman)
```
scragly: The examples don't really need the empty lines between them.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+                )
+
+            embed.description = f'{valentine} \n **{EMOJI_2}From {ctx.author}{EMOJI_1}**'
+            await channel.send(user.mention, embed=embed)
+
+    @commands.cooldown(1, 1800, BucketType.user)
+    @send_valentine.command(name='dm')
+    async def anonymous(self, ctx, user: typing.Optional[discord.Member] = None, *, valentine_type=None):
+        """
+    - This command DMs a valentine to be given anonymous to a user if specified or a random user having lovefest role.
```
scragly: The leading hyphen should be removed. The docstring contents need to align with the quotes indenting.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+    **This command should be DMed to the bot.**
+
+
```
scragly: Avoid double newlines in help output.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+    syntax : .bemyvalentine dm [user](optional) [p/poem/c/compliment/or you can type your own valentine message]
+    (optional)
+
+    example : .bemyvalentine dm (sends valentine as a poem or a compliment to a random user in DM making you anonymous)
+
+    example : .bemyvalentine dm Iceman#6508 p (sends a poem to Iceman in DM making you anonymous)
+
+    example : .bemyvalentine dm Iceman#6508 Hey I love you, wanna hang around ? (sends the custom message to Iceman in
```
scragly: The examples don't really need the empty lines between them.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+            embed.description = f'{valentine} \n **{EMOJI_2}From anonymous{EMOJI_1}**'
+            await ctx.author.send(f"Your message has been sent to {user}")
+            await user.send(embed=embed)
+
+    @staticmethod
+    def random_valentine():
+        """
+            grabs a random poem or a compliment (any message)
+        """
+        with open(Path('bot', 'resources', 'valentines', 'bemyvalentine_valentines.json'), 'r', encoding="utf8") as f:
```
scragly: Can we load the json data into a dict when the extension is loaded, so the file is only opened once, instead of on every command use. It'll speed things up in sacrifice of using some memory for storing it.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+            valentines = load(f)
+            valentine_poem = random.choice(valentines['valentine_poems'])
```
scragly: Once the file contents is loaded, any extra stuff can be dedented to exit the `open` context manager, closing the file access.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+                )
+
+            embed.description = f'{valentine} \n **{EMOJI_2}From anonymous{EMOJI_1}**'
+            await ctx.author.send(f"Your message has been sent to {user}")
+            await user.send(embed=embed)
+
+    @staticmethod
+    def random_valentine():
+        """
+            grabs a random poem or a compliment (any message)
```
scragly: Indent needs to be corrected, capitalise first word, and a period probably should be added to the end.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+            if random_valentine == valentine_poem:
+                message_type = 'A poem dedicated to'
+            else:
+                message_type = 'A compliment for '
+            return random_valentine['message'], message_type
+
+    @staticmethod
+    def valentine_poem():
+        """
+            grabs a random poem
```
scragly: Indent needs to be corrected, capitalise first word, and a period probably should be added to the end.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        with open(Path('bot', 'resources', 'valentines', 'bemyvalentine_valentines.json'), 'r', encoding="utf8") as f:
+            valentines = load(f)
+            valentine_poem = random.choice(valentines['valentine_poems'])
+            return valentine_poem['message']
+
+    @staticmethod
+    def valentine_compliment():
+        """
+            grabs a random compliment
```
scragly: Indent needs to be corrected, capitalise first word, and a period probably should be added to the end.

RohanJnr:
oh shoot ,i forgot to make the .refreshlovefest as mod only,lemme do it right away.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,228 @@
...
+from discord.ext import commands
+from discord.ext.commands.cooldowns import BucketType
+
+from bot.constants import Colours, Lovefest, Roles
+from bot.decorators import with_role
+
+log = logging.getLogger(__name__)
+
+HEART_EMOJIS = [":heart:", ":gift_heart:", ":revolving_hearts:", ":sparkling_heart:", ":two_hearts:"]
+EMOJI_1 = random.choice(HEART_EMOJIS)
```
sco1: If these emojis are defined at the module level they're never going to change unless the cog is reloaded. If you want to have them cycling for each valentine then it would probably be more appropriate to have a static method that returns a random emoji from the list.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    A cog that sends VALENTINES to other users !
+    """
+    id = Lovefest.role_id
+
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command(name="lovefest")
+    async def add_role(self, ctx):
```
sco1: This should be a command group with commands to both add and remove the role. Something like `.valentine subscribe` and `.valentine unsubscribe` or similar.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,228 @@
...
+        if id not in [role.id for role in ctx.message.author.roles]:
+            USER_LOVEFEST.append(ctx.author)
+            await user.add_roles(role)
+            await ctx.send("The Lovefest role has been added !")
+        else:
+            await ctx.send("You already have the role !")
+
+    @with_role(Roles.moderator)
+    @commands.command(name='refreshlovefest')
+    async def refresh_user_lovefestlist(self, ctx):
```
sco1: Can you provide some information on why this is needed as a command?

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,228 @@
...
+
+            if valentine_type is None:
+                # grabs a random valentine -can be a poem or a good message
+                valentine, title = self.random_valentine()
+                embed = discord.Embed(
+                    title=f"{EMOJI_1} {title} {user.display_name}{EMOJI_2}",
+                    color=Colours.pink
+                )
+
+            elif valentine_type in ['p', 'poem']:
```
sco1: Should be case insensitive

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,228 @@
...
+                valentine = self.valentine_poem()
+                embed = discord.Embed(
+                    title=f"{EMOJI_1} A poem dedicated to {user.display_name}{EMOJI_2}",
+                    color=Colours.pink
+                )
+
+            elif valentine_type in ['c', 'compliment']:
```
sco1: Should be case insensitive

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,228 @@
...
+            # Well a user cant valentine himself/herself.
+            await ctx.send('Come on dude, you cant send a valentine to yourself :expressionless:')
+
+        elif user is None:
+            # just making sure that the random dosent pick up the same user(ctx.author)
+            USER_LOVEFEST.remove(ctx.author)
+            user = random.choice(USER_LOVEFEST)
+            USER_LOVEFEST.append(ctx.author)
+
+            if valentine_type is None:
```
sco1: If both valentines commands share this block of code then there's probably a way it can be cleaned up into a separate function rather than having to copy/paste the same thing between the two.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,228 @@
...
+        elif user is None:
+            # just making sure that the random does not pick up the same user(ctx.author)
+            USER_LOVEFEST.remove(ctx.author)
+            user = random.choice(USER_LOVEFEST)
+            USER_LOVEFEST.append(ctx.author)
+
+            if valentine_type is None:
+                # grabs a random valentine -can be a poem or a good message
+                valentine, title = self.random_valentine()
+                embed = discord.Embed(
```
sco1: This embed generation doesn't need to be repeated for every conditional. The `f"{EMOJI_1} {title} {user.display_name}{EMOJI_2}"` format is exactly the same across them, so you can instead generate `title` inside of the conditions and just have a single embed statement after the logic block.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,228 @@
...
+from bot.constants import Colours, Lovefest, Roles
+from bot.decorators import with_role
+
+log = logging.getLogger(__name__)
+
+HEART_EMOJIS = [":heart:", ":gift_heart:", ":revolving_hearts:", ":sparkling_heart:", ":two_hearts:"]
+EMOJI_1 = random.choice(HEART_EMOJIS)
+EMOJI_2 = random.choice(HEART_EMOJIS)
+USER_LOVEFEST = []
+JSON_FILE = open(Path('bot', 'resources', 'VALENTINES', 'bemyvalentine_valentines.json'), 'r', encoding="utf8")
```
sco1: This shouldn't be left dangling open, load the contents into memory (in the cog's `__init__`) & close it.

bot/seasons/valentines/be_my_valentine.py
RohanJnr: well when people use .bemyvalentine (without providing a user to send a valentine to) ,then a random user having the lovefest role will be selected.So basically i have a list of people who have the lovefest role called USER_LOVEFEST.When people use the .lovefest command to get the lovefest role,then their name will get appended to this list.
But, INCASE the bot goes offline and comes back online,the list is lost, so thats why this .refreshlovefest

bot/seasons/valentines/be_my_valentine.py
RohanJnr: i did close it in the static methods  once i got what i need.

sco1: It shouldn't be left open at all.

RohanJnr: how do i load it into memory

sco1: Store it as an attribute of the cog when you `__init__`

RohanJnr: something like this :
JSON_FILE = open(Path('bot', 'resources', 'VALENTINES', 'bemyvalentine_valentines.json'), 'r', encoding="utf8")
VALENTINES = load(JSON_FILE)
bot.add_cog(BeMyValentine(bot, VALENTINES))
JSON_FILE.close()
log.debug("Be My Valentine cog loaded")

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,232 @@
...
+    def valentine_compliment(self):
+        """
+        Grabs a random compliment.
+        """
+        valentine_compliment = random.choice(self.valentines['valentine_compliments'])
+        return valentine_compliment['message']
+
+
+def setup(bot):
+    JSON_FILE = open(Path('bot', 'resources', 'valentines', 'bemyvalentine_valentines.json'), 'r', encoding="utf8")
```
sco1: The JSON load should be a method of your `BeMyValentine` and called as part of its `__init__`. 

You'll want to use a [context manager](https://docs.python.org/3/reference/compound_stmts.html#with) (`with open(...) as f:`) as well. This protects against errors during load and closes the file automatically when you're done with the file or if there is an error. You can see an example in the [Python 3 Tutorial](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files).

`Path` also provides an `Path.open()` method. You can see an example of this in [`pathlib`'s documentation](https://docs.python.org/3/library/pathlib.html#pathlib.Path.open)

bot/seasons/valentines/be_my_valentine.py
sco1: I don't see why this list of folks with the role is necessary. Checking [`Member.roles`](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.Member.roles) (probably with a helper method) accomplishes the same goal without needing to worry about maintaining a cache of users with the role in memory.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,232 @@
...
+    A cog that sends valentines to other users !
+    """
+    id = Lovefest.role_id
+
+    def __init__(self, bot, valentines):
+        self.bot = bot
+        self.valentines = valentines
+
+    @commands.group(name="lovefest", invoke_without_command=True)
+    async def add_role(self, ctx):
```
sco1: This needs to be its own command within the group, otherwise it's not going to be clear how to invoke the command (or what command is being invoked):

image

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,232 @@
...
+
+        embed = discord.Embed(
+            title=f'{emoji_1} {title} {user.display_name} {emoji_2}',
+            description=f'{valentine} \n **{emoji_2}From {ctx.author}{emoji_1}**',
+            color=Colours.pink
+        )
+        await channel.send(user.mention, embed=embed)
+
+    @commands.cooldown(1, 1800, BucketType.user)
+    @send_valentine.command(name='dm')
```
sco1: `'secret'` would probably be a better name

bot/seasons/valentines/be_my_valentine.py
RohanJnr: so i should use it whenever someone runs the command ? like instantly?

bot/seasons/valentines/be_my_valentine.py
RohanJnr: or lemme just make 2 seperate commands insted of a group?

bot/seasons/valentines/be_my_valentine.py
RohanJnr: like make a method that does the same thing but only when its needed?
like when someone uses the command without providing a users name

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,236 @@
...
+
+    @staticmethod
+    def load_json():
+        p = Path('bot', 'resources', 'valentines', 'bemyvalentine_valentines.json')
+        with p.open() as json_data:
+            valentines = load(json_data)
+            return valentines
+
+    @commands.group(name="lovefest", invoke_without_command=True)
+    async def lovefest_role(self, ctx):
```
sco1: This should just invoke the built-in help for the group. You can see an example of this in the AoC Cog:

https://github.com/python-discord/seasonalbot/blob/e2b175bbfdee41bc518f40e430af94e28396966d/bot/seasons/christmas/adventofcode.py#L137-L143

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,236 @@
...
+            valentine = valentine_type
+            title = f'A message for'
+
+        embed = discord.Embed(
+            title=f'{emoji_1}{title} {user.display_name}{emoji_2}',
+            description=f'{valentine} \n **{emoji_2}From anonymous{emoji_1}**',
+            color=Colours.pink
+        )
+        await ctx.author.send(f"Your message has been sent to {user}")
+        await user.send(embed=embed)
```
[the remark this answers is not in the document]
RohanJnr: So you mean that, the people who can receive valentine are those with the lovefest role? like even a person specifies a user while using the command,that user should have the lovefest role?

bot/seasons/valentines/be_my_valentine.py
RohanJnr: got it.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,236 @@
...
+
+    @staticmethod
+    def random_user(author, members):
```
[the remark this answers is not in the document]
RohanJnr: okay.

bot/seasons/valentines/be_my_valentine.py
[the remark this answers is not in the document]
sco1: Yes, otherwise people can spam random people in the server with DMs anonymously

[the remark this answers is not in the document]
RohanJnr: okay got it,working on it !

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+
+    @commands.group(name="lovefest", invoke_without_command=True)
+    async def lovefest_role(self, ctx):
+        """
+        You can have yourself the lovefest role or remove it.
+        The lovefest role makes you eligible to receive anonymous valentines from other users.
+
+        1) use the command \".lovefest sub\" to get the lovefest role.
+        2) use the command \".lovefest unsub\" to get rid of the lovefest role.
+            """
```
sco1: Unnecessary indent

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+        channel = self.bot.get_channel(Lovefest.channel_id)
+
+        if user is None:
+            author = ctx.author
+            user = self.random_user(author, Role.members)
+            if user is None:
+                return await ctx.send("There are no users avilable to whome your valentine can be sent.")
+            else:
+                pass
+            # just making sure that the random does not pick up the same user(ctx.author)
```
sco1: What does this comment refer to?

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+            if Lovefest.role_id not in [role.id for role in user.roles]:
+                message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
+                return await ctx.send(message)
+
+        if user == ctx.author:
+            # Well a user cant valentine himself/herself.
+            return await ctx.send('Come on dude, you cant send a valentine to yourself :expressionless:')
+
+        emoji_1, emoji_2 = self.random_emoji()
+        Role = discord.utils.get(ctx.guild.roles, id=Lovefest.role_id)
```
sco1: This needs a different, more specific name. `Role` would be assumed to reference the abstract `discord.Role` object rather than the object that represents the lovefest role.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+            title=f'{emoji_1}{title} {user.display_name}{emoji_2}',
+            description=f'{valentine} \n **{emoji_2}From anonymous{emoji_1}**',
+            color=Colours.pink
+        )
+        await ctx.author.send(f"Your message has been sent to {user}")
+        await user.send(embed=embed)
+
+    @staticmethod
+    def random_user(author, members):
+        members.remove(author)
```
sco1: This will throw an exception if the invoking user does not have the lovefest role

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+        Role = discord.utils.get(ctx.guild.roles, id=Lovefest.role_id)
+        channel = self.bot.get_channel(Lovefest.channel_id)
+
+        if user is None:
+            author = ctx.author
+            user = self.random_user(author, Role.members)
+            if user is None:
+                return await ctx.send("There are no users avilable to whome your valentine can be sent.")
+            else:
+                pass
```
jchristgit: This is superfluous

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+            title = f'A poem dedicated to'
+
+        elif valentine_type.lower() in ['c', 'compliment']:
+            valentine = self.valentine_compliment()
+            title = f'A compliment for'
+
+        else:
+            # in this case, the user decides to type his own valentine.
+            valentine = valentine_type
+            title = f'A message for'
```
jchristgit: All of these titles don't need to be f-strings

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+        Role = discord.utils.get(guild.roles, id=Lovefest.role_id)
+        print(Role)
+
+        if user is None:
+            author = ctx.author
+            user = self.random_user(author, Role.members)
+            if user is None:
+                return await ctx.send("There are no users avilable to whome your valentine can be sent.")
+            else:
+                pass
```
jchristgit: Superfluous

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+            title = f'A poem dedicated to'
+
+        elif valentine_type.lower() in ['c', 'compliment']:
+            valentine = self.valentine_compliment()
+            title = f'A compliment for'
+
+        else:
+            # in this case, the user decides to type his own valentine.
+            valentine = valentine_type
+            title = f'A message for'
```
jchristgit: I've seen this before. Would be cleaner to factor it out to a function

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+            description=f'{valentine} \n **{emoji_2}From anonymous{emoji_1}**',
+            color=Colours.pink
+        )
+        await ctx.author.send(f"Your message has been sent to {user}")
+        await user.send(embed=embed)
+
+    @staticmethod
+    def random_user(author, members):
+        members.remove(author)
+        if members.__len__() == 0:
```
jchristgit: Why not `if not members`?

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        valentine_poem = random.choice(self.valentines['valentine_poems'])
+        return valentine_poem['message']
+
+    def valentine_compliment(self):
+        """
+        Grabs a random compliment.
+        """
+        valentine_compliment = random.choice(self.valentines['valentine_compliments'])
+        return valentine_compliment['message']
```
jchristgit: I don't understand why all of these are dictionaries instead of strings if you're only ever using `message` anyways, or am I missing something?

RohanJnr: im using a json file.

bot/seasons/valentines/be_my_valentine.py
RohanJnr: Well,in the initial stages,people might just do .bemyvalentine before evening thinking of the lovefest role.

bot/seasons/valentines/be_my_valentine.py
RohanJnr: Seriously? that is just 2 LINES ! No need for a separate function, I am not doing anything major also there.

bot/seasons/valentines/be_my_valentine.py
jchristgit: sorry, what? `else: pass` is no-op

bot/seasons/valentines/be_my_valentine.py
jchristgit: I'm referring to the entire `if valentine_type.lower` in ['x', 'y']` parts.

bot/seasons/valentines/be_my_valentine.py
jchristgit: Yes, but what does that have to do with this?

bot/seasons/valentines/be_my_valentine.py
RohanJnr: oh okay,lol,i thought u were refering to the entier if

bot/seasons/valentines/be_my_valentine.py
RohanJnr: ?

jchristgit: In the JSON file you have objects but all of them are just in the format
```json
{
    "message": "something"
}
```
Why not have the list containing these objects contain the message directly instead?

bot/seasons/valentines/be_my_valentine.py
jchristgit: this wandered to line 191 due to some other changes

bot/seasons/valentines/be_my_valentine.py
RohanJnr: oh okay,got it.

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,235 @@
...
+            if user is None:
+                return await ctx.send("There are no users avilable to whome your valentine can be sent.")
+
+        embed = discord.Embed(
+            title=f'{emoji_1}{title} {user.display_name}{emoji_2}',
+            description=f'{valentine} \n **{emoji_2}From anonymous{emoji_1}**',
+            color=Colours.pink
+        )
+        await ctx.author.send(f"Your message has been sent to {user}")
+        await user.send(embed=embed)
```
scragly: Some users have DMs disabled in their privacy settings. This will result in the command failing to send anything, but the message indicates that it was sent successfully before the attempt happens.

Best thing to do here is to try sending it to the user first, and wrap a `try: except discord.Forbidden` around it to catch the case where it fails.

```py
        try:
            await user.send(embed=embed)
        except discord.Forbidden:
            await ctx.author.send(f"{user} has DMs disabled, so I couldn't send the message. Sorry!")
        else:
            await ctx.author.send(f"Your message has been sent to {user}")
```

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,235 @@
...
+            title = 'A message for'
+        return valentine, title
+
+    @staticmethod
+    def random_user(author, members):
+        # just making sure that the random does not pick up the same user(ctx.author)
+
+        if members.__len__() != 0:
+            if author in members:
+                members.remove(author)
```
scragly: The length check is done before removing the original author, so if the author is the only person with the role, it'll pass the if check, but then return an empty list, later raising an exception since `random.choice` fails with empty collections.

Check membership first, and then check length after removing them if they're present. This ensures it'll be seen as no valid members to receive it, and avoids `random.choice` failing.

```py
    @staticmethod
    def random_user(author, members):
        # just making sure that the random does not pick up the same user(ctx.author)
        if author in members:
            members.remove(author)
        if len(members) != 0:
            user = random.choice(members)
        else:
            user = None
        return user
```

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,235 @@
...
+        Grabs a random poem or a compliment (any message).
+        """
+        valentine_poem = random.choice(self.valentines['valentine_poems'])
+        valentine_compliment = random.choice(self.valentines['valentine_compliments'])
+        random_valentine = random.choice([valentine_compliment, valentine_poem])
+        if random_valentine == valentine_poem:
+            title = 'A poem dedicated to'
+        else:
+            title = 'A compliment for '
+        return random_valentine['message'], title
```
scragly: As the structure of the json data has changed, your `random_valentine` is actually already the message you wanted. Trying to fetch the value of a `message` key here is actually attempting to get it from a string value, and raises an exception, making the command fail. Best thing to do here is just remove that bit:
```suggestion
        return random_valentine, title
```

bot/resources/valentines/bemyvalentine_valentines.json
```diff
@@ -0,0 +1,45 @@
...
+    "When I say I love you, please believe it's true.\nWhen I say forever, know I'll never leave you.\nWhen I say goodbye, promise me you won't cry,\nBecause the day I'll be saying that will be the day I die.",
+    "Beauty isn't seen by eyes.\nIt's felt by hearts,\nRecognized by souls,\nIn the presence of love.",
+    "L is for \"laughter\" we had along the way.\nO is for \"optimism\" you gave me every day.\nV is for \"value\" of being my best friend.\nE is for \"eternity,\" a love that has no end.",
+    "If roses were red and violets could be blue,\nI'd take us away to a place just for two.\nYou'd see my true colors and all that I felt.\nI'd see that you could love me and nobody else.\nWe'd build ourselves a fortress where we'd run and play.\nYou'd be mine and I'd be yours till our dying day.\nThen I wake and realize you were never here.\nIt's all just my thoughts, my dreams, my hopes...\nBut now it's only tears!"
+
+  ],
+  "valentine_compliments": [
+
+    "To the love of my life. I’m the luckiest person on Earth because I have you! I adore you!You’ve taught me the best thing to hold onto in life is each other.You are my sweetheart, you are my life, you are my everything.",
```
AnonGuy:
```suggestion
    "To the love of my life. I’m the luckiest person on Earth because I have you! I adore you! You’ve taught me that the best thing to hold onto in life is each other. You are my sweetheart, you are my life, you are my everything.",
```

bot/resources/valentines/bemyvalentine_valentines.json
```diff
@@ -0,0 +1,45 @@
...
+    "It was a million tiny little things that, when you added them all up, they meant we were supposed to be together.",
```
AnonGuy:
```suggestion
    "There were a million tiny little things that, when added up, meant we were supposed to be together.",
```

bot/resources/valentines/bemyvalentine_valentines.json
```diff
@@ -0,0 +1,45 @@
...
+    "when you smile, the whole world stops and stares for a while, cause you're amazing, just the way you are",
```
AnonGuy:
```suggestion
    "When you smile, the whole world stops and stares for a while, cause you're amazing, just the way you are.",
```

bot/resources/valentines/bemyvalentine_valentines.json
```diff
@@ -0,0 +1,45 @@
...
+    "Take love, multiply it by infinity and take it to the depths of forever... and you still have only a glimpse of how I feel for you.",
+    "When you kiss me, the world just melts away. When you hold me in your arms, I feel safe. Being in love with you has made me see how wonderful life can be. I love you. ",
```
AnonGuy:
```suggestion
    "When you kiss me, the world just melts away. When you hold me in your arms, I feel safe. Being in love with you has made me see how wonderful life can be. I love you.",
```

bot/resources/valentines/bemyvalentine_valentines.json
```diff
@@ -0,0 +1,45 @@
...
+    "No matter how dark the days get, you still know how to make me smile and Even after all this time, you still take my breath away. ",
```
AnonGuy:
```suggestion
    "No matter how dark the days get, you still know how to make me smile. Even after all this time, you still take my breath away.",
```

bot/resources/valentines/bemyvalentine_valentines.json
```diff
@@ -0,0 +1,45 @@
...
+    "I don't know what my future holds but I know I want you in it but I would travel to the moon and back just to see you smile.",
```
AnonGuy:
```suggestion
    "I don't know what my future holds, but I know I want you in it. I would travel to the moon and back just to see you smile.",
```

bot/resources/valentines/bemyvalentine_valentines.json
```diff
@@ -0,0 +1,45 @@
...
+    "I may not always say it but know that With every breath in my body and every beat of my heart I know I will always love you. ",
```
AnonGuy:
```suggestion
    "I may not always say it, but know that with every breath in my body and every beat of my heart I know I will always love you.",
```

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,244 @@
...
+            msg = "You are supposed to use this command in the server."
+            return await ctx.send(msg)
+
+        if user:
+            if Lovefest.role_id not in [role.id for role in user.roles]:
+                message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
+                return await ctx.send(message)
+
+        if user == ctx.author:
+            # Well a user cant valentine himself/herself.
```
AnonGuy:
```suggestion
            # Well a user can't valentine himself/herself.
```

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -0,0 +1,244 @@
...
+            return await ctx.send('Come on dude, you cant send a valentine to yourself :expressionless:')
```
AnonGuy:
```suggestion
            return await ctx.send("Come on dude, you can't send a valentine to yourself :expressionless:")
```

bot/resources/valentines/bemyvalentine_valentines.json
scragly: This is a quote. It exists in the correct form.

bot/resources/valentines/bemyvalentine_valentines.json
scragly: This is a quote again, so the dropping of "that" is acceptable.

AnonGuy:
Besides those minor clerical errors, this PR looks good to me!

bot/resources/valentines/bemyvalentine_valentines.json
AnonGuy: Ah, okay

scragly:
99% @RohanJnr's work. Hope to see them more in the repo here and maybe in other projects we have :+1:

PR #116, opened 2019-02-15
[WIP] Remove Pipfiles and add pyproject.toml

AnonGuy:
This PR removes the files required by `pipenv` and adds `poetry`'s requirements instead.
Using poetry will decrease our build time significantly, and make it much easier to publish projects to PyPi.


poetry.lock
```diff
@@ -0,0 +1,455 @@
+[[package]]
```
SebastiaanZ: Just testing if review works.

sco1:
What is keeping this WIP?

AnonGuy:
@sco1 I need to add the appropriate changes to `azure-pipelines.yml`. I'll see if I can get this done tonight.

MarkKoz:
Any ideas on this? Why are we changing the pip cache directory anyway?

These warnings also appear on master.

```
The directory '/home/vsts/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
The directory '/home/vsts/.cache/pip' or its parent directory is not owned by the current user and caching wheels has been disabled. check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
```

MarkKoz:
As it is, the docker image builds and runs just fine. One minor thing is a suggestion from `fuzzywuzzy` to install `python-Levenshtein`. Don't know if we care enough to bother. If we do, we should consider making it extra (optional) dependency.

I've experimented with building a wheel using `poetry build -f wheel` and then installing it in the docker image. However, when installing the wheel, it doesn't respect VCS dependencies and fetches discord.py from PyPI. The same behaviour is exhibited if sdist is used (and it takes way longer to install anyway).

Apart from this dependency issue, it's pretty easy to use a wheel in the image. If it worked, I think it would speed up building the image since poetry wouldn't have to be installed. The time saved doesn't seem to be huge though.

heavysaturn:
Levenshtein is really slow to install, which is why we decided against it. build times are significantly increased by its inclusion, and its only used for an incredibly rarely used snake feature.

MarkKoz:
I have some concerns regarding dependencies. We don't actually run the bot in the Lint & Test job; we only run flake8. How many of the base packages are actually needed to install the Python dependencies? Some examples which don't require them to build:

* PIL will install fine but the bot will crash if `ImageFont` is imported and libfreetype isn't installed.
* lxml (along with its dependencies libxml2 and xlst) is an extra dependency of bs4. Our code base only seems to make use of `html.parser` so the dependencies are not needed. They aren't in the Dockerfile anyway.

And there are some other packages which I can't figure out the purpose of:

* curl
* zlib

GhostofGoes:
Definitely ditch lxml, it's not being used anywhere. You can probably remove git as well, since I think discord.py is installed via http.

sco1:
Git is needed for git dependencies, Poetry farms out to git with a subprocess call

MarkKoz:
I'm taking a stab at updating the readme, but Poetry has some differences from pipenv that make this problematic:

#### Installation of Poetry
The documentation for installation recommends using their custom installer, which is retrieved via `curl`. Windows users don't have `curl` by default. There is an alternate installation section which installs it with `pip install --user poetry`, but it doesn't explain what a user install is. The concern here is that the user site may not be on the PATH.

#### Run Scripts
Poetry's scripts section isn't the same as pipenv's. Rather, it's more akin to `setup.py`'s scripts. This isn't very useful for us; I don't think it can even run modules directly. A suggestion here https://github.com/sdispater/poetry/issues/241#issuecomment-470200353 is to use a makefile. A makefile is an interesting suggestion, but it'd add `make` as yet another thing contributors have to install (more of an annoyance on Windows if anything). Another idea pointed out by Joseph was https://github.com/sdispater/poetry/issues/241#issuecomment-445434646.

#### `.env` Files
Poetry does not support `.env` files. There's some some stuff with bash or `env` that can take care of this, but it's quite cumbersome. We'd need a cross-platform solution anyway. One idea is to have `__main__.py` try to load the `.env` file if it exists using [python-dotenv](https://github.com/theskumar/python-dotenv).

#### No PyCharm Support
PyCharm will not automatically create a virtual environment from the Poetry files. However, I believe it will automatically detect an existing virtual environment and automatically add it as an interpreter. It probably only looks in the project's directory so `poetry config settings.virtualenvs.in-project true` is needed. The guide would need to be shuffled around a bit so that the virtual environment is made before the project is opened in PyCharm.

sco1:
I understand the logic for not recommending the pip install but ultimately it boils down to the same pitfall as using the wrong Python environment's `pip` version to install any other module. I don't know if there's a scenario where a successful `pip install --user` install would fail to provide the correct entry point. Regardless, I've non-user installed it and it works just fine.

Run scripts are definitely an annoying one, for the sake of making folks' lives easier we'll just have to toy around with the proposed alternatives until the point where Poetry provides first class support (if it happens).

Loss of `.env` support seems well mitigated by `python-dotenv`.

From seeing some recent issues folks have been having with PyCharm's pipenv support I don't know if the loss of support here is too big of a deal. I don't use PyCharm so I can't attest but it seems like its support can be spotty. Since Poetry does not install to the typical environments folder, users will have to pay attention to the `settings.virtualenvs.in-project` configuration option, which introduces a pain point, but it shouldn't be too terrible to resolve if someone isn't paying attention and runs into it. Unless poetry respects a project-specific `config.toml`? The docs are unclear.

MarkKoz:
> I don't know if there's a scenario where a successful pip install --user install would fail to provide the correct entry point. 

I had to add the user site to `PATH` on my machine. The path of least resistance for less experience contributors would be a non-user pip install, but I think that's worth coming to a consensus on before moving forward.

> Unless poetry respects a project-specific config.toml

It doesn't seem to, unfortunately.

We need to come to some decisions so this PR can move along. My proposals:
* Create a Python script to delegate launch options. It's technically abuse of the scripts feature of Poetry but screw it.
* Use python-dotenv to load a `.env` file if one exists.

Still undecided on how to install Poetry.

sco1:
Sorry, my first paragraph was bad and wrong

MarkKoz:
Sorry this half turned into a CI revamp PR. Updates on the situation:

#### Azure Pipelines
At this point, I think I've done all I can to clean things up. The Lint & Test job has been sped up by almost 40 seconds. Both jobs in total, without actually building the image, have been sped up by over 1.5 minutes. Unfortunately, this has all been due to clean up, not because of Poetry; Poetry is actually about twice as slow to install dependencies compared to pipenv.

* All apt-get packages were removed. They were either not used or already came with the Ubuntu agent from MS. Removing the redundant docker especially saved time.
* setuptools and salt-pepper were removed, the former of which took especially long to install. They were remnants of a legacy deployment system.

#### Docker Image
The image's size has been reduced by over 200 MB; it's currently at about 140 MB. This is largely thanks to disabling caches and removing build dependencies once done building. There are some important changes to note, however:

* At least Docker Engine 18.06 is required now.
* The base image was updated to `python:3.7.2-alpine3.9`.

There's more info on these changes in the commit messages. These two changes are actually for quite minor reasons, so I am not against rolling them back if someone has a problem with them.

MarkKoz:
> The downside to going with a non-user install is it could potentially require sudo/admin permissions.

@sco1 Good point. Maybe we should go with a user install and leave a "note" about adding the user site to `PATH` if `poetry` isn't a recognised command. If so, how extensive would the instructions in this "note" be? We could just model it off pipenv's docs. After all, that's what we were relying on before.

MarkKoz:
I think it may have been a mistake to cut down on the intermediate layers so much. It looks like `docker push/pull` only download layers that have changed rather than the whole image. Will have to see what the size of a layer for just poetry installing the dependencies is.

MarkKoz:
We've decided to stick with pipenv after all.

PR #118, opened 2019-02-15
Refine love calculator command

avayert:
Things that have changed
- Now accepts arbitrary input in addition to members
  - And therefore changed the %-age calculating method too
- Escapes discord markdown for given arguments
- __Globally__ resets cooldown on commands upon user input errors (seemed a bit harsh)

\
Things that I have to finish before merge
- [x] The command help docstring

\
I opened the PR to get review for the actual code whilst I figure out the docstring for the command.

### DIFF
--- a/bot/seasons/valentines/lovecalculator.py
+++ b/bot/seasons/valentines/lovecalculator.py
@@ -4,12 +4,12 @@ import json
 import logging
 import random
 from pathlib import Path
+from typing import Union
 
 import discord
 from discord import Member
 from discord.ext import commands
 from discord.ext.commands import clean_content
-from typing import Union
 
 from bot.constants import Roles
 
--- a/bot/seasons/valentines/lovecalculator.py
+++ b/bot/seasons/valentines/lovecalculator.py
@@ -9,7 +9,7 @@ from typing import Union
 import discord
 from discord import Member
 from discord.ext import commands
-from discord.ext.commands import clean_content
+from discord.ext.commands import clean_content, BadArgument
 
 from bot.constants import Roles
 
@@ -30,7 +30,11 @@ class LoveCalculator:
 
     @commands.command(aliases=('love_calculator', 'love_calc'))
     @commands.cooldown(rate=1, per=5, type=commands.BucketType.user)
-    async def love(self, ctx, who: Union[Member, str], whom: Union[Member, str] = None):
+    async def love(
+            self, ctx,
+            who: Union[Member, clean_content],
+            whom: Union[Member, clean_content] = None
+    ):
         """
         Tells you how much the two love each other.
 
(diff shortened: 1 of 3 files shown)


bot/seasons/valentines/lovecalculator.py
```diff
@@ -1,70 +1,94 @@
...
+        # if inputs were members, make sure to show name#discrim
+        # also make sure to escape all discord markdown to prevent mischief
+        cleaner = clean_content(escape_markdown=True, use_nicknames=False)
+        who, whom = [await cleaner.convert(ctx, str(arg)) for arg in (who, whom)]
+
+        # hash inputs to guarantee consistent results (hashing algorithm choice arbitrary)
+        #
+        # hashlib is used over the builtin hash() function
+        # to guarantee same result over multiple runtimes
+        m = hashlib.sha256(who.encode() + whom.encode())
```
sco1: For the `str` case you probably want to normalize them (punctuation, case, etc.) otherwise they're going to hash differently.

image

bot/seasons/valentines/lovecalculator.py
```diff
@@ -1,70 +1,94 @@
...
-
-        love_meter = (name_one.id + name_two.id) % 101
-        love_idx = str(sorted(x for x in LOVE_LEVELS if x <= love_meter)[-1])
-        love_status = choice(LOVE_DATA[love_idx]["titles"])
-
+            whom = random.choice(staff)
+
+        # if inputs were members, make sure to show name#discrim
+        # also make sure to escape all discord markdown to prevent mischief
+        cleaner = clean_content(escape_markdown=True, use_nicknames=False)
```
sco1: Assuming the intent is to eliminate spoiler tags, this doesn't clean `str` inputs.

image

avayert: works fine for me, are you sure your discord.py isn't out of date

sco1: Whoops

bot/seasons/valentines/lovecalculator.py
avayert: Normalizing punctuation seemed excessive, but I've pushed a change which somewhat normalizes case and checks for empty input

bot/seasons/valentines/lovecalculator.py
```diff
@@ -1,70 +1,105 @@
...
+from discord import Member
 from discord.ext import commands
+from discord.ext.commands import BadArgument, clean_content
 
 from bot.constants import Roles
 
 log = logging.getLogger(__name__)
 
-with open(Path("bot", "resources", "valentines", "love_matches.json"), "r") as file:
+with Path('bot', 'resources', 'valentines', 'love_matches.json').open() as file:
```
eivl: Lovely change <3

scragly:
As a post-merge note on this PR for those that were involved in it:

This PR added a bot method error handler when we already have a global error handler that the project relies on in `/bot/seasons/evergreen/error_handler.py`. This is loaded always and is where the cooldown reversion should have been placed instead.

The usage of cooldown reset also presents a very simple way to bypass the entire cooldown by purposely providing incorrect input. Instead, a token reversion should have been done that removes the single counter that was added by the incorrect command usage.

Both of these issues are going to be addressed in a commit within #132.

PR #120, opened 2019-02-17
added help message for .savethedate and edited the error_message for cooldown error

RohanJnr:
1)Added a help text to .savethedate
2)Edited the error message for cooldowns so that it shows not only seconds remaining but also minutes(if cooldowntime>60 seconds)

### DIFF
--- a/bot/seasons/evergreen/error_handler.py
+++ b/bot/seasons/evergreen/error_handler.py
@@ -44,10 +44,10 @@ class CommandErrorHandler:
             )
             seconds = error.retry_after
             remaining_minutes, remaining_seconds = divmod(seconds, 60)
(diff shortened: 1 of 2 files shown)


bot/seasons/evergreen/error_handler.py
```diff
@@ -42,9 +42,12 @@ def __init__(self, bot):
                 f"{ctx.author} called the command '{ctx.command}' "
                 "but they were on cooldown!"
             )
+            seconds = error.retry_after
+            remaining_minutes, remaining_seconds = divmod(seconds, 60)
+            error_message = f'{int(remaining_minutes)} minutes {math.ceil(remaining_seconds)} seconds'
```
heavysaturn: This is not an error message. Why did you call it that? Variables should be named something that logically corresponds with what they are.

PR #129, opened 2019-02-28
Add .myvalenstate

### LINKED ISSUE #108: [Feature request] myvalenstate
Which country or state would be more befitting to me for a beautiful valentine, if could go there?

I would like you to write a bot, that based on my tag name, would find a state with the most matching characters. And that state will be my (probably) next valentine vacation spot.  

## Implementation details

- [ ] Write the bot command !myvalenstate
- [ ] Return a state with the most matching character with tag name (plus some comment).

## Example
.myvalenstate
Your valenState is `Sweden`
Sweden is a Scandinavian nation with thousands of coastal islands and inland lakes, along with vast boreal forests and glaciated mountains.
image

darthdelay:
Closes #108

This pull request adds a beta version of the command myvalenstate.

This command sends an embed containing a users valenstate to the channel
it has been called in. The process used here is to first put the username
into a spaceless, lower case form and than compare it to the countries
listed under valenstates.json, put into the same form, using an implementation
of the Levenshtein algorithm.

This beta version already implements everything needed for the command to
work properly. The follwing things are missing, not meant for the future
full implementation or are subject to change if needed:

- Short comments (25 missing)
- name parameter in myvalenstate (not meant for staying/Debug)
- Object names in valenstates.json (subject to change) (1)
- Docstring for levenshtein (subject to change)
- Docstring for myvalenstate (subject to change)

(1) - The object names can be put into the right case to omit the operation
        operation on line 54 `lower_state = state.lower().replace(' ', '')`

### DIFF
--- a/bot/seasons/valentines/myvalenstate.py
+++ b/bot/seasons/valentines/myvalenstate.py
@@ -1,0 +1,80 @@
+import collections
+import json
+import logging
+from pathlib import Path
+from random import choice
+
+import discord
+from discord.ext import commands
+
+from bot.constants import Colours
+
+log = logging.getLogger(__name__)
+
+with open(Path('bot', 'resources', 'valentines', 'valenstates.json'), 'r') as file:
+    STATES = json.load(file)
+    # STATES.sort()
+
+
+class MyValenstate:
+    def __init__(self, bot):
+        self.bot = bot
+
+    def levenshtein(self, source, goal):
+        """
+        Calculates the Levenshtein Distance between source and goal.
+        """
+        if len(source) < len(goal):
+            return self.levenshtein(goal, source)
+        if len(source) == 0:
+            return len(goal)
+        if len(goal) == 0:
+            return len(source)
+
+        pre_row = list(range(0, len(source) + 1))
+        for i, source_c in enumerate(source):
+            cur_row = [i + 1]
+            for j, goal_c in enumerate(goal):
+                if source_c != goal_c:
+                    cur_row.append(min(pre_row[j], pre_row[j + 1], cur_row[j]) + 1)
+                else:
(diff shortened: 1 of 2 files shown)


bot/seasons/valentines/myvalenstate.py
```diff
@@ -0,0 +1,80 @@
...
+import discord
+from discord.ext import commands
+
+from bot.constants import Colours
+
+log = logging.getLogger(__name__)
+
+with open(Path('bot', 'resources', 'valentines', 'valenstates.json'), 'r') as file:
+    STATES = json.load(file)
+    # STATES.sort()
```
sco1: Don't need the comment

bot/seasons/valentines/myvalenstate.py
```diff
@@ -0,0 +1,80 @@
...
+                value=f"You have {len(matches)} more matches, these being {leftovers}."
+            )
+        embed.set_image(
+            url=STATES[valenstate]["flag"]
+        )
+        await ctx.channel.send(embed=embed)
+
+
+def setup(bot):
+    bot.add_cog(MyValenstate(bot))
```
sco1: Need a logging statement for the cog load

bot/seasons/valentines/myvalenstate.py
CyrolX: Done

bot/seasons/valentines/myvalenstate.py
CyrolX: Added as `log.debug("MyValenstate cog loaded")`

bot/seasons/valentines/myvalenstate.py
darthdelay: I dont know who this CyrolX is but he did a great job.

PR #135, opened 2019-03-07
Prevent SeasonalBot from changing server icon during off-season periods

### LINKED ISSUE #134: Different bot and server icon for evergreen
We now have a new icon for the SeasonalBot to use in lovely green colors. I did a terribly naive implementation of this so I could see it in action, but forgot that the bot would then change the server icon to green as well. 

So - we need to be able to have the evergreen icon for the bot and the standard blue icon for the server during off-season periods, but the same avatar and icon during seasonal events.

Please implement this soonish so we don't keep changing the server icon every time we merge something to seasonalbot :smile:

kosayoda:
Closes #134 

Prevents SeasonalBot from applying the evergreen server icon during off-season periods. Server icon
will change accordingly during seasonal events.

### DIFF
--- a/bot/seasons/evergreen/__init__.py
+++ b/bot/seasons/evergreen/__init__.py
@@ -2,4 +2,4 @@ from bot.seasons import SeasonBase
 
 
 class Evergreen(SeasonBase):
-    pass
+    bot_icon = "/logos/logo_seasonal/evergreen/logo_evergreen.png"
--- a/bot/seasons/season.py
+++ b/bot/seasons/season.py
@@ -334,8 +334,11 @@ class SeasonBase:
         if not Client.debug:
             log.info("Applying avatar.")
             await self.apply_avatar()
-            log.info("Applying server icon.")
-            await self.apply_server_icon()
+            if self.name != "evergreen":
+                log.info("Applying server icon.")
+                await self.apply_server_icon()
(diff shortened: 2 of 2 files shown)


kosayoda:
> The original server icon still needs to be applied on season change.
> 
> Perhaps a good method to handle bot-only avatar icons is to add a new attribute to the `SeasonBase` class of `bot_icon` or `avatar` and default it to None. The base class should have `icon` revert back to the original logo: `"/logos/logo_full/logo_full.png"` and the `bot_icon` attribute in the Evergreen class can be redefined with `"/logos/logo_seasonal/evergreen/logo_evergreen.png"`.
> 
> `get_icon` method will likely need to change to:
> 
> ```python
>     async def get_icon(self, avatar: bool = False) -> bytes:
>         ...
>         if avatar:
>             icon = self.bot_icon or self.icon
>         else:
>             icon = self.icon
>         full_url = base_url + icon
> ```
> 
> `apply_avatar` will then need to pass the `avatar=True` argument when it calls `get_icon`
> 
> The if evergreen condition will have be taken out of `load` to have it action properly, but it may be best to place the server application within the `if username_changed` block to have it only try change the server icon when there's been a legitimate season change.

Modified accordingly, let me know if it needs amendments.

PR #141, opened 2019-03-14
Spooky Rating

### LINKED ISSUE #54: [Feature Request] Spooky rating
This command would genrate a random spooky rating from a range of 1-100 [or any other range].It would play display a hallooween character picture according to the rating given [like witch,dracula,etc etc]

***Implementation details***
1. To build a function which would generate a random number in the range 1-100 [or any other range] which would resemble the spooky rating of the user.
2. The rating is then put up in a embed and displayed out.It will also grab a picture from the internet of a random hallooween character and post it.
[ 2nd part of the code is optional ⬆️ 
- [ ] Random numebr generation
- [ ] To look for free pics online for the 2nd code [optional]

Suhail6inkling:
---
name: Spooky Rating
about: A command that calculates how spooky someone is (for Hacktober)
Closes #54 

---

This adds a SpookyRating cog containing the `spookyrating` command, which rates a person based on how spooky they are (random number)

An embed with a small "whisper" and a percentage along with an appropriate image is displayed

Hope this is okay :D


bot/seasons/halloween/spookyrating.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Calculates the spooky rating of someone.
+
+        This command accepts an optional user as an argument, defaulting to the command executor.
+        Users are converted from:
+          - User ID
+          - Mention
+          - name#discrim
+          - name
+          - nickname
```
jchristgit: I don't think we need to document this. Otherwise, to stay consistent, we would need this on every function that takes a member as input.

Suhail6inkling: I mean I saw it on one already - I just copied and pasted it. I can remove it if necessary

jchristgit: Sorry, I missed that. But yeah, would be great if you could remove this.

bot/resources/halloween/spooky_rating.json
```diff
@@ -0,0 +1,47 @@
+{
+    "0": {
```
avayert: You'll have to make this -1 due to how `bisect` works or change the implementation, this bug is present in the other command too I just haven't gotten around to fixing it (yet).

I hope this example from the [bisect documentation page](https://docs.python.org/3/library/bisect.html#other-examples) illustrates the issue:
```py
>>> def grade(score, breakpoints=[60, 70, 80, 90], grades='FDCBA'):
...     i = bisect(breakpoints, score)
...     return grades[i]
...
>>> [grade(score) for score in [33, 99, 77, 70, 89, 90, 100]]
['F', 'A', 'C', 'C', 'B', 'A', 'A']
```

bot/seasons/halloween/spookyrating.py
```diff
@@ -0,0 +1,77 @@
...
+
+        Any user will always yield the same result, no matter who calls the command
+        """
+
+        if who is None:
+            who = ctx.author
+
+        # This ensures that the same result over multiple runtimes
+        random.seed(who.id)
```
avayert: This will change the _global_ seed, which means (since this command gets constants as its input) it will render other commands completely useless. Consider I used this command on myself, then used the terning command 3 times. Well if I used this command again the output of those dice rolls would be the exact same.

This can trivially be fixed by doing
```py
r = random.Random()
r.seed(...)
# use r here
```
You could even make `r` be a member of the class in `__init__` since it'll get seeded each call anyway. (could also come up with a better name?)

I don't know if this is actually a concern for the bot, but since the fix is trivial I decided to comment anyway.

jchristgit:
hello @avayert can you take a look at your discussions were they resolved?

Suhail6inkling:
Fixed the images! Sorry it took so long - I know other PRs were waiting for this one to finish!

Hope this is okay :D

PR #146, opened 2019-03-19
Implement flake8-docstrings

### LINKED ISSUE #137: Add flake8-docstrings
Documentation is the answer to the ultimate question of life, the universe, and everything.

Let's add [`flake8-docstrings`](https://github.com/PyCQA/flake8-docstrings) to our repository to enforce adding docstrings to our functions.

Since the default docstring linting this provides is fairly strict, we will instead implement a (TBD) subset of the [error codes](http://www.pydocstyle.org/en/stable/error_codes.html#grouping)

sco1:
Implement flake8 docstring linting, as described in #137 

Closes #137

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -16,6 +16,8 @@ __all__ = ('SeasonalBot',)
 
 
 class SeasonalBot(Bot):
+    """Base bot instance."""
+
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.http_session = ClientSession(
@@ -26,9 +28,7 @@ class SeasonalBot(Bot):
         )
 
     def load_extensions(self, exts: List[str]):
-        """
-        Unload all current cogs, then load in the ones passed into `cogs`
-        """
+        """Unload all current cogs, then load in the ones passed into `cogs`."""
 
         # Unload all cogs
         extensions = list(self.extensions.keys())
@@ -46,9 +46,8 @@ class SeasonalBot(Bot):
                 log.error(f'Failed to load extension {cog}: {repr(e)} {format_exc()}')
 
     async def send_log(self, title: str, details: str = None, *, icon: str = None):
-        """
-        Send an embed message to the devlog channel
-        """
+        """Send an embed message to the devlog channel."""
+
         devlog = self.get_channel(constants.Channels.devlog)
 
         if not devlog:
@@ -64,7 +63,8 @@ class SeasonalBot(Bot):
         await devlog.send(embed=embed)
 
     async def on_command_error(self, context, exception):
-        # Don't punish the user for getting the arguments wrong
+        """Check command errors for UserInputError and reset the cooldown if thrown."""
+
         if isinstance(exception, commands.UserInputError):
             context.command.reset_cooldown(context)
         else:
--- a/bot/pagination.py
+++ b/bot/pagination.py
@@ -57,7 +57,7 @@ class LinePaginator(Paginator):
         """
         Adds a line to the current page.
 
-        If the line exceeds the :attr:`max_size` then an exception is raised.
+        If the line exceeds the `max_size` then an exception is raised.
 
         Overrides the Paginator.add_line from inside discord.ext.commands in order to allow
         configuration of the maximum number of lines per page.
@@ -72,7 +72,7 @@ class LinePaginator(Paginator):
         Raises
         ------
         RuntimeError
-            The line was too big for the current :attr:`max_size`.
+            The line was too big for the current `max_size`.
         """
 
         if len(line) > self.max_size - len(self.prefix) - 2:
--- a/bot/pagination.py
+++ b/bot/pagination.py
@@ -20,8 +20,6 @@ log = logging.getLogger(__name__)
 class EmptyPaginatorEmbed(Exception):
     """Base Exception class for an empty paginator embed."""
 
-    pass
-
 
 class LinePaginator(Paginator):
     """
--- a/bot/pagination.py
+++ b/bot/pagination.py
@@ -18,6 +18,8 @@ log = logging.getLogger(__name__)
 
 
 class EmptyPaginatorEmbed(Exception):
+    """Base Exception class for an empty paginator embed."""
+
     pass
 
 
@@ -37,14 +39,13 @@ class LinePaginator(Paginator):
         The maximum amount of lines allowed in a page.
     """
 
-    def __init__(self, prefix='```', suffix='```',
-                 max_size=2000, max_lines=None):
+    def __init__(self, prefix='```', suffix='```', max_size=2000, max_lines=None):
         """
-        This function overrides the Paginator.__init__
-        from inside discord.ext.commands.
-        It overrides in order to allow us to configure
-        the maximum number of lines per page.
+        Overrides the Paginator.__init__ from inside discord.ext.commands.
+
+        Allows for configuration of the maximum number of lines per page.
         """
+
         self.prefix = prefix
         self.suffix = suffix
         self.max_size = max_size - len(suffix)
@@ -55,15 +56,13 @@ class LinePaginator(Paginator):
         self._pages = []
 
     def add_line(self, line='', *, empty=False):
-        """Adds a line to the current page.
+        """
+        Adds a line to the current page.
 
-        If the line exceeds the :attr:`max_size` then an exception
-        is raised.
+        If the line exceeds the :attr:`max_size` then an exception is raised.
 
-        This function overrides the Paginator.add_line
-        from inside discord.ext.commands.
-        It overrides in order to allow us to configure
-        the maximum number of lines per page.
+        Overrides the Paginator.add_line from inside discord.ext.commands in order to allow
+        configuration of the maximum number of lines per page.
 
         Parameters
         -----------
@@ -77,6 +76,7 @@ class LinePaginator(Paginator):
         RuntimeError
             The line was too big for the current :attr:`max_size`.
         """
+
         if len(line) > self.max_size - len(self.prefix) - 2:
             raise RuntimeError('Line exceeds maximum page size %s' % (self.max_size - len(self.prefix) - 2))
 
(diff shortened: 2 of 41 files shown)


bot/pagination.py
```diff
@@ -18,6 +18,8 @@
 
 
 class EmptyPaginatorEmbed(Exception):
+    """Base Exception class for an empty paginator embed."""
+
     pass
```
jchristgit: can we remove the `pass` here now with the new docstring?

bot/seasons/evergreen/snakes/snakes_cog.py
```diff
@@ -278,9 +276,8 @@ def _snakify(message):
         return message
 
     async def _fetch(self, session, url, params=None):
-        """
-        Asyncronous web request helper method.
-        """
+        """Asyncronous web request helper method."""
```
jchristgit: while here, might as well fix the typo:
```suggestion
        """Asynchronous web request helper method."""
```

bot/seasons/evergreen/snakes/snakes_cog.py
```diff
@@ -465,7 +465,7 @@ def predicate(reaction, user):
     @locked()
     async def antidote_command(self, ctx: Context):
         """
-        Antidote - Can you create the antivenom before the patient dies?
+        Antidote - Can you create the antivenom before the patient dies.
```
jchristgit: This doesn't really read well without a question mark. How about

```suggestion
        A game where you race to create an antivenom before the patient dies.
```

instead?

bot/seasons/evergreen/snakes/snakes_cog.py
```diff
@@ -1049,16 +1057,16 @@ def predicate(reaction_: Reaction, user_: Member):
...
+        """Invokes the help command for the Snakes Cog."""
+
         log.debug(f"{ctx.author} requested info about the snakes cog")
         return await ctx.invoke(self.bot.get_command("help"), "Snakes")
 
     @snakes_group.command(name='snakify')
     async def snakify_command(self, ctx: Context, *, message: str = None):
         """
-        How would I talk if I were a snake?
+        How would I talk if I were a snake.
```
jchristgit: This doesn't really read well either. How about something like
```suggestion
        Shows the given message in snake-speak.
```

sco1:
I'm leaning towards adding `D400` to the ignore list. Having additional punctuation options for the command briefs seems more pleasant. Alternatively, we could implement the brief in the decorator itself but that seems like more effort than it's worth when weighed against what we gain from enforcing a period.

bot/bot.py
```diff
@@ -26,9 +28,7 @@ def __init__(self, **kwargs):
         )
 
     def load_extensions(self, exts: List[str]):
-        """
-        Unload all current cogs, then load in the ones passed into `cogs`
-        """
+        """Unload all current cogs, then load in the ones passed into `cogs`."""
```
MarkKoz: The function has "extensions" in the name so it's more consistent to use "extensions" over "cogs".

```suggestion
        """Unload all current extensions, then load the given extensions."""
```

bot/pagination.py
```diff
@@ -55,15 +56,13 @@ def __init__(self, prefix='```', suffix='```',
         self._pages = []
 
     def add_line(self, line='', *, empty=False):
-        """Adds a line to the current page.
+        """
+        Adds a line to the current page.
 
-        If the line exceeds the :attr:`max_size` then an exception
-        is raised.
+        If the line exceeds the :attr:`max_size` then an exception is raised.
```
MarkKoz:
```suggestion
        If the line exceeds the max_size then an exception is raised.
```

bot/pagination.py
```diff
@@ -98,21 +98,26 @@ def add_line(self, line='', *, empty=False):
...
-        Use a paginator and set of reactions to provide pagination over a set of lines. The reactions are used to
-        switch page, or to finish with pagination.
-        When used, this will send a message using `ctx.send()` and apply a set of reactions to it. These reactions may
-        be used to change page, or to remove pagination from the message. Pagination will also be removed automatically
-        if no reaction is added for five minutes (300 seconds).
+        Use a paginator and set of reactions to provide pagination over a set of lines.
+
+        The reactions are used to switch page, or to finish with pagination.
+
+        When used, this will send a message using `ctx.send()` and apply a set of reactions to it.
```
MarkKoz:
```suggestion
        When used, this will send a message using ctx.send() and apply a set of reactions to it.
```

bot/pagination.py
```diff
@@ -339,16 +345,14 @@ def add_image(self, image: str = None) -> None:
...
+        Use a paginator and set of reactions to provide pagination over a set of title/image pairs.
 
-        When used, this will send a message using `ctx.send()` and
-        apply a set of reactions to it. These reactions may
-        be used to change page, or to remove pagination from the message.
+        The reactions are used to switch page, or to finish with pagination.
 
-        Note: Pagination will be removed automatically
-        if no reaction is added for five minutes (300 seconds).
+        When used, this will send a message using `ctx.send()` and apply a set of reactions to it.
```
MarkKoz:
```suggestion
        When used, this will send a message using ctx.send() and apply a set of reactions to it.
```

bot/seasons/evergreen/snakes/utils.py
```diff
@@ -117,43 +112,54 @@
...
-        than 4 octaves is rather excessive.
-        ``tile`` can be used to make a seamlessly tiling pattern.  For example:
+        """
+        Create a new Perlin noise factory in the given number of dimensions.
+
+        dimension should be an integer and at least 1.
+
+        More octaves create a foggier and more-detailed noise pattern.  More than 4 octaves is rather excessive.
+
+        ``tile`` can be used to make a seamlessly tiling pattern.
```
MarkKoz:
```suggestion
        tile can be used to make a seamlessly tiling pattern.
```

bot/seasons/evergreen/snakes/utils.py
```diff
@@ -117,43 +112,54 @@
...
             pnf = PerlinNoiseFactory(2, tile=(0, 3))
-        This will produce noise that tiles every 3 units vertically, but never
-        tiles horizontally.
-        If ``unbias`` is true, the smoothstep function will be applied to the
-        output before returning it, to counteract some of Perlin noise's
-        significant bias towards the center of its output range.
+
+        This will produce noise that tiles every 3 units vertically, but never tiles horizontally.
+
+        If ``unbias`` is true, the smoothstep function will be applied to the output before returning
```
MarkKoz:
```suggestion
        If unbias is true, the smoothstep function will be applied to the output before returning
```

bot/seasons/season.py
```diff
@@ -130,37 +118,36 @@ def end(cls) -> datetime.datetime:
...
         return f"New Season, {self.name_clean}!"
 
     async def get_icon(self, avatar: bool = False) -> bytes:
         """
-        Retrieves the icon image from the branding repository, using the
-        defined icon attribute for the season. If `avatar` is True, uses
-        optional bot-only avatar icon if present.
+        Retrieve the season's icon from the branding repository using the Season's icon attribute.
+
+        If `avatar` is True, uses optional bot-only avatar icon if present.
```
MarkKoz:
```suggestion
        If avatar is True, uses optional bot-only avatar icon if present.
```

bot/seasons/season.py
```diff
@@ -130,37 +118,36 @@ def end(cls) -> datetime.datetime:
...
 
-        The icon attribute must provide the url path, starting from the master
-        branch base url, including the starting slash:
-        `https://raw.githubusercontent.com/python-discord/branding/master`
+        The icon attribute must provide the url path, starting from the master branch base url,
+        including the starting slash.
+        e.g. `/logos/logo_seasonal/valentines/loved_up.png`
```
MarkKoz:
```suggestion
        e.g. /logos/logo_seasonal/valentines/loved_up.png
```

bot/seasons/season.py
```diff
@@ -175,8 +162,9 @@ def greeting(self) -> str:
 
     async def apply_username(self, *, debug: bool = False) -> Union[bool, None]:
         """
-        Applies the username for the current season. Only changes nickname if
-        `bool` is False, otherwise only changes the nickname.
+        Applies the username for the current season.
+
+        Only changes nickname if `bool` is False, otherwise only changes the nickname.
```
MarkKoz:
```suggestion
        Only changes nickname if bool is False, otherwise only changes the nickname.
```

bot/seasons/valentines/be_my_valentine.py
```diff
@@ -191,27 +193,30 @@ def valentine_check(self, valentine_type):
     @staticmethod
     def random_user(author, members):
         """
-        Picks a random member from the list provided in `members`, ensuring
-        the author is not one of the options.
+        Picks a random member from the list provided in `members`.
```
MarkKoz:
```suggestion
        Picks a random member from the list provided in members.
```

bot/pagination.py
```diff
@@ -77,6 +76,7 @@ def add_line(self, line='', *, empty=False):
         RuntimeError
             The line was too big for the current :attr:`max_size`.
```
MarkKoz:
```suggestion
            The line was too big for the current max_size.
```

bot/pagination.py
scragly: Unsure about the `:attr:` declaration, but inline code and attributes I prefer to keep wrapped in backticks for correct inline code formatting, both for IDE output and for the possible future of having autogen docs.

bot/pagination.py
scragly: Same as before, prefer to keep wrapped in backticks.

bot/pagination.py
scragly: Same as before, prefer to keep wrapped in backticks.

bot/pagination.py
scragly: Same as before, prefer to keep wrapped in backticks.

bot/seasons/evergreen/snakes/utils.py
scragly: Same as before, prefer to keep wrapped in backticks.

sco1:
The agreed on changes have been made & I've deferred the autodoc decision until a consensus is reached on https://github.com/python-discord/organisation/issues/55. This mostly only affects the Snakes cog since it was written with a different prevailing style than what we're currently using, though there are some more sprinkled throughout. I agree that there should be a consistent style but should probably fall under its own scope.

scragly:
This will have to be held until the current PRs are merged in order to have the new lint specifications applied to what's already made but not quite merged.

sco1:
This is ready for review so it can get in place before the April event's PRs start coming through.

Issue #147, opened 2019-03-19
Move Save The Date Static Resource Load


sco1:
The static JSON resource for the Save the Date Cog is currently loaded every time the command is invoked:

https://github.com/python-discord/seasonalbot/blob/3e8a07af29428f33f51942bc9568fde8175611d3/bot/seasons/valentines/savethedate.py#L29-L30

This load should be moved either to the module namespace or the Cog's `__init__` so it is only loaded once.

Suhail6inkling:
I'll work on this.

PR #148, opened 2019-03-19
Save the Date - Static Resource Load

Suhail6inkling:
---
name: Save the Date fix
about: Move the loading of the static JSON in the Save the Date cog
issue: #147 

---

Solves this issue by moving the loading of the JSON file into the module namespace, so that it is not loaded every time the command is called

Hope this is okay :D

### DIFF
--- a/bot/seasons/valentines/savethedate.py
+++ b/bot/seasons/valentines/savethedate.py
@@ -13,7 +13,7 @@ log = logging.getLogger(__name__)
 HEART_EMOJIS = [":heart:", ":gift_heart:", ":revolving_hearts:", ":sparkling_heart:", ":two_hearts:"]
 
 with open(Path('bot', 'resources', 'valentines', 'date_ideas.json'), 'r', encoding="utf8") as f:
-    valentine_dates = load(f)
+    VALENTINES_DATES = load(f)
 
 
 class SaveTheDate:
@@ -29,7 +29,7 @@ class SaveTheDate:
         """
         Gives you ideas for what to do on a date with your valentine.
         """
-        random_date = random.choice(valentine_dates['ideas'])
+        random_date = random.choice(VALENTINES_DATES['ideas'])
         emoji_1 = random.choice(HEART_EMOJIS)
         emoji_2 = random.choice(HEART_EMOJIS)
         embed = discord.Embed(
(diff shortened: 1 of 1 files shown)


jchristgit:
I think the most preferable solution would be only loading the resource when the *cog* is actually loaded, so loading it in the `setup` function and passing it to the class. Curious what the others think.

sco1:
Is there a functional difference between when it's loaded (module vs. init vs. setup) with how we're using the cogs?

Regardless, it would probably be handy to settle on one and enforce it across the org since they're pretty scattered.

MarkKoz:
I think it would only have different behaviour when a cog is reloaded. I can't confirm that though.

scragly:
There would be no functional difference between placing it in setup vs cog init, and only the difference of referring to it via `self` compared to defining it on the module-level.

When reloading an extension, the entire extension (full package if package-style, full module if module-style) will be unloaded, reimported and the setup run again. Due to this, there's no functional difference for reload on any of the options.

The standard thing to do here is to put it on the module level, keeping it in full namespace throughout the module. I personally feel doing it any other way is likely just thinking too much about it.

bot/seasons/valentines/savethedate.py
```diff
@@ -12,6 +12,9 @@
 
 HEART_EMOJIS = [":heart:", ":gift_heart:", ":revolving_hearts:", ":sparkling_heart:", ":two_hearts:"]
 
+with open(Path('bot', 'resources', 'valentines', 'date_ideas.json'), 'r', encoding="utf8") as f:
+    valentine_dates = load(f)
```
scragly:
```suggestion
    VALENTINE_DATES = load(f)
```

bot/seasons/valentines/savethedate.py
```diff
@@ -26,17 +29,15 @@ def __init__(self, bot):
...
-            random_date = random.choice(valentine_dates['ideas'])
-            emoji_1 = random.choice(HEART_EMOJIS)
-            emoji_2 = random.choice(HEART_EMOJIS)
-            embed = discord.Embed(
-                title=f"{emoji_1}{random_date['name']}{emoji_2}",
-                description=f"{random_date['description']}",
-                colour=Colours.pink
-            )
-            await ctx.send(embed=embed)
+        random_date = random.choice(valentine_dates['ideas'])
```
scragly:
```suggestion
        random_date = random.choice(VALENTINE_DATES['ideas'])
```

PR #151, opened 2019-03-21
set up bot directory for pride event

jodth07:
this directory will be used for all pride/lgbtq related events codes.


jodth07:
Also add .DS_Store to .gitignore as well, in order to prevent that for other users

Issue #12, opened 2018-10-05
[Feature Request] Spooky  8-ball 


jodth07:
The goal of this task is to create an 8-ball that will generate random answers in a spookifying way, to support the halloween theme.

![screen shot 2018-10-07 at 2 04 29 pm](https://user-images.githubusercontent.com/8127555/46585252-87344b80-ca3c-11e8-83fb-ea04a0cc197e.png)

A user will ask a the bot a question via a command `!8ball, is discord python awesome?` and the bot will answer just like an 8-ball. (hhmm)

![screen shot 2018-10-07 at 2 02 13 pm](https://user-images.githubusercontent.com/8127555/46585256-987d5800-ca3c-11e8-9f4d-9240e5b0b1a5.png)

- [ ] list of spooky answers, 
- [ ] 8-ball image or emoji

WrightKD:
Hey , can I give this a shot ?

jodth07:
Hey @WrightKD how is it coming? Are you still working on this?

AnonGuy:
Closing until next October.

RubenJ01:
Id like to take a shot at this, if no one else has.

heavysaturn:
it's all yours,

jodth07:
merged

sco1:
No PR for this has been created

RohanRadia:
Has this been completed? If not I am down to take this on. Please do let me know.

PR #152, opened 2019-03-22
Magic 8ball command

### LINKED ISSUE #124: [Feature Request] Magic 8 Ball
The [Magic 8 Ball](https://en.m.wikipedia.org/wiki/Magic_8-Ball) has been helping folks answer life’s most difficult questions for decades, and it’s only fitting that Seasonalbot help us do the same!

This cog should accept any arbitrary question from the user and return sage wisdom from a prepopulated file of responses.

bramz:
Added an 8ball command, I added a question length requirement of 3 words to ensure a question was being asked. This can be reduced, increased or removed.

Closes #124


bot/seasons/evergreen/magic_8ball.py
```diff
@@ -0,0 +1,55 @@
...
+    "Concentrate and ask again",
+    "Don't count on it",
+    "My reply is no",
+    "My sources say no",
+    "Outlook not so good",
+    "Very doubtful",
+]
+
+
+class Magic_8ball:
```
jodth07: follow class naming conventions CamelCase

bot/seasons/evergreen/magic_8ball.py
```diff
@@ -0,0 +1,55 @@
...
+    "Ask again later",
+    "Better not tell you now",
+    "Cannot predict now",
+    "Concentrate and ask again",
+    "Don't count on it",
+    "My reply is no",
+    "My sources say no",
+    "Outlook not so good",
+    "Very doubtful",
+]
```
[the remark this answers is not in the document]
bramz: This command was listed as part of the evergreen season, I've added this to the Halloween resources directory. Should I add this file to a new directory named `evergreen` in `resources` directory? Keeping the data a bit more seasonally organized. #12 seems to be a Halloween version of this command, this may cause confusion in the future.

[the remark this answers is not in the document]
sco1: The resources for this cog should go into an evergreen resources folder, this is a separate cog from the functionality proposed in #12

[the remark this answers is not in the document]
jodth07: I apologize for that, it should be in evergreen, as @sco1 mentioned.

bot/seasons/evergreen/magic_8ball.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        if len(question) >= 3:
+            answer = random.choice(self.answers)
+            await ctx.send(answer)
+
+
+# Required in order to load the cog, use the class name in the add_cog function.
+def setup(bot):
+    bot.add_cog(Magic8ball(bot))
+    log.debug("Magic 8ball cog loaded")
```
sco1: `log.info`, per the contributor document

bot/seasons/evergreen/magic_8ball.py
```diff
@@ -0,0 +1,34 @@
...
+        self.bot = bot
+        with open(Path("bot", "resources", "evergreen", "magic8ball.json"), "r") as file:
+            self.answers = json.load(file)
+
+    @commands.command(name="8ball")
+    async def output_answer(self, ctx, question: list):
+        """
+        Return a magic 8 ball answer from answers list.
+        """
+        if len(question) >= 3:
```
sco1: The user should get some feedback if this criteria isn't met rather than failing silently.

bramz: I was under the impression that this was giving the user some feedback. The bout outputs a message when the criteria is not met.
image

bramz: Are you suggesting, something like a command usage output on fail instead?

sco1: That is only shown by the [error handler](https://github.com/python-discord/seasonalbot/blob/master/bot/seasons/evergreen/error_handler.py) if `commands.UserInputError` is raised, which doesn't happen if `question` is shorter than 3 characters (nothing happens).

bot/seasons/evergreen/magic_8ball.py
```diff
@@ -0,0 +1,34 @@
...
+    A Magic 8ball command to respond to a users question.
+    """
+
+    def __init__(self, bot):
+        self.bot = bot
+        with open(Path("bot", "resources", "evergreen", "magic8ball.json"), "r") as file:
+            self.answers = json.load(file)
+
+    @commands.command(name="8ball")
+    async def output_answer(self, ctx, question: list):
```
sco1: Why is `question` being treated as a list here, rather than a string?

Take a look at the [Command parameters](https://discordpy.readthedocs.io/en/rewrite/ext/commands/commands.html#parameters) documentation, a keyword-only argument would probably be more appropriate here. Since this will currently only capture the first word of the user's input as `question`, this can lead to false positives on the length check if the question starts with a short word, e.g. `Do Androids Dream of Electric Sheep?`

PR #162, opened 2019-03-30
Always announce if docstring, better formatting.

scragly:
Announcements were setup previously to not send anything if no seasonal commands were found. Since we want to announce regardless to entice people to contribute, I've adjusted the logic to always announce, but only add the commands info if there were any found.

There was also a formatting issue with newlines and wrapping that was a result of using the docstring which we explicitly make a newline for when it gets close to linting max length. The new formatting process takes a docstring, splits by paragraphs, and then replaces any newlines with a single space, letting single paragraph blocks wrap correctly in the announcement embed. Example:

**Docstring**
```py
    """
    Easter is a beautiful time of the year often celebrated after the first Full Moon of the new spring season.
    This time is quite beautiful due to the colorful flowers coming out to greet us. So. let's greet Spring
    in an Easter celebration of contributions.
    """
```
**Before**
image
**After**
image

### DIFF
--- a/bot/seasons/season.py
+++ b/bot/seasons/season.py
@@ -17,6 +17,8 @@ from bot.decorators import with_role
 
 log = logging.getLogger(__name__)
 
+ICON_BASE_URL = "https://raw.githubusercontent.com/python-discord/branding/master"
+
 
 def get_seasons() -> List[str]:
     """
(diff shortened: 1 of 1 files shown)


scragly:
That's a great idea @heavysaturn. Should it be main image in the embed or just a thumbnail off to the right side?

scragly:
I went with adding the icon as the main image.

PR #176, opened 2019-04-01
Easter Egg Decorating

### LINKED ISSUE #168: [Feature Request] Easter Egg Decorating
What's Easter without decorated Easter Eggs? Implement a command takes a color palette as an argument and returns a newly "painted" Easter Egg with a nice pattern/design in the specified colors.

Seasonal bot already has Pillow installed as a dependency, so try using that to manipulate your images. One possible implementation idea is to create bunch of template image files with very well defined color areas/patterns. You could then replace the colored areas in the image array by the colors specified in the command and send the result back to the channel.

### Implementation details:

- [ ] Create a command that takes a couple of colors as arguments
- [ ] "Paint" a random egg design based off of those provided colors
- [ ] Post the resulting Easter Egg, along with the user's nickname, in an embed to the channel

Suhail6inkling:
---
name: Easter Egg Decorating
about: A command that adds the ability to decorate Easter Eggs
Closes #168 

---

Adds the `EggDecorating` cog containing the `eggdecorate` command, which allows users to input a bunch of colours and create an Easter Egg based on premade designs

image

This image shows what the command looks like. Parameters accepted are equivalent to whatever `discord.ext.commands.ColourConverter` accepts

Hope this is okay :D


bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+        converted = []
+        for c in colours:
+            try:
+                colour = await commands.ColourConverter().convert(ctx, c)
+                # Attempts to convert the arguments into discord.Colour
+                converted.append(colour)
+            except commands.BadArgument:
+                invalid.append(c)
+
+        if len(invalid) > 1:
```
sco1: Rather than returning out of this block, how about selecting some random colors to replace the invalid ones and then saving this feedback for the output?

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if len(colours) < 2:
+            return await ctx.send("You must include at least 2 colours!")
+
+        invalid = []
+        converted = []
+        for c in colours:
+            try:
+                colour = await commands.ColourConverter().convert(ctx, c)
+                # Attempts to convert the arguments into discord.Colour
```
sco1: It's probably worth having some more common colors defined as fallbacks for those that discord.py doesn't define (e.g. yellow, pink).

This could also be more or less mitigated by the comment below.

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+
+        It picks from a random set of designs and alters the colours to the user's liking
+        """
+
+        if len(colours) < 2:
+            return await ctx.send("You must include at least 2 colours!")
+
+        invalid = []
+        converted = []
+        for c in colours:
```
sco1: Can some handling be added for folks who try defining colors with comma separators? We don't necessarily want, for example, `.eggdecorate blue, blurple, green` to fail if it can be avoided.

scragly: I don't really think the comma separation needs to be handled other than maybe mentioning what's accepted in the help output.
This is doubly so if this was altered to using a proper type-hinted converter which is going to be my recommendation coming up in the review.
If we start accepting comma separated arguments now, it'll be inconsistent with other command formats and would make us have to support it for other commands which isn't a valid, natively handled argument parsing format with the d.py commands extension.

bot/seasons/easter/egg_decorating.py
scragly: Random replacements isn't a great idea, as it's unclear behaviour. It would be preferred if more common colour names were supported instead and reword the error to state it doesn't know the colour if it's still not found.

bot/seasons/easter/egg_decorating.py
scragly: Better support for a wider array of colour names would be a good idea, for sure.

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+            try:
+                colour = await commands.ColourConverter().convert(ctx, c)
+                # Attempts to convert the arguments into discord.Colour
+                converted.append(colour)
+            except commands.BadArgument:
+                invalid.append(c)
```
scragly: If we added the type hinted converter properly, we don't need to manually convert to discord.Colour for those that are supported. 

For anything that isn't, we can put the colour through the recommended helper method from before that tries to match the colour with the HTML or XKCD colour names. It will return None if there's no match, so it gets appended to `invalid` like before, otherwise it just replaces the current item in the list by index.

```py
        invalid = []
        colours = list(colours)
        for idx, colour in enumerate(colours):
            if isinstance(colour, discord.Colour):
                continue
            value = self.replace_invalid(colour)
            if value:
                colours[idx] = discord.Colour(value)
            else:
                invalid.append(colour)
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if len(invalid) > 1:
+            return await ctx.send(f"The following colours are invalid: {' '.join(invalid)}")
```
scragly: Sometimes what a user thinks and quite possibly could be a valid colour would end up not being in the colour names we support. Rather than tell them it's invalid, because aside from code it might be, it's better to say that we don't know that colour name.
```suggestion
            return await ctx.send(f"Sorry, I don't know these colours: {' '.join(invalid)}")
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+        elif len(invalid) == 1:
+            return await ctx.send(f"{invalid[0]} is an invalid colour!")
```
scragly: Same as above:
```suggestion
            return await ctx.send(f"Sorry, I don't know the colour {invalid[0]}!")
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+
+        colours = converted
```
scragly: If the above suggestions are accepted, `converted` is no longer a variable so can be removed here:
```suggestion
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+
```
scragly: If the above line is removed, this one should be too to avoid double newlines.
```suggestion
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+        async with ctx.typing():
+            colours *= 4
```
scragly: The comment for this should be above it, also you can happily repeat and expand this list only for the amount necessary and if it's actually necessary with the following:
```suggestion
            # expand list to 8 colours
            colours_n = len(colours)
            if colours_n < 8:
                q, r = divmod(8, colours_n)
                colours = colours * q + colours[:r]
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+            # This is to ensure that no IndexErrors are raised since the most amount of colours on an egg is 8
```
scragly: With the comment being moved above the relevant line, this can be removed.
```suggestion
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+            num = random.randint(1, 6)
+            im = Image.open(Path("bot", "resources", "easter", "easter_eggs", f"design{num}.png"))
+            data = list(im.getdata())
+
+            replaceable = {x for x in data if x not in IRREPLACEABLE}  # Turns it into a set to avoid duplicates
```
scragly: Using a set here is self-explanatory so the comment isn't necessary.
```suggestion
            replaceable = {x for x in data if x not in IRREPLACEABLE}
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+            replaceable = sorted(replaceable, key=COLOURS.index)  # Sorts it by colour order
```
scragly: Using sorted here is self-explanatory so the comment isn't necessary.
```suggestion
            replaceable = sorted(replaceable, key=COLOURS.index)
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+            new_im = Image.new(im.mode, im.size)
+            new_im.putdata(new_data)
+
+            bufferedio = BytesIO()
+            new_im.save(bufferedio, format="PNG")
+
+            bufferedio.seek(0)
+
+            file = discord.File(bufferedio, filename="egg.png")  # Creates file to be used in embed
+            embed = discord.Embed(title="Your egg", description="Here is your pretty little egg. Hope you like it!")
```
scragly: I feel like the title is a bit bland, maybe something like the following:
```suggestion
            embed = discord.Embed(
                title="Your Colourful Easter Egg",
                description="Here is your pretty little egg. Hope you like it!"
            )
```

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+            embed.set_image(url="attachment://egg.png")
+            embed.set_author(name=ctx.author.display_name, icon_url=ctx.author.avatar_url)
```
scragly: Having the author at the top distracts from the title and main info. This would be better as footer info imo:
```suggestion
            embed.set_footer(text=f"Made by {ctx.author.display_name}", icon_url=ctx.author.avatar_url)
```
image

bot/seasons/easter/egg_decorating.py
```diff
@@ -0,0 +1,95 @@
...
+
+            await ctx.send(file=file, embed=embed)
```
scragly: Sending needs to be dedented so it's outside of the typing context manager and doesn't end up having the bot keep typing after the message is sent:
```suggestion
        await ctx.send(file=file, embed=embed)
```

bot/seasons/easter/egg_decorating.py
scragly: I've added in two new resources to the PR that can be used:
- HTML colour names and values
- XKCD colour names and values

My review has suggested changes that make use of them.

PR #193, opened 2019-04-17
Update egghead_questions.json

scOwez:
- Fixed question "Who is considered to be the founder of Earth Day?"
- Fixed question "When was the first Earth Day?"

### DIFF
--- a/bot/resources/easter/egghead_questions.json
+++ b/bot/resources/easter/egghead_questions.json
@@ -47,7 +47,7 @@
             "1999",
             "1970"
         ],
-        "correct_answer": 2
+        "correct_answer": 3
     },
     {
         "question": "Who is considered to be the founder of Earth Day?",
@@ -55,7 +55,7 @@
             "President Jimmy Carter",
             "President John F. Kennedy",
             "Vice President Al Gore",
-            "Senator Gaylord Nelson"
+            "John McConnell"
         ],
         "correct_answer": 3
     },
--- a/bot/resources/easter/egghead_questions.json
+++ b/bot/resources/easter/egghead_questions.json
@@ -55,7 +55,7 @@
             "President Jimmy Carter",
             "President John F. Kennedy",
             "Vice President Al Gore",
-            "John McConnell"
+           "Senator Gaylord Nelson"
         ],
         "correct_answer": 3
     },


bot/resources/easter/egghead_questions.json
```diff
@@ -47,15 +47,15 @@
...
+        "correct_answer": 3
     },
     {
         "question": "Who is considered to be the founder of Earth Day?",
         "answers": [
             "President Jimmy Carter",
             "President John F. Kennedy",
             "Vice President Al Gore",
-            "Senator Gaylord Nelson"
+            "John McConnell"
```
SebastiaanZ: I'm not sure if I agree with this correction.

The current Earth Day we celebrate, the April 22 one, was founded by Senator Gaylord Nelson in 1970. There was a separate day "to honour the earth and piece" held on March 21, 1970 that was founded by John McConnell, but that's not strictly the predecessor of the current Earth Day.

The official Earth Day website states this:

> The idea for a national day to focus on the environment came to Earth Day founder Gaylord Nelson

[source](https://www.earthday.org/about/the-history-of-earth-day/)

And Wikipedia describes it as such as well:

> Senator Gaylord Nelson, the original founder of Earth Day, was honorary chairman for both groups.

Since we're truly talking about this Earth Day on April 22, I think we should go with the official source of the Earth Day organization that states that Senator Gaylord Nelson was the founder.

So, 

```suggestion
           "Senator Gaylord Nelson"
```

jodth07: I thought 4/20 was earth day?
so, now there is 4/21 and 4/22? that is cool. I did not know that.

Can we have all these in there as well?

sco1: I agree with Ves that we should defer to the history of the event as it exists in current times and discard the change proposed by this PR.

A couple questions on its 1969 origins would be desirable instead.

scOwez:
Added change

bot/resources/easter/egghead_questions.json
```diff
@@ -47,15 +47,15 @@
...
+        "correct_answer": 3
     },
     {
         "question": "Who is considered to be the founder of Earth Day?",
         "answers": [
             "President Jimmy Carter",
             "President John F. Kennedy",
             "Vice President Al Gore",
-            "Senator Gaylord Nelson"
+           "Senator Gaylord Nelson"
```
sco1:
```suggestion
            "Senator Gaylord Nelson"
```

For my sanity

PR #204, opened 2019-04-28
Bunny Name Generator

### LINKED ISSUE #170: [Feature Request] Bunny name generator
`.bunny_name` generates a random name for a bunny, like Thumper or Floofy. It should also be able to _bunnify_ a user's existing username, so that lemon becomes lemonfoot and Volcyy becomes Voolfy, or something like that.

This issue intentionally leaves a lot of room for creativity. Please find your own way to solve this. As long as it's fun to play with and the names generated sound like they might make okay bunny names, you've solved this problem.

In your pull request, please give examples of what your code outputs.

glowingrunes:
### Closes #170

This cog contains two commands: '.bunnyname' and '.bunnifyme'.

**.bunnyname**
Simply picks a random bunny name from a JSON file of pre-generated names and returns the name as plain text.

**.bunnifyme**
This one is a bit more complex. The command uses regular expressions to match the user's display name against several different patterns.
First the display name is searched for separators and whitespace. If it contains either of these, one of the words in the name will be replaced with a bunny-related word. E.g. "Operative Loren" could become "Operative Cotton" or "Cotton Loren".
If the display name contains no spaces or separators, it will search for vowels in the name. If the display name contains a vowel and the letter y, it will replace the part between those two characters with something cutesy. For instance, "Volcyy" should become "Voofy". This could result in some odd names, but we'll just have to see how it turns out!
Finally, if the display name doesn't match either of those patterns, then by default a suffix will be added to the end of it. For example "lemon" could become "lemontail".

I'm rather new to Python, but I tried to make my code as Pythonic as I could. Hopefully it's not too spaghetti...

### DIFF
--- a/bot/seasons/easter/bunny_name_generator.py
+++ b/bot/seasons/easter/bunny_name_generator.py
@@ -18,6 +18,54 @@ class BunnyNameGenerator(commands.Cog):
     def __init__(self, bot):
         self.bot = bot
 
+    def find_separators(self, displayname):
+        """
+        Check if Discord name contains spaces so we can bunnify an individual word in the name.
+        """
+
+        new_name = re.split(r'[_.\s]', str(displayname))
+        if displayname not in new_name:
+            return new_name
+
+    def find_vowels(self, displayname):
+        """
+        Finds vowels in the user's display name.
+
+        If the Discord name contains a vowel and the letter y,
+        it will match one or more of these patterns.
+        Only the most recently matched pattern will apply the changes.
+        """
+
+        new_name = None
+
+        option1 = re.sub(r'a.+y', 'patchy', displayname)
+        option2 = re.sub(r'e.+y', 'ears', displayname)
+        option3 = re.sub(r'i.+y', 'ditsy', displayname)
+        option4 = re.sub(r'o.+y', 'oofy', displayname)
+        option5 = re.sub(r'u.+y', 'uffy', displayname)
+
+        if option1 != displayname:
+            new_name = option1
+        if option2 != displayname:
+            new_name = option2
+        if option3 != displayname:
+            new_name = option3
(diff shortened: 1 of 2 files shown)


jodth07:
@glowingrunes @sco1 provided some nice reviews for you. Take care of these, and you are good to go, and we can merge it.

glowingrunes:
@sco1 Thank you so much your all your feedback :) I've made a few changes based on what you recommended. Hopefully the code makes more sense now. Unfortunately, I can't think of anything right now to make the find_vowels function any better.

bot/seasons/easter/bunny_name_generator.py
```diff
@@ -0,0 +1,106 @@
...
+        Finds vowels in the user's display name.
+
+        If the Discord name contains a vowel and the letter y,
+        it will match one or more of these patterns.
+        Only the most recently matched pattern will apply the changes.
+        """
+
+        new_name = None
+
+        option1 = re.sub(r'a.+y', 'patchy', displayname)
```
GhostofGoes: Not a huge fan of the hard-coded and duplicate logic in `find_vowels`. The substitution logic should be put into a `for` loop that iterates over, say, a list of tuples with vowel/sub pairs. Other than that, this looks great!

sco1: Something like this should maintain the desired behavior and looks a bit cleaner.

```py
expressions = [
    (r'a.+y', 'patchy'),
    (r'e.+y', 'ears'),
    (r'i.+y', 'ditsy'),
    (r'o.+y', 'oofy'),
    (r'u.+y', 'uffy'),
]

for exp, vowel_sub in expressions:
    new_name = re.sub(exp, vowel_sub, displayname)
    if new_name != displayname:
        return new_name
```

PR #201, opened 2019-04-24
Change d.py Dependency from GH to PyPI

sco1:
Closes #191

PR #209, opened 2019-05-12
Pride Avatar

### LINKED ISSUE #206: [Feature Request] Profile Picture Pride-ify
User would run a command and upon doing so their profile picture would be taken, would have a rainbow gradient put on top and then would have the image sent back in the same channel.

Any questions? Feel free to ask below! Try and make sure that the picture generation time doesn't take too long! If it is going to take a while maybe make the bot have the 'typing' status so the user knows that something is happening instead of presuming the command has not worked.

Suhail6inkling:
---
name: Pride Avatar
about: Wraps an avatar with a flag as a border
Closes #206 

---

Adds the `PrideAvatar` cog containing the `prideavatar` command which wraps an avatar using a given LGBT flag.

This PR also includes the images of several pride flags from all different sexual orientation and gender identities, that may be used with future pride commands.

[When nothing else is given, a standard LGBT rainbow flag is added as a border](https://user-images.githubusercontent.com/38522108/57578022-05961780-747c-11e9-8c86-1b2e2cbc7ee8.png)

[A sexual orientation or gender identity can be given to provide the respective flag as a border instead](https://user-images.githubusercontent.com/38522108/57578027-21012280-747c-11e9-9a5a-aac9d238887b.png)

[A number (between 0 and 512) can be added to adjust the thickness/padding of the border](https://user-images.githubusercontent.com/38522108/57578034-645b9100-747c-11e9-9a0f-36427e045b88.png)

Please note: This PR will not work until #201 has been merged due to a new attribute from a new discord.py update being used.

Hope this is okay :D


bot/seasons/pride/pride_avatar.py
```diff
@@ -0,0 +1,133 @@
...
+        async with ctx.typing():
+
+            # Get avatar bytes
+            image_bytes = ctx.author.avatar_url.read()
+            avatar = Image.open(BytesIO(image_bytes))
+            avatar = avatar.convert("RGBA").resize((1024, 1024))
+
+            avatar = self.crop_avatar(avatar)
+
+            ring = Image.open(Path("bot", "resources", "pride", "flags", f"{flag}.png")).resize((1024, 1024))
```
sco1: You need to ensure that the [`mode`](https://pillow.readthedocs.io/en/5.1.x/handbook/concepts.html#concept-modes) of the loaded flag matches that of the avatar, otherwise `alpha_composite()` will raise a `ValueError`.

The transgender flag, for example, loads as `PA`, and just turns into a black ring if you try `ring.convert("RGBA")`.

Suhail6inkling: I understand the requirement for the mode of the loaded flag to match, but I couldn't replicate your specific example.

For me, the flag loads as a normal `RGBA` image and works fine with the command.

I'm not sure if it's a difference between OSs or PIL versions or something.

sco1: Surely there's a way to enforce it even if you can't replicated it on your machine. "It works for me" really doesn't help the cog not crash when deployed by the bot.

Suhail6inkling: I was able to replicate the transgender flag loading as `PA` however no matter what I tried, doing `ring.convert("RGBA")` would always work for me.

I added that line to this PR to enforce that the ring is RGBA however I cannot replicate the blank ring.

The way to enforce the mode is to convert it to the preferred mode however I cannot see why it turns into a black ring on your device, no matter how hard I've tried

sco1: Seems like this is probably just my computer/debugger's fault. The black ring of death doesn't happen anymore.

PR #211, opened 2019-05-13
#110 Reactions on messages in Projects channel

### LINKED ISSUE #110: [Feature Request] List of reactions for each message in #show-your-projects
So every message that is sent in the `#show-your-projects` channel, the bot sends this list of reactions in order:
👍 👎 ❤️ 🚩 👀 
This can be changed any way you want, but essentially whenever someone posts a message showing their project, the bot immediately sends these reactions, so that people can quickly react with those emojis without looking for them in a menu. You can even have it so you turn off the ability for people to add new reactions, only using the reactions above. An extra technical detail, since some tend to post their project as more than one message, maybe a 30-second slowmode on the channel can deter that too.

## Implementation details

- [ ] Write the bot command (duh)
    - [ ] Choose the reactions and the order
- [ ] Change Permissions (optional)
- [ ] Add slowmode (optional)

## Additional information
This is an evergreen request.

I hope I didn't miss anything, but if I did you can leave a comment telling me what. Also if anyone has suggestions for certain emojis to use as reactions you can leave those in too.

RohanRadia:
Closes #110

Adds reactions to messages in the Show Your Projects channel. If a user posts consecutively it will only reply to their first post. This is to avoid the emoji set being on the link/file and the description.


bot/seasons/evergreen/showprojects.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self, bot):
+        self.bot = bot
+        self.lastPoster = 0  # Given 0 as the default last poster ID as no user can actually have 0 assigned to them
+
+    @commands.Cog.listener()
+    async def on_message(self, message):
+        """Adds reactions to posts in #show-your-projects"""
+
+        reactions = ["\U0001f44d", "\U00002764", "\U0001f440", "\U0001f389", "\U0001f680", "\U00002b50", "\U0001f6a9"]
+        if message.channel.id == Channels.show_your_projects and message.author.id != 528937022996611082 \
```
LordHemlock: I'd personally prefer if this checked whether the author is a bot rather than just having SeasonalBot's ID hard coded in. However if we do decide to keep it like this, we should have a comment to explain that this is SeasonalBot's ID, or better yet name it as a variable or constant so there's no confusion.  

But ideally, it should just check to see if it's a bot or not.

You might also consider:
```suggestion
        if (message.channel.id == Channels.show_your_projects
            and message.author.id != 528937022996611082
            and message.author.id != self.lastPoster):
```
Or alternatively 
```suggestion
        if (message.channel.id == Channels.show_your_projects 
            and message.author.id not in (self.lastPoster, 528937022996611082):
```

sco1: You don't need an ID at all to check for itself. `User` and `Member` objects both support direct equality checks between themselves & the other.

RohanRadia: I sweat d.py has something like isbot? I can't remember. Will need to check docs when on PC.

PR #223, opened 2019-06-03
Pride Avatar Flags

### LINKED ISSUE #222: [Feature Request] List possible flag options to prideavatar
The prideavatar command takes a flag option string to select which background should be used. Presently, there isn't a way that I know of to view what options are available without peeking at the bots code or bugging a developer (sorry Ref). This feature request seeks to rectify that.

## Implementation details

- [ ] Extend the `prideavatar` command to show the list of options either in the help message (e.g. `.help prideavatar`) or as a discrete option (`.prideavatar options` or `.prideavatar flags`)

Suhail6inkling:
A simple addition to #209 that fixes #222 by adding a command listing the possible options for flags

Hope this is okay :D

### DIFF
--- a/bot/seasons/pride/pride_avatar.py
+++ b/bot/seasons/pride/pride_avatar.py
@@ -129,7 +129,7 @@ class PrideAvatar(commands.Cog):
     async def flags(self, ctx):
         """This lists the flags that can be used with the prideavatar command."""
         choices = sorted(set(OPTIONS.values()))
-        options = "- " + "\n- ".join(choices)
+        options = "• " + "\n• ".join(choices)
         embed = discord.Embed(
             title="I have the following flags:",
             description=options,
--- a/bot/seasons/pride/pride_avatar.py
+++ b/bot/seasons/pride/pride_avatar.py
@@ -6,6 +6,8 @@ import discord
 from PIL import Image, ImageDraw
 from discord.ext import commands
 
+from bot.constants import Colours
+
 log = logging.getLogger(__name__)
 
 OPTIONS = {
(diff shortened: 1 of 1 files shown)


bot/seasons/pride/pride_avatar.py
```diff
@@ -123,6 +123,12 @@ def crop_ring(ring, px):
 
         await ctx.send(file=file, embed=embed)
 
+    @prideavatar.command()
+    async def flags(self, ctx):
+        """This lists the flags that can be used with the prideavatar command."""
+        options = "\n".join(set(OPTIONS.values()))
+        await ctx.send(f"**I have the following flags:**\n{options}")
```
SebastiaanZ: Instead of a plain text message, I would like to see this in a formatted embed. Maybe you could sort them as well, which may make looking up a name in the list somewhat easier (since it's relatively long).

This is just a boring example, without color, of what I mean (without sorting):

bot/seasons/pride/pride_avatar.py
```diff
@@ -123,6 +125,19 @@ def crop_ring(ring, px):
 
         await ctx.send(file=file, embed=embed)
 
+    @prideavatar.command()
+    async def flags(self, ctx):
+        """This lists the flags that can be used with the prideavatar command."""
+        choices = sorted(set(OPTIONS.values()))
+        options = "- " + "\n- ".join(choices)
```
SebastiaanZ: Hi! Looks good!

Most of our other embeds with lists use `•` instead of `-` as the list character. You can probably just accept the change suggested below:

```suggestion
        options = "• " + "\n• ".join(choices)
```

Should be pretty much ready to go as far as I'm concerned after that.

PR #230, opened 2019-06-21
Cycle Evergreen Animated Server Icons

### LINKED ISSUE #229: [Feature Request] Cycle Evergreen Animated Server Icons
Thanks to @kwzrd, we now have [four animated versions](https://github.com/python-discord/branding/tree/master/logos/logo_animated) of the default logo we use as the server icon between seasonalbot events (i.e., the so called "evergreen" seasons). We would like to have a seasonalbot feature that allows us to cycle through the different versions on a daily basis. 

## Implementation details

- [ ] Create a background task that changes the server logo on a daily basis.
- [ ] Make sure that it's easy to include more versions of the logo after they've been added.
- [ ] Make sure that we cycle through the logos instead of selecting one at random

## Additional information

The next time the season will change to evergreen is on the 1st of July. This is not an essential feature that absolutely needs to be done before then, but it would if we were able to show off the new, awesome logos in near future.

Suhail6inkling:
---
name: Evergreen Animated Server Icon Cycling
about: Changes to the season logic to allow for cycling of icons daily
Closes #229 

---
This adds the four animated versions of the default logo to the evergreen icons and alters the season logic to allow for seasons to contain a list of items that can be cycled through daily

Hope this is okay :D

### DIFF
--- a/bot/seasons/evergreen/__init__.py
+++ b/bot/seasons/evergreen/__init__.py
@@ -9,5 +9,5 @@ class Evergreen(SeasonBase):
         "/logos/logo_animated/heartbeat/heartbeat.gif",
         "/logos/logo_animated/spinner/spinner.gif",
         "/logos/logo_animated/tongues/tongues.gif",
-        "/logos/logo_animated/winky/winky.gif"
+        "/logos/logo_animated/winky/winky.gif",
     )
--- a/bot/seasons/evergreen/__init__.py
+++ b/bot/seasons/evergreen/__init__.py
@@ -5,9 +5,9 @@ class Evergreen(SeasonBase):
     """Evergreen Seasonal event attributes."""
 
     bot_icon = "/logos/logo_seasonal/evergreen/logo_evergreen.png"
-    icon = [
+    icon = (
         "/logos/logo_animated/heartbeat/heartbeat.gif",
         "/logos/logo_animated/spinner/spinner.gif",
         "/logos/logo_animated/tongues/tongues.gif",
         "/logos/logo_animated/winky/winky.gif"
-    ]
+    )
--- a/bot/seasons/evergreen/__init__.py
+++ b/bot/seasons/evergreen/__init__.py
@@ -5,3 +5,9 @@ class Evergreen(SeasonBase):
     """Evergreen Seasonal event attributes."""
 
     bot_icon = "/logos/logo_seasonal/evergreen/logo_evergreen.png"
+    icon = [
+        "/logos/logo_animated/heartbeat/heartbeat.gif",
+        "/logos/logo_animated/spinner/spinner.gif",
+        "/logos/logo_animated/tongues/tongues.gif",
+        "/logos/logo_animated/winky/winky.gif"
+    ]
--- a/bot/seasons/season.py
+++ b/bot/seasons/season.py
@@ -6,7 +6,7 @@ import inspect
 import logging
 import pkgutil
 from pathlib import Path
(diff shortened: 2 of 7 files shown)


bot/seasons/evergreen/__init__.py
```diff
@@ -5,3 +5,9 @@ class Evergreen(SeasonBase):
     """Evergreen Seasonal event attributes."""
 
     bot_icon = "/logos/logo_seasonal/evergreen/logo_evergreen.png"
+    icon = [
```
MarkKoz: Is there a reason to make this a separate variable? Also, this should be a tuple rather than a list.

bot/seasons/season.py
```diff
@@ -132,11 +134,18 @@ def greeting(self) -> str:
         """
         return f"New Season, {self.name_clean}!"
 
-    async def get_icon(self, avatar: bool = False) -> bytes:
+    @property
+    def cycle(self) -> bool:
+        """Return whether the icon attribute is a list and hence needs to be cycled through."""
+        return isinstance(self.icon, list)
```
MarkKoz: I think it'd be better to check if it's a `typing.Sequence` rather than locking it into being a list.

bot/seasons/season.py
```diff
@@ -132,11 +134,18 @@ def greeting(self) -> str:
         """
         return f"New Season, {self.name_clean}!"
 
-    async def get_icon(self, avatar: bool = False) -> bytes:
+    @property
+    def cycle(self) -> bool:
```
MarkKoz: Making a property just for this seems unnecessary.

bot/seasons/season.py
```diff
@@ -132,11 +134,18 @@ def greeting(self) -> str:
...
+        """Return whether the icon attribute is a list and hence needs to be cycled through."""
+        return isinstance(self.icon, list)
+
+    async def get_icon(self, avatar: bool = False, index: int = 0) -> List[Union[bytes, str]]:
```
MarkKoz: The return type annotation is wrong because it's actually returning a tuple; use `typing.Tuple` instead. You don't need `Union` for it because you specify the type of each position in the tuple directly.

Suhail6inkling:
>I think you could just make it always use a tuple for icons because even with only a single element in the tuple, it should work (the cycle will just appear to do nothing since the index will never change).

Surely that means that all of the already-set icon URLs will need to be changed in the future? There might be times where only one is needed and whoever sets it will set it as a string rather than a tuple. I just think to be safe the logic should check for string or sequence

bot/seasons/evergreen/__init__.py
Suhail6inkling: All seasons have an `icon` attribute and I thought it'd be better to change the evergreen icon and leave the default icon in `season.py` to any new season that may not have an icon yet

MarkKoz:
Yes, it means everything will need to be a tuple and I don't think it's a bad thing. Maybe others have different opinions on this. You've already done the work so I don't mind leaving it in but if I were to do this from scratch I would have not bothered supporting both strings and tuples as it makes the implementation simpler.

bot/seasons/evergreen/__init__.py
MarkKoz: So `bot_icon` is the default for seasons? That's fair.

Suhail6inkling:
Done the other changes - although I'd love and appreciate getting more opinions on whether to convert all icon variables to tuples or support both formats

PR #231, opened 2019-06-30
Edit dates

Suhail6inkling:
Some seasons have dates that end a day early because the end date is exclusive


lemonsaurus:
Let's just go ahead and get this merged.

PR #251, opened 2019-08-07
Limit SeasonalBot commands to whitelisted channels

kosayoda:
This PR restricts SeasonalBot commands to these channels:
* bot-commands, seasonalbot_commands
* ot0, ot1 and ot2 channels
* dev-test

To do so, functionality was added `in_channel` to allow a list of roles that bypass the check. The predicate function and the actual `discord.ext.commands.check` was separated so the predicate can be added as a global check to the bot.

The `in_channel` check can be overriden for certain commands using a decorator. Commands that currently have the override are `!issue` and `!adventofcode`.

To avoid circular imports, the bot instance definition was moved to the same file as the `SeasonalBot` class, `bot.py`. 

Constants brought over from the Python bot:
* MODERATION_ROLES, STAFF_ROLES
* NEGATIVE_REPLIES, POSITIVE_REPLIES

### DIFF
--- a/bot/decorators.py
+++ b/bot/decorators.py
@@ -24,8 +24,10 @@ def with_role(*role_ids: int):
     """Check to see whether the invoking user has any of the roles specified in role_ids."""
     async def predicate(ctx: Context):
         if not ctx.guild:  # Return False in a DM
-            log.debug(f"{ctx.author} tried to use the '{ctx.command.name}'command from a DM. "
-                      "This command is restricted by the with_role decorator. Rejecting request.")
+            log.debug(
+                f"{ctx.author} tried to use the '{ctx.command.name}'command from a DM. "
+                "This command is restricted by the with_role decorator. Rejecting request."
+            )
             return False
 
         for role in ctx.author.roles:
@@ -33,8 +35,10 @@ def with_role(*role_ids: int):
                 log.debug(f"{ctx.author} has the '{role.name}' role, and passes the check.")
                 return True
 
-        log.debug(f"{ctx.author} does not have the required role to use "
-                  f"the '{ctx.command.name}' command, so the request is rejected.")
+        log.debug(
+            f"{ctx.author} does not have the required role to use "
+            f"the '{ctx.command.name}' command, so the request is rejected."
+        )
         return False
     return commands.check(predicate)
 
@@ -43,14 +47,18 @@ def without_role(*role_ids: int):
     """Check whether the invoking user does not have all of the roles specified in role_ids."""
     async def predicate(ctx: Context):
         if not ctx.guild:  # Return False in a DM
-            log.debug(f"{ctx.author} tried to use the '{ctx.command.name}' command from a DM. "
-                      "This command is restricted by the without_role decorator. Rejecting request.")
+            log.debug(
+                f"{ctx.author} tried to use the '{ctx.command.name}' command from a DM. "
+                "This command is restricted by the without_role decorator. Rejecting request."
+            )
             return False
 
         author_roles = [role.id for role in ctx.author.roles]
         check = all(role not in author_roles for role in role_ids)
-        log.debug(f"{ctx.author} tried to call the '{ctx.command.name}' command. "
-                  f"The result of the without_role check was {check}.")
+        log.debug(
+            f"{ctx.author} tried to call the '{ctx.command.name}' command. "
+            f"The result of the without_role check was {check}."
+        )
         return check
     return commands.check(predicate)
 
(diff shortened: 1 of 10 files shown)


bot/seasons/evergreen/error_handler.py
```diff
@@ -34,6 +39,16 @@ def revert_cooldown_counter(command, message):
...
 
+        if isinstance(error, InChannelCheckFailure):
+            logging.debug(
+                f"{ctx.author} the command '{ctx.command}', but they did not have "
+                f"permissions to run commands in the channel {ctx.channel}!"
+            )
+            embed = Embed(colour=Colour.red())
+            embed.title = random.choice(NEGATIVE_REPLIES)
+            embed.description = str(error)
+            return await ctx.send(embed=embed)
```
MarkKoz: Using an embed is not consistent with the rest of the errors which are just plain messages.

bot/decorators.py
```diff
@@ -48,14 +55,53 @@ def without_role(*role_ids: int):
...
+def in_channel_check(*channels: int, bypass_roles: typing.Container[int] = None) -> typing.Callable[[Context], bool]:
+    """Checks that the message is in a whitelisted channel or optionally has a bypass role."""
+    def predicate(ctx: Context) -> bool:
+        if not ctx.guild:
+            log.debug(f"{ctx.author} tried to use the '{ctx.command.name}' command from a DM.")
+            return True
+
+        if ctx.channel.id in channels:
+            log.debug(f"{ctx.author} tried to call the '{ctx.command.name}' command. "
+                      f"The command was used in a whitelisted channel.")
```
MarkKoz: I believe the prevalent style is newline after the opening parenthesis and before the closing parenthesis.

bot/decorators.py
```diff
@@ -48,14 +55,53 @@ def without_role(*role_ids: int):
...
+
+        channels_str = ', '.join(f"<#{c_id}>" for c_id in channels)
+        raise InChannelCheckFailure(
+            f"Sorry, but you may only use this command within {channels_str}."
+        )
+
+    return predicate
+
+
+in_channel = commands.check(in_channel_check)
```
MarkKoz: What's this for?

bot/decorators.py
```diff
@@ -48,14 +55,53 @@ def without_role(*role_ids: int):
...
+            return True
+
+        if hasattr(ctx.command.callback, "in_channel_override"):
+            log.debug(f"{ctx.author} tried to call the '{ctx.command.name}' command. "
+                      f"The command was not used in a whitelisted channel, "
+                      f"but the command was whitelisted to bypass the in_channel check.")
+            return True
+
+        if bypass_roles:
+            if any(r.id in bypass_roles for r in ctx.author.roles):
```
MarkKoz: Can this be merged with `and` into the if statement above?

bot/decorators.py
```diff
@@ -48,14 +55,53 @@ def without_role(*role_ids: int):
...
+                log.debug(f"{ctx.author} tried to call the '{ctx.command.name}' command. "
+                          f"The command was not used in a whitelisted channel, "
+                          f"but the author had a role to bypass the in_channel check.")
```
MarkKoz: Personal taste but these log messages are too verbose for my liking.

bot/seasons/evergreen/error_handler.py
kosayoda: I wanted it to be consistent with Python bot's wrong channel error message. Do you think the rest of the errors should be converted to an embed (they're not long error messages) to be consistent with Python bot? Or should we keep SeasonalBot's errors as a regular message?

bot/decorators.py
kosayoda: This will be fixed.

bot/decorators.py
kosayoda: This is taken from the Python bot. How do you propose the statement be?

bot/seasons/evergreen/error_handler.py
MarkKoz: I'm not sure. This issue is still open python-discord/bot#126 and it's not entirely clear if we want to abolish embeds but in recent PRs it seems to have been going in that direction. Maybe best to leave this alone and come back to it later. At the least I can say you shouldn't change everything else to an embed.

bot/decorators.py
kosayoda: My implementation adds the check to the bot using `bot.add_check`, which takes a predicate function that accepts a `Context` and returns whether or not it passes the check. Currently, the predicates are passed to `commands.check` before returning, which returns a decorator suitable to be added to individual commands. `in_channel` is still the decorator, preserving functionality, but now the predicate check can be reused and added as a global check to the bot.
It’s explained in the PR message and the commit message, I wasn’t clear about it though.

bot/decorators.py
MarkKoz: Something like
> {ctx.author} called '{ctx.command.name}' and bypassed the in_channel check

Maybe throw in there that it's specifically because of a role.

bot/decorators.py
kosayoda: It could, I nabbed it from the Python bot without looking through it much. I’ll test and change this.

bot/decorators.py
MarkKoz: Basically `in_channel` is meant to be used as a decorator and `in_channel_check` is for use with `bot.add_check()`?

kosayoda: Yes.

bot/seasons/evergreen/error_handler.py
kosayoda: I have left this alone, the rest has been updated.

PR #253, opened 2019-08-08
snake and ladder game fix.

### LINKED ISSUE #252: Snake and ladders(sal) not stopping.
Recently encountered an error with the sal game where the game would not respond to any reaction, its been running and cannot be stopped due to some reason.

## Steps to Reproduce
possible fix:
- Make a timer after which the game shuts down automatically due to inactivity.

## Other
Comment if you got any any other ideas for this fix.

RohanJnr:
closes #252 

fixed the issue of the game not stopping.
Now the server mods also can step in and stop the game.


bot/seasons/evergreen/snakes/utils.py
```diff
@@ -532,8 +535,9 @@ def startup_event_check(reaction_: Reaction, user_: Member):
 
     async def cancel_game(self, user: Member):
         """Allow the game author to cancel the running game."""
-        if not user == self.author:
-            await self.channel.send(user.mention + " Only the author of the game can cancel it.", delete_after=10)
+        if not user == self.author and Roles.moderator not in [role.id for role in user.roles]:
+            await self.channel.send(user.mention + " Only the author of the game and server mods can cancel it.",
```
SebastiaanZ: In this case, the preferred indentation style in our repos would be:

```py
await self.channel.send(
    user.mention + " Only the author of the game and server mods can cancel it.",
    delete_after=10,
)
```

Like here: https://github.com/python-discord/seasonalbot/blob/abec4415d73ee5838c79e0a488d914589ae20ac2/bot/seasons/evergreen/snakes/utils.py#L556-L559

I also think that more recent additions to either bot or seasonalbot would use an `f-string` here, but that's not the prevalent style in this particular file. (The code here is quite old in PyDis terms.)

RohanJnr:
> 
> 
> This does not fully fix the issue, there is still are still missing permission checks on the reaction add:
> 
> https://github.com/python-discord/seasonalbot/blob/e906696dbc880f283146cb04aa8916f04f8ab545/bot/seasons/evergreen/snakes/utils.py#L449-L454
> 
> https://github.com/python-discord/seasonalbot/blob/e906696dbc880f283146cb04aa8916f04f8ab545/bot/seasons/evergreen/snakes/utils.py#L637-L642
> 
> This needs to account for the possibility of the moderator being part of the game & just wants to leave rather than cancel the entire game.
> 
> edit: I don't see any reason why there needs to be a user check in `cancel_game` at all, it's only ever invoked via reaction or via timeout. The permission check only needs to happen when a user clicks the reaction.

the code is using the same reaction to close game or to exit from the game if not author, should we add another reaction for leaving? because if remove that user check, then if a mod is playing and wants to leave,he/she will end up closing the game.

lemonsaurus:
The game should not end just because the author leaves. What if there are 3 other players and they want to keep playing?

Let's instead have the game end only if one of these two conditions are met:
1) Game has timed out.
2) There is only one player left. The game cannot be played with less than 2 players.

sco1:
> because if remove that user check, then if a mod is playing and wants to leave,he/she will end up closing the game

Correct, which is why I said there needs to be logic to account for this. You have a list of players you can reference to see if the mod is part of the game or not.

lemonsaurus:
Okay, keeping in mind the mod idea, a third condition for closing the game occurs:

1. Game has timed out.
2. There is only one player left. The game cannot be played with less than 2 players.
3. A moderator clicks on the cancel button and _is not currently playing the game_. If the moderator is playing, they will simply leave the game.

SebastiaanZ:
>     2. There is only one player left. The game cannot be played with less than 2 players.

You can currently start it with only one player, too. If we want to prohibit people playing it by themselves, we should prevent the game from starting with a single player, too. (Since I assume that we'll only check for a stop condition when someone leaves.)

Anyway, if people play it by themselves, what harm does it do? Why not just stop the game when it drops to zero players?

lemonsaurus:
:man_shrugging: I don't have a good reason why we shouldn't allow people to play solo, so let's allow that I guess.

RohanJnr:
> 
> 
> 🤷‍♂ I don't have a good reason why we shouldn't allow people to play solo, so let's allow that I guess.

who plays sal alone?

SebastiaanZ:
> > man_shrugging I don't have a good reason why we shouldn't allow people to play solo, so let's allow that I guess.
> 
> who plays sal alone?

People that are trying out the commands to see what they are about. It would kinda suck if you can't see what it looks like because no one else wants to join you at that time.

RohanJnr:
I am going to close this now and reopen it later.

PR #254, opened 2019-08-09
Add speedrun cog and resource

jakeHebert:
Created a speedrun cog that allows the command .speedrun which will grab a link from the supplied speedrun_links.json and send it to the discord channel.

### DIFF
--- a/bot/resources/evergreen/speedrun_links.json
+++ b/bot/resources/evergreen/speedrun_links.json
@@ -1,0 +1,20 @@
+{
+    "links": [
+        "https://www.youtube.com/watch?v=jNE28SDXdyQ",
+        "https://www.youtube.com/watch?v=iI8Giq7zQDk",
+        "https://www.youtube.com/watch?v=VqNnkqQgFbc",
+        "https://www.youtube.com/watch?v=Gum4GI2Jr0s",
+        "https://www.youtube.com/watch?v=5YHjHzHJKkU",
+        "https://www.youtube.com/watch?v=X0pJSTy4tJI",
+        "https://www.youtube.com/watch?v=aVFq0H6D6_M",
+        "https://www.youtube.com/watch?v=1O6LuJbEbSI",
+        "https://www.youtube.com/watch?v=Bgh30BiWG58",
+        "https://www.youtube.com/watch?v=wwvgAAvhxM8",
+        "https://www.youtube.com/watch?v=0TWQr0_fi80",
+        "https://www.youtube.com/watch?v=hatqZby-0to",
+        "https://www.youtube.com/watch?v=X0pJSTy4tJI",
+        "https://www.youtube.com/watch?v=tmnMq2Hw72w",
+        "https://www.youtube.com/watch?v=UTkyeTCAucA",
+        "https://www.youtube.com/watch?v=67kQ3l-1qMs"
+    ]
+}
--- a/bot/resources/evergreen/speedrun_links.json
+++ b/bot/resources/evergreen/speedrun_links.json
@@ -1,5 +1,4 @@
-{
-    "links": [
+ [
         "https://www.youtube.com/watch?v=jNE28SDXdyQ",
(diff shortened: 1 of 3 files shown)


bot/resources/evergreen/speedrun_links.json
```diff
@@ -0,0 +1,20 @@
...
+        "https://www.youtube.com/watch?v=Gum4GI2Jr0s",
+        "https://www.youtube.com/watch?v=5YHjHzHJKkU",
+        "https://www.youtube.com/watch?v=X0pJSTy4tJI",
+        "https://www.youtube.com/watch?v=aVFq0H6D6_M",
+        "https://www.youtube.com/watch?v=1O6LuJbEbSI",
+        "https://www.youtube.com/watch?v=Bgh30BiWG58",
+        "https://www.youtube.com/watch?v=wwvgAAvhxM8",
+        "https://www.youtube.com/watch?v=0TWQr0_fi80",
+        "https://www.youtube.com/watch?v=hatqZby-0to",
+        "https://www.youtube.com/watch?v=X0pJSTy4tJI",
```
MarkKoz: This is a duplicate.

jakeHebert: Removed the duplicate punch out link and added the link you requested.

jakeHebert:
I've got most of the issues fixed, except converting the dict to an array. I'll finish working on that this afternoon.

bot/seasons/evergreen/speedrun.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command(name="speedrun")
+    async def get_speedrun(self, ctx):
+        """Sends a link to a video of a random speedrun."""
+        with open(Path('bot/resources/evergreen/speedrun_links.json')) as file:
+            data = json.load(file)
+        # links = data['links']
```
MarkKoz: Would be nice to get rid of this comment
```suggestion
```

bot/seasons/evergreen/speedrun.py
```diff
@@ -0,0 +1,29 @@
...
+class Speedrun(commands.Cog):
+    """Commands about the video game speedrunning community."""
+
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command(name="speedrun")
+    async def get_speedrun(self, ctx):
+        """Sends a link to a video of a random speedrun."""
+        with open(Path('bot/resources/evergreen/speedrun_links.json')) as file:
```
MarkKoz: To be safe, the encoding should be explicitly specified. Also, I think `Path().open()` is cleaner than `open(Path())`

```suggestion
        with Path('bot/resources/evergreen/speedrun_links.json').open(encoding="utf-8") as file:
```

MarkKoz: Sorry for not catching this with my first review. It would be better to load the JSON file only once when the cog is loaded rather than loading it every time the command is called.

bot/seasons/evergreen/speedrun.py
jakeHebert: Ahhh! Totally meant to do that!

bot/seasons/evergreen/speedrun.py
jakeHebert: Good point, would it make sense to do this on init, creating a self.data to hold the value? Or should I be trying to do this in the setup function?

MarkKoz: You can either do it in `__init__` or do it at the top of the module and store it as a module-level constant. Both ways have been used for other features on seasonalbot. It's up to you.

sco1: Module level is preferred

jakeHebert: Whoops, @sco1 just saw your comment after I pushed the commit. I'll get it moved to the module level. I'm also going to change its name from `data` to `LINKS` so it's a bit more descriptive.

PR #255, opened 2019-08-10
Minesweeper

vivax3794:
added the minesweeper command,

it is played in dms, and a mirror of the board is updated in the channel the command was run in. 
it is 10 by 10, you can specify the chance of a cell being a bomb (defaults to .2) must be between 0 and 1.

### DIFF
--- a/bot/seasons/evergreen/minesweeper.py
+++ b/bot/seasons/evergreen/minesweeper.py
@@ -31,42 +31,46 @@ log = logging.getLogger(__name__)
 class CoordinateConverter(commands.Converter):
     """Converter for Coordinates."""
 
-    async def convert(self, ctx, Coordinate: str) -> typing.Tuple[int, int]:
-        """Take in a Coordinate string and turn it into x, y"""
-        if not 2 <= len(Coordinate) <= 3:
+    async def convert(self, ctx, coordinate: str) -> typing.Tuple[int, int]:
+        """Take in a coordinate string and turn it into x, y"""
+        if not 2 <= len(coordinate) <= 3:
             raise commands.ArgumentParsingError()
-        value1 = Coordinate[0]
-        value2 = Coordinate[1:]
+
+        value1 = coordinate[0]
+        value2 = coordinate[1:]
+
         if not value2.isdigit():
             raise commands.ArgumentParsingError()
+
         x = ord(value1) - 97
         y = int(value2) - 1
+
         if (not 0 <= x <= 9) or (not 0 <= y <= 9):
             raise commands.ArgumentParsingError()
         return x, y
 
 
-GameDict = typing.List[typing.List[typing.Union[str, int]]]
+GameBoard = typing.List[typing.List[typing.Union[str, int]]]
 
 
 @dataclass
 class Game:
     """The data for a game."""
 
-    board: GameDict
-    revealed: GameDict
+    board: GameBoard
+    revealed: GameBoard
     dm_msg: discord.message
     chat_msg: discord.message
 
 
-DictOfGames = typing.Dict[int, Game]
+GamesDict = typing.Dict[int, Game]
 
 
 class Minesweeper(commands.Cog):
     """Play a game of Minesweeper."""
 
     def __init__(self, bot: commands.Bot) -> None:
-        self.games: DictOfGames = {}  # Store the currently running games
+        self.games: GamesDict = {}  # Store the currently running games
 
     @staticmethod
     def get_neighbours(x: int, y: int) -> typing.Generator:
@@ -76,9 +80,9 @@ class Minesweeper(commands.Cog):
                 if x_ != -1 and x_ != 10 and y_ != -1 and y_ != 10:
                     yield x_, y_
 
-    def generate_board(self, bomb_chance: float) -> GameDict:
+    def generate_board(self, bomb_chance: float) -> GameBoard:
         """Generate a 2d array for the board."""
-        board: GameDict = [["bomb" if random() <= bomb_chance else "number" for _ in range(10)] for _ in range(10)]
+        board: GameBoard = [["bomb" if random() <= bomb_chance else "number" for _ in range(10)] for _ in range(10)]
         for y, row in enumerate(board):
             for x, cell in enumerate(row):
                 if cell == "number":
@@ -91,7 +95,7 @@ class Minesweeper(commands.Cog):
         return board
 
     @staticmethod
-    def format_for_discord(board: GameDict) -> str:
+    def format_for_discord(board: GameBoard) -> str:
         """Format the board as a string for Discord."""
         discord_msg = (
             ":stop_button:    :regional_indicator_a::regional_indicator_b::regional_indicator_c:"
@@ -118,8 +122,8 @@ class Minesweeper(commands.Cog):
             return
 
         # Add game to list
-        board: GameDict = self.generate_board(bomb_chance)
-        revealed_board: GameDict = [["hidden"] * 10 for _ in range(10)]
+        board: GameBoard = self.generate_board(bomb_chance)
+        revealed_board: GameBoard = [["hidden"] * 10 for _ in range(10)]
 
         await ctx.send(f"{ctx.author.mention} is playing Minesweeper")
         chat_msg = await ctx.send(self.format_for_discord(revealed_board))
@@ -147,10 +151,10 @@ class Minesweeper(commands.Cog):
 
     @commands.dm_only()
     @commands.command(name="flag")
-    async def flag_command(self, ctx: commands.Context, *Coordinates: CoordinateConverter) -> None:
+    async def flag_command(self, ctx: commands.Context, *coordinates: CoordinateConverter) -> None:
         """Place multiple flags on the board"""
-        board: GameDict = self.games[ctx.author.id].revealed
-        for x, y in Coordinates:
+        board: GameBoard = self.games[ctx.author.id].revealed
+        for x, y in coordinates:
             if board[y][x] == "hidden":
                 board[y][x] = "flag"
 
@@ -170,7 +174,7 @@ class Minesweeper(commands.Cog):
         await ctx.author.send(":tada:  You won! :tada: ")
         await game.chat_msg.channel.send(f":tada: {ctx.author.mention} just won minesweeper :tada:")
 
-    def reveal_zeros(self, revealed: GameDict, board: GameDict, x: int, y: int) -> None:
+    def reveal_zeros(self, revealed: GameBoard, board: GameBoard, x: int, y: int) -> None:
         """Recursively reveal adjacent cells when a 0 cell is encountered."""
         for x_, y_ in self.get_neighbours(x, y):
             if revealed[y_][x_] != "hidden":
@@ -179,7 +183,7 @@ class Minesweeper(commands.Cog):
             if board[y_][x_] == 0:
                 self.reveal_zeros(revealed, board, x_, y_)
 
-    async def check_if_won(self, ctx, revealed: GameDict, board: GameDict) -> bool:
+    async def check_if_won(self, ctx, revealed: GameBoard, board: GameBoard) -> bool:
         """Checks if a player has won"""
         for x in range(10):
             for y in range(10):
@@ -189,26 +193,29 @@ class Minesweeper(commands.Cog):
             await self.won(ctx)
             return True
 
-    async def reveal_one(self, ctx: commands.Context, revealed: GameDict, board: GameDict, x: int, y: int) -> bool:
-        """Reveal one square."""
+    async def reveal_one(self, ctx: commands.Context, revealed: GameBoard, board: GameBoard, x: int, y: int) -> bool:
+        """
+        Reveal one square.
+
+        return is True if the game ended, breaking the loop in `reveal_command` and deleting the game
+        """
         revealed[y][x] = board[y][x]
         if board[y][x] == "bomb":
             await self.lost(ctx)
             return True  # game ended
         elif board[y][x] == 0:
             self.reveal_zeros(revealed, board, x, y)
-        won = await self.check_if_won(ctx, revealed, board)
-        return won
+        return await self.check_if_won(ctx, revealed, board)
 
     @commands.dm_only()
     @commands.command(name="reveal")
-    async def reveal_command(self, ctx: commands.Context, *Coordinates: CoordinateConverter) -> None:
+    async def reveal_command(self, ctx: commands.Context, *coordinates: CoordinateConverter) -> None:
         """Reveal multiple cells"""
         game = self.games[ctx.author.id]
-        revealed: GameDict = game.revealed
-        board: GameDict = game.board
+        revealed: GameBoard = game.revealed
+        board: GameBoard = game.board
 
-        for x, y in Coordinates:
+        for x, y in coordinates:
             if await self.reveal_one(ctx, revealed, board, x, y):  # game ended
                 await self.update_boards(ctx)
                 del self.games[ctx.author.id]
@@ -222,8 +229,9 @@ class Minesweeper(commands.Cog):
         game = self.games[ctx.author.id]
         game.revealed = game.board
         await self.update_boards(ctx)
-        await ctx.author.send(":no_entry: you canceled the game :no_entry:")
-        await game.chat_msg.channel.send(f"{ctx.author.mention} just canceled minesweeper")
+        new_msg = f":no_entry: Game canceled :no_entry:\n{self.format_for_discord(game.revealed)}"
+        await game.dm_msg.edit(content=new_msg)
+        await game.chat_msg.edit(content=new_msg)
         del self.games[ctx.author.id]
 
 
--- a/bot/seasons/evergreen/minesweeper.py
+++ b/bot/seasons/evergreen/minesweeper.py
@@ -84,7 +84,8 @@ class Minesweeper(commands.Cog):
         await ctx.send(f"{ctx.author.mention} is playing minesweeper")
         chat_msg = await ctx.send(self.format_for_discord(reveled_board))
 
-        await ctx.author.send("play by typing: `.reveal xy xy ...` or `.flag xy xy ...` \nclose the game with `.end`")
+        await ctx.author.send("play by typing: `.reveal xy xy ...` or `.flag xy xy ...` \nclose the game with `.end`\n"
+                              "cords must be in format `<letter><number>")
         dm_msg = await ctx.author.send(self.format_for_discord(reveled_board))
 
         self.games[ctx.author] = {
@@ -97,10 +98,7 @@ class Minesweeper(commands.Cog):
     @staticmethod
     def get_cords(value1: str, value2: str) -> typing.Tuple[int, int]:
         """Take in 2 values for the cords and turn them into numbers"""
-        if value1.isnumeric():
-            return ord(value2.lower()) - 97, int(value1) - 1
-        else:
-            return ord(value1.lower()) - 97, int(value2) - 1
+        return ord(value1.lower()) - 97, int(value2) - 1
 
     async def reload_board(self, ctx: commands.Context) -> None:
         """Update both playing boards."""
@@ -115,7 +113,7 @@ class Minesweeper(commands.Cog):
         """Place multiple flags on the board"""
         board = self.games[ctx.author]["reveled"]
         for cord in cords:
-            x, y = self.get_cords(cord[0], cord[1])
+            x, y = self.get_cords(cord[0], cord[1:])
             if board[y][x] == "hidden":
                 board[y][x] = "flag"
 
@@ -178,7 +176,7 @@ class Minesweeper(commands.Cog):
         reveled = game["reveled"]
         board = game["board"]
         for cord in cords:
-            x, y = self.get_cords(cord[0], cord[1])
+            x, y = self.get_cords(cord[0], cord[1:])
             await self.reveal_one(ctx, reveled, board, x, y)
 
     @commands.dm_only()
--- a/bot/seasons/evergreen/minesweeper.py
+++ b/bot/seasons/evergreen/minesweeper.py
@@ -1,42 +1,41 @@
 import typing
 from random import random
 
-import discord
 from discord.ext import commands
+
+GameBoard = typing.List[typing.List[typing.Union[str, int]]]
+DictOfGames = typing.Dict[int, typing.Dict]
 
 
 class Minesweeper(commands.Cog):
     """Play a game of minesweeper."""
 
     def __init__(self, bot: commands.Bot) -> None:
-        self.bot = bot
-        self.games: typing.Dict[discord.member, typing.Dict] = {}  # Store the currently running games
+        self.games: DictOfGames = {}  # Store the currently running games
 
     @staticmethod
     def is_bomb(cell: typing.Union[str, int]) -> int:
         """Returns 1 if `cell` is a bomb if not 0"""
-        return 1 if cell == "bomb" else 0
+        return cell == "bomb"
 
-    def generate_board(self, bomb_chance: float) -> typing.List[typing.List[typing.Union[str, int]]]:
+    def generate_board(self, bomb_chance: float) -> GameBoard:
         """Generate a 2d array for the board."""
-        board: typing.List[typing.List[typing.Union[str, int]]] = [
-            ["bomb" if random() <= bomb_chance else "number" for _ in range(10)] for _ in range(10)]
+        board: GameBoard = [["bomb" if random() <= bomb_chance else "number" for _ in range(10)] for _ in range(10)]
         for y, row in enumerate(board):
             for x, cell in enumerate(row):
                 if cell == "number":
                     # calculate bombs near it
-                    to_check = []
-                    for xt in [x - 1, x, x + 1]:
-                        for yt in [y - 1, y, y + 1]:
-                            if xt != -1 and xt != 10 and yt != -1 and yt != 10:
-                                to_check.append(board[yt][xt])
+                    bombs = 0
+                    for x_ in [x - 1, x, x + 1]:
+                        for y_ in [y - 1, y, y + 1]:
+                            if x_ != -1 and x_ != 10 and y_ != -1 and y_ != 10 and board[y_][x_] == "bomb":
+                                bombs += 1
 
-                    bombs = sum(map(self.is_bomb, to_check))
                     board[y][x] = bombs
         return board
 
     @staticmethod
-    def format_for_discord(board: typing.List[typing.List[typing.Union[str, int]]]) -> str:
+    def format_for_discord(board: GameBoard) -> str:
         """Format the board to a string for discord."""
         mapping = {
             0: ":stop_button:",
@@ -55,9 +54,11 @@ class Minesweeper(commands.Cog):
             "flag": ":triangular_flag_on_post:"
         }
 
-        discord_msg = ":stop_button:    :regional_indicator_a::regional_indicator_b::regional_indicator_c:" \
-                      ":regional_indicator_d::regional_indicator_e::regional_indicator_f::regional_indicator_g:" \
-                      ":regional_indicator_h::regional_indicator_i::regional_indicator_j:\n\n"
+        discord_msg = (
+            ":stop_button:    :regional_indicator_a::regional_indicator_b::regional_indicator_c:"
+            ":regional_indicator_d::regional_indicator_e::regional_indicator_f::regional_indicator_g:"
+            ":regional_indicator_h::regional_indicator_i::regional_indicator_j:\n\n"
+        )
         rows: typing.List[str] = []
         for row_number, row in enumerate(board):
             new_row = mapping[row_number + 1] + "    "
@@ -71,26 +72,27 @@ class Minesweeper(commands.Cog):
     @commands.command(name="minesweeper")
     async def minesweeper_command(self, ctx: commands.Context, bomb_chance: float = .2) -> None:
         """Start a game of minesweeper."""
-        if ctx.author in self.games.keys():  # Player is already playing
+        if ctx.author.id in self.games.keys():  # Player is already playing
             msg = await ctx.send(f"{ctx.author.mention} you already have a game running")
             await msg.delete(delay=2)
             await ctx.message.delete(delay=2)
             return
 
         # Add game to list
-        board = self.generate_board(bomb_chance)
-        reveled_board = [["hidden" for _ in range(10)] for _ in range(10)]
+        board: GameBoard = self.generate_board(bomb_chance)
+        revealed_board: GameBoard = [["hidden" for _ in range(10)] for _ in range(10)]
 
         await ctx.send(f"{ctx.author.mention} is playing minesweeper")
-        chat_msg = await ctx.send(self.format_for_discord(reveled_board))
+        chat_msg = await ctx.send(self.format_for_discord(revealed_board))
 
-        await ctx.author.send("play by typing: `.reveal xy xy ...` or `.flag xy xy ...` \nclose the game with `.end`\n"
-                              "cords must be in format `<letter><number>")
-        dm_msg = await ctx.author.send(self.format_for_discord(reveled_board))
+        await ctx.author.send("play by typing: `.reveal xy [xy]` or `.flag xy [xy]` \n"
+                              "close the game with `.end`\n"
+                              "cords must be in format `<letter><number>`")
+        dm_msg = await ctx.author.send(self.format_for_discord(revealed_board))
 
-        self.games[ctx.author] = {
+        self.games[ctx.author.id] = {
             "board": board,
-            "reveled": reveled_board,
+            "revealed": revealed_board,
             "dm_msg": dm_msg,
             "chat_msg": chat_msg
         }
@@ -102,16 +104,16 @@ class Minesweeper(commands.Cog):
 
     async def reload_board(self, ctx: commands.Context) -> None:
         """Update both playing boards."""
-        game = self.games[ctx.author]
+        game = self.games[ctx.author.id]
         await game["dm_msg"].delete()
-        game["dm_msg"] = await ctx.author.send(self.format_for_discord(game["reveled"]))
-        await game["chat_msg"].edit(content=self.format_for_discord(game["reveled"]))
+        game["dm_msg"] = await ctx.author.send(self.format_for_discord(game["revealed"]))
+        await game["chat_msg"].edit(content=self.format_for_discord(game["revealed"]))
 
     @commands.dm_only()
     @commands.command(name="flag")
     async def flag_command(self, ctx: commands.Context, *cords) -> None:
         """Place multiple flags on the board"""
-        board = self.games[ctx.author]["reveled"]
+        board: GameBoard = self.games[ctx.author.id]["revealed"]
         for cord in cords:
             x, y = self.get_cords(cord[0], cord[1:])
             if board[y][x] == "hidden":
@@ -121,74 +123,73 @@ class Minesweeper(commands.Cog):
 
     async def lost(self, ctx: commands.Context) -> None:
         """The player lost the game"""
-        game = self.games[ctx.author]
-        game["reveled"] = game["board"]
+        game = self.games[ctx.author.id]
+        game["revealed"] = game["board"]
         await self.reload_board(ctx)
         await ctx.author.send(":fire: You lost :fire: ")
         await game["chat_msg"].channel.send(f":fire: {ctx.author.mention} just lost minesweeper :fire:")
-        del self.games[ctx.author]
+        del self.games[ctx.author.id]
 
-    async def won(self, ctx: commands.Context):
+    async def won(self, ctx: commands.Context) -> None:
         """The player won the game"""
-        game = self.games[ctx.author]
-        game["reveled"] = game["board"]
+        game = self.games[ctx.author.id]
+        game["revealed"] = game["board"]
         await self.reload_board(ctx)
         await ctx.author.send(":tada:  You won! :tada: ")
         await game["chat_msg"].channel.send(f":tada: {ctx.author.mention} just won minesweeper :tada:")
-        del self.games[ctx.author]
+        del self.games[ctx.author.id]
 
-    def reveal_zeros(self, reveled: typing.List, board: typing.List, x: int, y: int) -> None:
+    def reveal_zeros(self, revealed: GameBoard, board: GameBoard, x: int, y: int) -> None:
         """Used when a 0 is encountered to do a flood fill"""
         for x_ in [x - 1, x, x + 1]:
             for y_ in [y - 1, y, y + 1]:
-                if x_ == -1 or x_ == 10 or y_ == -1 or y_ == 10 or reveled[y_][x_] != "hidden":
+                if x_ == -1 or x_ == 10 or y_ == -1 or y_ == 10 or revealed[y_][x_] != "hidden":
                     continue
-                reveled[y_][x_] = board[y_][x_]
+                revealed[y_][x_] = board[y_][x_]
                 if board[y_][x_] == 0:
-                    self.reveal_zeros(reveled, board, x_, y_)
+                    self.reveal_zeros(revealed, board, x_, y_)
 
-    async def reveal_one(self, ctx: commands.Context, reveled: typing.List, board: typing.List, x: int, y: int) -> None:
+    async def check_if_won(self, ctx, revealed: GameBoard, board: GameBoard) -> None:
+        """Checks if a player has won"""
+        for x_ in range(10):
+            for y_ in range(10):
+                if revealed[y_][x_] == "hidden" and board[y_][x_] != "bomb":
+                    return
+        else:
+            await self.won(ctx)
+
+    async def reveal_one(self, ctx: commands.Context, revealed: GameBoard, board: GameBoard, x: int, y: int) -> None:
         """Reveal one square."""
-        reveled[y][x] = board[y][x]
+        revealed[y][x] = board[y][x]
         if board[y][x] == "bomb":
             await self.lost(ctx)
         elif board[y][x] == 0:
-            self.reveal_zeros(reveled, board, x, y)
+            self.reveal_zeros(revealed, board, x, y)
+        await self.check_if_won(ctx, revealed, board)
 
-        # check if won
-        break_ = False
-        for x_ in range(10):
-            for y_ in range(10):
-                if reveled[y_][x_] == "hidden" and board[y_][x_] != "bomb":
-                    break_ = True
-                    break
-            if break_:
-                break
-        else:
-            await self.won(ctx)
         await self.reload_board(ctx)
 
     @commands.dm_only()
     @commands.command(name="reveal")
-    async def reveal_command(self, ctx: commands.Context, *cords):
+    async def reveal_command(self, ctx: commands.Context, *cords) -> None:
         """Reveal multiple cells"""
-        game = self.games[ctx.author]
-        reveled = game["reveled"]
-        board = game["board"]
+        game = self.games[ctx.author.id]
+        revealed: GameBoard = game["revealed"]
+        board: GameBoard = game["board"]
         for cord in cords:
             x, y = self.get_cords(cord[0], cord[1:])
-            await self.reveal_one(ctx, reveled, board, x, y)
+            await self.reveal_one(ctx, revealed, board, x, y)
 
     @commands.dm_only()
     @commands.command(name="end")
     async def end_command(self, ctx: commands.Context):
         """End the current game"""
-        game = self.games[ctx.author]
-        game["reveled"] = game["board"]
+        game = self.games[ctx.author.id]
+        game["revealed"] = game["board"]
         await self.reload_board(ctx)
         await ctx.author.send(":no_entry: you canceled the game :no_entry:")
         await game["chat_msg"].channel.send(f"{ctx.author.mention} just canceled minesweeper")
-        del self.games[ctx.author]
+        del self.games[ctx.author.id]
 
 
 def setup(bot: commands.Bot) -> None:
--- a/bot/seasons/evergreen/minesweeper.py
+++ b/bot/seasons/evergreen/minesweeper.py
@@ -1,11 +1,30 @@
+import typing
+
+import discord
 from discord.ext import commands
 
 
 class Minesweeper(commands.Cog):
-    """play a game of minesweeper"""
+    """Play a game of minesweeper."""
 
     def __init__(self, bot: commands.Bot) -> None:
         self.bot = bot
+        self.games: typing.Dict[discord.member, typing.Dict] = {}  # Store the currently running games
+
+    @commands.command(name="minesweeper")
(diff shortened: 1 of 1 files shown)


vivax3794:
I think I found a bug.

vivax3794:
I think there still is a bug, going to look into it.

vivax3794:
I solved the bug

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+    async def reveal_one(self, ctx: commands.Context, reveled: typing.List, board: typing.List, x: int, y: int) -> None:
+        """Reveal one square."""
+        reveled[y][x] = board[y][x]
+        if board[y][x] == "bomb":
+            await self.lost(ctx)
+        elif board[y][x] == 0:
+            self.reveal_zeros(reveled, board, x, y)
+
+        # check if won
+        break_ = False
```
tagptroll1: Would it be better to use a function for this so you can return instead of double breaking?

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+import discord
+from discord.ext import commands
+
+
+class Minesweeper(commands.Cog):
+    """Play a game of minesweeper."""
+
+    def __init__(self, bot: commands.Bot) -> None:
+        self.bot = bot
+        self.games: typing.Dict[discord.member, typing.Dict] = {}  # Store the currently running games
```
tagptroll1: Could this typing be stored as a variable at the top of the file and named something like DictOfGames to avoid the long typing and comment explaining it here?

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+
+import discord
+from discord.ext import commands
+
+
+class Minesweeper(commands.Cog):
+    """Play a game of minesweeper."""
+
+    def __init__(self, bot: commands.Bot) -> None:
+        self.bot = bot
```
tagptroll1: `self.bot` is not used anywhere in the file, maybe just exclude it?

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+        self.games: typing.Dict[discord.member, typing.Dict] = {}  # Store the currently running games
+
+    @staticmethod
+    def is_bomb(cell: typing.Union[str, int]) -> int:
+        """Returns 1 if `cell` is a bomb if not 0"""
+        return 1 if cell == "bomb" else 0
```
tagptroll1: This could probably just be `return cell=="bomb"` as `True` is basically the same as `1`

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+            for x, cell in enumerate(row):
+                if cell == "number":
+                    # calculate bombs near it
+                    to_check = []
+                    for xt in [x - 1, x, x + 1]:
+                        for yt in [y - 1, y, y + 1]:
+                            if xt != -1 and xt != 10 and yt != -1 and yt != 10:
+                                to_check.append(board[yt][xt])
+
+                    bombs = sum(map(self.is_bomb, to_check))
```
tagptroll1: Personally prefer a generator / list comprehension here as `sum(self.is_bomb(cell) for cell in to_check)` or better yet just check the cell in the loop to prevent 2 iteration?  You don't use `to_check` anywhere else that i can see, maybe a `collections.Counter`?

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+        self.games: typing.Dict[discord.member, typing.Dict] = {}  # Store the currently running games
+
+    @staticmethod
+    def is_bomb(cell: typing.Union[str, int]) -> int:
+        """Returns 1 if `cell` is a bomb if not 0"""
+        return 1 if cell == "bomb" else 0
+
+    def generate_board(self, bomb_chance: float) -> typing.List[typing.List[typing.Union[str, int]]]:
+        """Generate a 2d array for the board."""
+        board: typing.List[typing.List[typing.Union[str, int]]] = [
```
tagptroll1: This line is a bit messy, a few linebreaks and maybe a variable for the typings would be nice

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+            7: ":seven:",
+            8: ":eight:",
+            9: ":nine:",
+            10: ":keycap_ten:",
+            "bomb": ":bomb:",
+            "hidden": ":grey_question:",
+            "flag": ":triangular_flag_on_post:"
+        }
+
+        discord_msg = ":stop_button:    :regional_indicator_a::regional_indicator_b::regional_indicator_c:" \
```
tagptroll1: Use () for wrapping code instead of `\`

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+                      ":regional_indicator_d::regional_indicator_e::regional_indicator_f::regional_indicator_g:" \
+                      ":regional_indicator_h::regional_indicator_i::regional_indicator_j:\n\n"
+        rows: typing.List[str] = []
+        for row_number, row in enumerate(board):
+            new_row = mapping[row_number + 1] + "    "
```
tagptroll1: string formatting instead of concatenation ?

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+
+        discord_msg += "\n".join(rows)
+        return discord_msg
+
+    @commands.command(name="minesweeper")
+    async def minesweeper_command(self, ctx: commands.Context, bomb_chance: float = .2) -> None:
+        """Start a game of minesweeper."""
+        if ctx.author in self.games.keys():  # Player is already playing
+            msg = await ctx.send(f"{ctx.author.mention} you already have a game running")
+            await msg.delete(delay=2)
```
tagptroll1: You can also add this as a parameter to send `ctx.send("..", delete_after= x)` but not a big deal

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+            return
+
+        # Add game to list
+        board = self.generate_board(bomb_chance)
+        reveled_board = [["hidden" for _ in range(10)] for _ in range(10)]
+
+        await ctx.send(f"{ctx.author.mention} is playing minesweeper")
+        chat_msg = await ctx.send(self.format_for_discord(reveled_board))
+
+        await ctx.author.send("play by typing: `.reveal xy xy ...` or `.flag xy xy ...` \nclose the game with `.end`")
```
tagptroll1: This could probably be displayed ad `.reveal xy [xy]` or something similar to display it's optional or just say you can use as many coordinates as you want with the command

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+        dm_msg = await ctx.author.send(self.format_for_discord(reveled_board))
+
+        self.games[ctx.author] = {
+            "board": board,
+            "reveled": reveled_board,
```
tagptroll1: revealed* :^(

vivax3794:
there is a small, bug with the last commit. fixing it now

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+        await game["chat_msg"].edit(content=self.format_for_discord(game["reveled"]))
+
+    @commands.dm_only()
+    @commands.command(name="flag")
+    async def flag_command(self, ctx: commands.Context, *cords) -> None:
+        """Place multiple flags on the board"""
+        board = self.games[ctx.author]["reveled"]
+        for cord in cords:
+            x, y = self.get_cords(cord[0], cord[1])
+            if board[y][x] == "hidden":
```
tagptroll1: Maybe an Enum of the states for each cell?

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+                board[y][x] = "flag"
+
+        await self.reload_board(ctx)
+
+    async def lost(self, ctx: commands.Context) -> None:
+        """The player lost the game"""
+        game = self.games[ctx.author]
+        game["reveled"] = game["board"]
+        await self.reload_board(ctx)
+        await ctx.author.send(":fire: You lost :fire: ")
```
tagptroll1: Does it have to DM the user?

tagptroll1: Applies for most of the game, Unaware if this was discussed or not prior to make it

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,198 @@
...
+        await game["chat_msg"].channel.send(f":fire: {ctx.author.mention} just lost minesweeper :fire:")
+        del self.games[ctx.author]
+
+    async def won(self, ctx: commands.Context):
```
tagptroll1: You have a return value typing for every other function, why not this one?

bot/seasons/evergreen/minesweeper.py
tagptroll1: Bonus question, why not ID over member for key?  Just curious

bot/seasons/evergreen/minesweeper.py
tagptroll1: `sends it to the channel for us to play` is what the issue said at least

bot/seasons/evergreen/minesweeper.py
vivax3794: just tough it would be easier, would id be better?

bot/seasons/evergreen/minesweeper.py
vivax3794: guess I forgot

bot/seasons/evergreen/minesweeper.py
vivax3794: yes, it was decided that it should be played in dms.

bot/seasons/evergreen/minesweeper.py
tagptroll1: I'm unsure how a member is hashed, so it might change, ids dont

vivax3794: okay then

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,197 @@
...
+            "board": board,
+            "revealed": revealed_board,
+            "dm_msg": dm_msg,
+            "chat_msg": chat_msg
+        }
+
+    @staticmethod
+    def get_cords(value1: str, value2: str) -> typing.Tuple[int, int]:
+        """Take in 2 values for the cords and turn them into numbers"""
+        return ord(value1.lower()) - 97, int(value2) - 1
```
tagptroll1: This throws an error if the users uses <num><letter>.  I believe this should be caught and either try to flip them, or give the user a message that it was the wrong format and what is correct.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+        await self.reload_board(ctx)
+        await ctx.author.send(":tada:  You won! :tada: ")
+        await game["chat_msg"].channel.send(f":tada: {ctx.author.mention} just won minesweeper :tada:")
+        del self.games[ctx.author.id]
+
+    def reveal_zeros(self, revealed: GameBoard, board: GameBoard, x: int, y: int) -> None:
+        """Used when a 0 is encountered to do a flood fill"""
+        for x_ in [x - 1, x, x + 1]:
+            for y_ in [y - 1, y, y + 1]:
+                if x_ == -1 or x_ == 10 or y_ == -1 or y_ == 10 or revealed[y_][x_] != "hidden":
```
MarkKoz: This is similar to the code in `generate_board`. I think it'd be good to put this in a generic search function which yields the matching coordinates.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+        """The player won the game"""
+        game = self.games[ctx.author.id]
+        game["revealed"] = game["board"]
+        await self.reload_board(ctx)
+        await ctx.author.send(":tada:  You won! :tada: ")
+        await game["chat_msg"].channel.send(f":tada: {ctx.author.mention} just won minesweeper :tada:")
+        del self.games[ctx.author.id]
+
+    def reveal_zeros(self, revealed: GameBoard, board: GameBoard, x: int, y: int) -> None:
+        """Used when a 0 is encountered to do a flood fill"""
```
MarkKoz:
```suggestion
        """Recursively reveal adjacent cells when a 0 cell is encountered."""
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+    @commands.dm_only()
+    @commands.command(name="reveal")
+    async def reveal_command(self, ctx: commands.Context, *cords: CordConverter) -> None:
+        """Reveal multiple cells"""
+        game = self.games[ctx.author.id]
+        revealed: GameBoard = game["revealed"]
+        board: GameBoard = game["board"]
+
+        reload_board = True
+        for x, y in cords:
```
MarkKoz: This loop will keep going even if the player has won or lost.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+            if not await self.reveal_one(ctx, revealed, board, x, y):
+                reload_board = False
```
MarkKoz: What is this for? From what I can tell this only is set to `False` when  a player wins or loses. However, `won()` and `lost()` call `reload_board()` anyway so why not let this function handle revealing the board for all cases? The return values that accommodate this current system are unintuitive (e.g. `check_if_won` returning `False` when won) and difficult to follow.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class Minesweeper(commands.Cog):
+    """Play a game of minesweeper."""
+
+    def __init__(self, bot: commands.Bot) -> None:
+        self.games: DictOfGames = {}  # Store the currently running games
+
+    @staticmethod
+    def is_bomb(cell: typing.Union[str, int]) -> int:
```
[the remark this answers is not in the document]
vivax3794: it was left from an early version of the algorithm using map, forgot to remove it.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+            7: ":seven:",
+            8: ":eight:",
+            9: ":nine:",
+            10: ":keycap_ten:",
+            "bomb": ":bomb:",
+            "hidden": ":grey_question:",
+            "flag": ":triangular_flag_on_post:"
+        }
+
+        discord_msg = (
```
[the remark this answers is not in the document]
vivax3794: should it be named MESSAGE_MAPPING or just mapping?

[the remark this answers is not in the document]
MarkKoz: The former since it's a constant.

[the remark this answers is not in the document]
vivax3794: 👍

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+            4: ":four:",
+            5: ":five:",
+            6: ":six:",
+            7: ":seven:",
+            8: ":eight:",
+            9: ":nine:",
+            10: ":keycap_ten:",
+            "bomb": ":bomb:",
+            "hidden": ":grey_question:",
+            "flag": ":triangular_flag_on_post:"
```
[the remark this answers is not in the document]
vivax3794: only the last 3? would it not be better to just keep it as it is then?

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+                              "close the game with `.end`\n"
+                              "cords must be in format `<letter><number>`")
+        dm_msg = await ctx.author.send(self.format_for_discord(revealed_board))
+
+        self.games[ctx.author.id] = {
+            "board": board,
+            "revealed": revealed_board,
+            "dm_msg": dm_msg,
+            "chat_msg": chat_msg
+        }
```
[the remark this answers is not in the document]
vivax3794: idk, haven't used them before. would it be much to change?

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,212 @@
...
+        revealed[y][x] = board[y][x]
+        if board[y][x] == "bomb":
+            await self.lost(ctx)
+            return False
+        elif board[y][x] == 0:
+            self.reveal_zeros(revealed, board, x, y)
+        return await self.check_if_won(ctx, revealed, board)
+
+    @commands.dm_only()
+    @commands.command(name="reveal")
```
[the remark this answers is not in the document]
vivax3794: it works as it is. should I switch it?

bot/seasons/evergreen/minesweeper.py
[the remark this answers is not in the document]
MarkKoz: Hmm yeah, using an enum may make it more convoluted... You could just keep them in this dict but change the keys from string to enum instead of relying on enum values directly for the emoji conversions.

My thought was that using strings instead of enum constants is more prone to programmer errors. Enums work well with static type checkers so it'll be easier to spot when an incorrect value is used.

bot/seasons/evergreen/minesweeper.py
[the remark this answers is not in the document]
MarkKoz: If it works then no need to touch it I guess.

bot/seasons/evergreen/minesweeper.py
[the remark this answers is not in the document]
MarkKoz: It would mainly involve changing value access from `game["revealed"]` to `game.revealed` for example.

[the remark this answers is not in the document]
vivax3794: hm, it sounds better. okay I will do it

MarkKoz:
@vivax3794 You're resolving the suggestions too. Do you know about the GitHub feature that lets you batch and commit those suggestions directly?

vivax3794:
@MarkKoz no, I am just changing the code and resolving them as I go

MarkKoz:
> 
> 
> @MarkKoz no, I am just changing the code and resolving them as I go

I tried to make it easier for you by making the changes myself in some of the simple suggestions. The idea is that you can commit my work instead of re-doing it manually yourself.

vivax3794:
ah, okay. going to remember that.

vivax3794:
what files tab?
image

MarkKoz:
bild

vivax3794:
I see now, I though there were like a button `create batch` or somthing

bot/seasons/evergreen/minesweeper.py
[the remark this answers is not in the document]
vivax3794: I am not sure I completely understand what you mean? have an Enum class for the emojis and then use that as the values?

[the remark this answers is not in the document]
MarkKoz:
```py
class Cell(Enum):
    BOMB
    HIDDEN
    FLAG

{
    10: ":keycap_ten:",
    Cell.BOMB = ":bomb:"
}

board[y][x] = Cell.BOMB
```

Still not sure if it's a good idea.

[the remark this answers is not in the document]
vivax3794: hm, idk. I think it good with just strings. are there any performance improvements using that?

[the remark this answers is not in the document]
MarkKoz: Doubtful. I wasn't concerned with performance to begin with.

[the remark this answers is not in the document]
vivax3794: then, I think it's okay the way it is unless somebody wants to add that.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -25,188 +46,188 @@ class CordConverter(commands.Converter):
         return x, y
 
 
+GameDict = typing.List[typing.List[typing.Union[str, int]]]
```
MarkKoz: This was fine being called `GameBoard`. My previous review was for `DictOfGames`, not this.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -1,21 +1,42 @@
...
+    "flag": ":triangular_flag_on_post:"
+}
+
+log = logging.getLogger(__name__)
+
+
+class CoordinateConverter(commands.Converter):
+    """Converter for Coordinates."""
+
+    async def convert(self, ctx, Coordinate: str) -> typing.Tuple[int, int]:
```
MarkKoz: Argument names should be lowercase. Surprised the linter didn't catch this...

bot/seasons/evergreen/minesweeper.py
```diff
@@ -25,188 +46,188 @@ class CordConverter(commands.Converter):
...
+class Game:
+    """The data for a game."""
+
+    board: GameDict
+    revealed: GameDict
+    dm_msg: discord.message
+    chat_msg: discord.message
+
+
+DictOfGames = typing.Dict[int, Game]
```
MarkKoz: This is the one that I think should be called `GamesDict` (plural is better).

bot/seasons/evergreen/minesweeper.py
```diff
@@ -25,188 +46,188 @@ class CordConverter(commands.Converter):
...
-        game["dm_msg"] = await ctx.author.send(self.format_for_discord(game["revealed"]))
-        await game["chat_msg"].edit(content=self.format_for_discord(game["revealed"]))
+        await game.dm_msg.delete()
+        game.dm_msg = await ctx.author.send(self.format_for_discord(game.revealed))
+        await game.chat_msg.edit(content=self.format_for_discord(game.revealed))
 
     @commands.dm_only()
     @commands.command(name="flag")
-    async def flag_command(self, ctx: commands.Context, *cords: CordConverter) -> None:
+    async def flag_command(self, ctx: commands.Context, *Coordinates: CoordinateConverter) -> None:
```
MarkKoz: Argument name should be lowercase

MarkKoz:
@vivax3794 I think you messed up the merge in 67bb02b and overwrote the previous work in 1a05919

vivax3794:
oh, that's irritating.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+                board[y][x] = "flag"
+
+        await self.update_boards(ctx)
+
+    async def lost(self, ctx: commands.Context) -> None:
+        """The player lost the game"""
+        game = self.games[ctx.author.id]
+        game.revealed = game.board
+        await ctx.author.send(":fire: You lost! :fire: ")
+        await game.chat_msg.channel.send(f":fire: {ctx.author.mention} just lost minesweeper :fire:")
```
MarkKoz:
```suggestion
        await game.chat_msg.channel.send(f":fire: {ctx.author.mention} just lost Minesweeper! :fire:")
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+            if board[y][x] == "hidden":
+                board[y][x] = "flag"
+
+        await self.update_boards(ctx)
+
+    async def lost(self, ctx: commands.Context) -> None:
+        """The player lost the game"""
+        game = self.games[ctx.author.id]
+        game.revealed = game.board
+        await ctx.author.send(":fire: You lost! :fire: ")
```
MarkKoz:
```suggestion
        await ctx.author.send(":fire: You lost! :fire:")
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+        await game.chat_msg.channel.send(f":fire: {ctx.author.mention} just lost minesweeper :fire:")
+
+    async def won(self, ctx: commands.Context) -> None:
+        """The player won the game"""
+        game = self.games[ctx.author.id]
+        game.revealed = game.board
+        await ctx.author.send(":tada:  You won! :tada: ")
```
MarkKoz:
```suggestion
        await ctx.author.send(":tada: You won! :tada:")
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+        await game.chat_msg.channel.send(f":tada: {ctx.author.mention} just won minesweeper :tada:")
```
MarkKoz:
```suggestion
        await game.chat_msg.channel.send(f":tada: {ctx.author.mention} just won Minesweeper! :tada:")
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -25,184 +46,188 @@ class CordConverter(commands.Converter):
...
+            return True
 
-    async def reveal_one(self, ctx: commands.Context, revealed: GameBoard, board: GameBoard, x: int, y: int) -> bool:
+    async def reveal_one(self, ctx: commands.Context, revealed: GameDict, board: GameDict, x: int, y: int) -> bool:
         """Reveal one square."""
         revealed[y][x] = board[y][x]
         if board[y][x] == "bomb":
             await self.lost(ctx)
-            return False
+            return True  # game ended
```
MarkKoz: It would be clearer if the return value was described in the docstring.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+        await self.update_boards(ctx)
+        await ctx.author.send(":no_entry: you canceled the game :no_entry:")
+        await game.chat_msg.channel.send(f"{ctx.author.mention} just canceled minesweeper")
+        del self.games[ctx.author.id]
+
+
+def setup(bot: commands.Bot) -> None:
+    """Load the Minesweeper cog."""
+    bot.add_cog(Minesweeper(bot))
+    log.info("minesweeper cog loaded")
```
MarkKoz:
```suggestion
    log.info("Minesweeper cog loaded")
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+        else:
+            await self.update_boards(ctx)
+
+    @commands.command(name="end")
+    async def end_command(self, ctx: commands.Context):
+        """End the current game"""
+        game = self.games[ctx.author.id]
+        game.revealed = game.board
+        await self.update_boards(ctx)
+        await ctx.author.send(":no_entry: you canceled the game :no_entry:")
```
MarkKoz:
```suggestion
        await ctx.author.send(":no_entry: You canceled the game. :no_entry:")
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+    async def reveal_one(self, ctx: commands.Context, revealed: GameDict, board: GameDict, x: int, y: int) -> bool:
+        """Reveal one square."""
+        revealed[y][x] = board[y][x]
+        if board[y][x] == "bomb":
+            await self.lost(ctx)
+            return True  # game ended
+        elif board[y][x] == 0:
+            self.reveal_zeros(revealed, board, x, y)
+        won = await self.check_if_won(ctx, revealed, board)
+        return won
```
MarkKoz: Why was the return split from the function call? It was fine before.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+                continue
+            revealed[y_][x_] = board[y_][x_]
+            if board[y_][x_] == 0:
+                self.reveal_zeros(revealed, board, x_, y_)
+
+    async def check_if_won(self, ctx, revealed: GameDict, board: GameDict) -> bool:
+        """Checks if a player has won"""
+        for x in range(10):
+            for y in range(10):
+                if revealed[y][x] == "hidden" and board[y][x] != "bomb":
```
MarkKoz: The second check seemed redundant since it would always be true if the first one is true.

```suggestion
                if revealed[y][x] == "hidden":
```

At this point you could consider using `any()` but the for loops are fine too.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+            await self.update_boards(ctx)
+
+    @commands.command(name="end")
+    async def end_command(self, ctx: commands.Context):
+        """End the current game"""
+        game = self.games[ctx.author.id]
+        game.revealed = game.board
+        await self.update_boards(ctx)
+        await ctx.author.send(":no_entry: you canceled the game :no_entry:")
+        await game.chat_msg.channel.send(f"{ctx.author.mention} just canceled minesweeper")
```
MarkKoz: Seems you missed this review the first time:

Perhaps the message with the board could be edited to include this rather than sending a new message? Unlike a win/loss, most people won't care if a game has been cancelled so this message seems a bit spammy.

MarkKoz:
```suggestion
        await game.chat_msg.channel.send(f"{ctx.author.mention} just canceled Minesweeper.")
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+        board: GameDict = self.generate_board(bomb_chance)
+        revealed_board: GameDict = [["hidden"] * 10 for _ in range(10)]
+
+        await ctx.send(f"{ctx.author.mention} is playing Minesweeper")
+        chat_msg = await ctx.send(self.format_for_discord(revealed_board))
+
+        await ctx.author.send(
+            f"Play by typing: `{Client.prefix}reveal xy [xy]` or `{Client.prefix}flag xy [xy]` \n"
+            "Close the game with `.end`\n"
+            "Coordinates must be in format `<letter><number>`"
```
MarkKoz:
```suggestion
            "Coordinates must be in the format `<letter><number>`"
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+        # Add game to list
+        board: GameDict = self.generate_board(bomb_chance)
+        revealed_board: GameDict = [["hidden"] * 10 for _ in range(10)]
+
+        await ctx.send(f"{ctx.author.mention} is playing Minesweeper")
+        chat_msg = await ctx.send(self.format_for_discord(revealed_board))
+
+        await ctx.author.send(
+            f"Play by typing: `{Client.prefix}reveal xy [xy]` or `{Client.prefix}flag xy [xy]` \n"
+            "Close the game with `.end`\n"
```
MarkKoz:
```suggestion
            f"Close the game with `{Client.prefix}end`\n"
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,233 @@
...
+            raise commands.ArgumentParsingError()
+        value1 = Coordinate[0]
+        value2 = Coordinate[1:]
+        if not value2.isdigit():
+            raise commands.ArgumentParsingError()
+        x = ord(value1) - 97
+        y = int(value2) - 1
+        if (not 0 <= x <= 9) or (not 0 <= y <= 9):
+            raise commands.ArgumentParsingError()
+        return x, y
```
MarkKoz: This function is dense. Can you put spaces between chunks of code where it makes sense so this is easier to read?

bot/seasons/evergreen/minesweeper.py
vivax3794: wait this would make it so you are required to flag all the spots to, I think I like that.

MarkKoz: That was unintended. I was actually incorrect because I didn't notice they're different boards (`revealed ` vs `board`)

MarkKoz: If you like this idea then maybe make a "hard mode" :) ?

vivax3794: somebody mentioned that you actually need to mark with flags to win, but I don't know if that's correct.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -222,8 +229,9 @@ def reveal_zeros(self, revealed: GameDict, board: GameDict, x: int, y: int) -> N
         game = self.games[ctx.author.id]
         game.revealed = game.board
         await self.update_boards(ctx)
-        await ctx.author.send(":no_entry: You canceled the game. :no_entry:")
-        await game.chat_msg.channel.send(f"{ctx.author.mention} just canceled Minesweeper.")
+        new_msg = f":no_entry: Game canceled :no_entry:\n{self.format_for_discord(game.revealed)}"
```
MarkKoz: Can you use the contents of the previous message instead of calling the format function again?

vivax3794: probably.

bot/seasons/evergreen/minesweeper.py
tagptroll1: If i remember correctly you win both if you have all bombs flagged, or all non bombs opened.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+    def get_neighbours(x: int, y: int) -> typing.Generator:
+        """Get all the neighbouring x and y including it self."""
+        for x_ in [x - 1, x, x + 1]:
+            for y_ in [y - 1, y, y + 1]:
+                if x_ != -1 and x_ != 10 and y_ != -1 and y_ != 10:
+                    yield x_, y_
+
+    def generate_board(self, bomb_chance: float) -> GameBoard:
+        """Generate a 2d array for the board."""
+        board: GameBoard = [["bomb" if random() <= bomb_chance else "number" for _ in range(10)] for _ in range(10)]
```
tagptroll1: This line is really long, breaking it into multiple lines will help with readability.  

Probably not the best solution, but a suggestion.
```suggestion
        board: GameBoard = [
            [
                "bomb" if random() <= bomb_chance 
                else "number" 
                for _ in range(10)
            ] for _ in range(10)
        ]
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+    def format_for_discord(board: GameBoard) -> str:
+        """Format the board as a string for Discord."""
+        discord_msg = (
+            ":stop_button:    :regional_indicator_a::regional_indicator_b::regional_indicator_c:"
+            ":regional_indicator_d::regional_indicator_e::regional_indicator_f::regional_indicator_g:"
+            ":regional_indicator_h::regional_indicator_i::regional_indicator_j:\n\n"
+        )
+        rows: typing.List[str] = []
+        for row_number, row in enumerate(board):
+            new_row = MESSAGE_MAPPING[row_number + 1] + "    "
```
tagptroll1:
```suggestion
            new_row = f"{MESSAGE_MAPPING[row_number + 1]}    "
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+            for cell in row:
```
tagptroll1: Make sure this behaves the way I think it i will

```suggestion
            new_row = "".join(MESSAGE_MAPPING[cell] for cell in row)
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+                new_row += MESSAGE_MAPPING[cell]
```
tagptroll1: Following the suggestion before.
```suggestion

```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+        """Checks if a player has won"""
+        for x in range(10):
+            for y in range(10):
+                if revealed[y][x] == "hidden":
+                    return False
+        else:
+            await self.won(ctx)
+            return True
+
+    async def reveal_one(self, ctx: commands.Context, revealed: GameBoard, board: GameBoard, x: int, y: int) -> bool:
```
tagptroll1: This function  definition is really long, I suggest breaking it up into multiple lines for readability.

```suggestion
    async def reveal_one(
        self,
        ctx: commands.Context,
        revealed: GameBoard,
        board: GameBoard,
        x: int,
        y: int) -> bool:
```
Or alternatively:
```suggestion
    async def reveal_one(
        self,
        ctx: commands.Context,
        revealed: GameBoard,
        board: GameBoard,
        x: int,
        y: int
    ) -> bool:
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        Reveal one square.
+
+        return is True if the game ended, breaking the loop in `reveal_command` and deleting the game
+        """
+        revealed[y][x] = board[y][x]
+        if board[y][x] == "bomb":
+            await self.lost(ctx)
+            return True  # game ended
```
tagptroll1: Docstring explains this now.

```suggestion
            return True
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+    @commands.dm_only()
+    @commands.command(name="reveal")
+    async def reveal_command(self, ctx: commands.Context, *coordinates: CoordinateConverter) -> None:
+        """Reveal multiple cells"""
+        game = self.games[ctx.author.id]
+        revealed: GameBoard = game.revealed
+        board: GameBoard = game.board
+
+        for x, y in coordinates:
+            if await self.reveal_one(ctx, revealed, board, x, y):  # game ended
```
tagptroll1:
```suggestion
            # reveal_one returns True if the revealed cell is a bomb, ending the game
            if await self.reveal_one(ctx, revealed, board, x, y):
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+                await self.update_boards(ctx)
+                del self.games[ctx.author.id]
+                break
+        else:
+            await self.update_boards(ctx)
+
+    @commands.command(name="end")
+    async def end_command(self, ctx: commands.Context):
+        """End the current game"""
```
tagptroll1: Believe it might be a bit misleading saying it ends the current game.  
Whos game? All games? Which game is current? etc.
```suggestion
        """End your current game"""
```

bot/seasons/evergreen/minesweeper.py
vivax3794: I think I will go with that you need to flag or reveal the spaces.

vivax3794: on Wikipedia: `The game is won when all mine-free squares are revealed` so I think I will go with that.

vivax3794:
@MarkKoz about point 2, should they only be able to start the game from the server maybe?

MarkKoz:
That would be a simple fix but if you're up for it it'd be nice to still support it.

vivax3794:
hmm, okay. would need to add some logic where I am sending message to the channel the command was started in. maybe add an `in_dm` bool to the game.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+class Minesweeper(commands.Cog):
+    """Play a game of Minesweeper."""
+
+    def __init__(self, bot: commands.Bot) -> None:
+        self.games: GamesDict = {}  # Store the currently running games
+
+    @staticmethod
+    def get_neighbours(x: int, y: int) -> typing.Generator:
```
avayert: Needs to be `Generator[YieldType, SendType, ReturnType]`.

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @staticmethod
+    def format_for_discord(board: GameBoard) -> str:
+        """Format the board as a string for Discord."""
+        discord_msg = (
+            ":stop_button:    :regional_indicator_a::regional_indicator_b::regional_indicator_c:"
+            ":regional_indicator_d::regional_indicator_e::regional_indicator_f::regional_indicator_g:"
+            ":regional_indicator_h::regional_indicator_i::regional_indicator_j:\n\n"
+        )
+        rows: typing.List[str] = []
```
avayert: seems like an overzealous use of type hints to me, you can clearly see strings being inserted into it just two lines below

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+            rows.append(new_row)
+
+        discord_msg += "\n".join(rows)
+        return discord_msg
+
+    @commands.command(name="minesweeper")
+    async def minesweeper_command(self, ctx: commands.Context, bomb_chance: float = .2) -> None:
+        """Start a game of Minesweeper."""
+        if ctx.author.id in self.games:  # Player is already playing
+            msg = await ctx.send(f"{ctx.author.mention} you already have a game running!")
```
avayert:
```suggestion
            msg = await ctx.send(f"{ctx.author.mention} you already have a game running!", delete_after=2)
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,241 @@
...
+        for x_, y_ in self.get_neighbours(x, y):
+            if revealed[y_][x_] != "hidden":
+                continue
+            revealed[y_][x_] = board[y_][x_]
+            if board[y_][x_] == 0:
+                self.reveal_zeros(revealed, board, x_, y_)
+
+    async def check_if_won(self, ctx, revealed: GameBoard, board: GameBoard) -> bool:
+        """Checks if a player has won"""
+        for x in range(10):
```
avayert: `range` based iteration here seems unnecessary iterate over all of the rows and use `any(item == 'hidden' for item in row)`

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,266 @@
...
+    @commands.dm_only()
+    @minesweeper_group.command(name="reveal")
+    async def reveal_command(self, ctx: commands.Context, *coordinates: CoordinateConverter) -> None:
+        """Reveal multiple cells"""
+        game = self.games[ctx.author.id]
+        revealed: GameBoard = game.revealed
+        board: GameBoard = game.board
+
+        for x, y in coordinates:
+            # reveal_one returns True if the revealed cell is a bomb or the player won., ending the game
```
tagptroll1: Did i do that? whops :)
```suggestion
            # reveal_one returns True if the revealed cell is a bomb or the player won, ending the game
```

vivax3794: You forgot to mention that's it returns true when they win to, I guess I am just used to putting `.`

sco1:
I'm not understanding where or how I'm supposed to be playing the game.

In the channel?
image


In the DM?
image

vivax3794:
In DMS `.minesweeper reveal A1`

sco1:
image

That doesn't work

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,266 @@
...
+    """Converter for Coordinates."""
+
+    async def convert(self, ctx, coordinate: str) -> typing.Tuple[int, int]:
+        """Take in a coordinate string and turn it into x, y"""
+        if not 2 <= len(coordinate) <= 3:
+            raise commands.BadArgument('Invalid co-ordinate provided')
+
+        digit, letter = sorted(coordinate.lower())
+
+        if not letter.isdigit():
```
MarkKoz: This is probably the fix

```suggestion
        if not digit.isdigit():
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,266 @@
...
+        revealed_board: GameBoard = [["hidden"] * 10 for _ in range(10)]
+
+        if ctx.guild:
+            await ctx.send(f"{ctx.author.mention} is playing Minesweeper")
+            chat_msg = await ctx.send(self.format_for_discord(revealed_board))
+        else:
+            chat_msg = None
+
+        await ctx.author.send(
+            f"Play by typing: `{Client.prefix}reveal xy [xy]` or `{Client.prefix}flag xy [xy]` \n"
```
MarkKoz:
```suggestion
            f"Play by typing: `{Client.prefix}ms reveal xy [xy]` or `{Client.prefix}ms flag xy [xy]` \n"
```

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,266 @@
...
+            f"Close the game with `{Client.prefix}end`\n"
```
MarkKoz:
```suggestion
            f"Close the game with `{Client.prefix}ms end`\n"
```

vivax3794:
Fuck, I messed up the condition. Why did I do this late.

bot/seasons/evergreen/minesweeper.py
vivax3794: Yes, I think so. Thanks for that. On my phone so could not fix it myself

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,266 @@
...
+            await ctx.message.delete(delay=2)
+            return
+
+        # Add game to list
+        board: GameBoard = self.generate_board(bomb_chance)
+        revealed_board: GameBoard = [["hidden"] * 10 for _ in range(10)]
+
+        if ctx.guild:
+            await ctx.send(f"{ctx.author.mention} is playing Minesweeper")
+            chat_msg = await ctx.send(self.format_for_discord(revealed_board))
```
sco1: We'll need some kind of header line for this so the board gets rendered properly:

image


Something like `"Here's your board!\n"` should be fine (leading whitespace is stripped by Discord)

bot/seasons/evergreen/minesweeper.py
```diff
@@ -0,0 +1,266 @@
...
+        else:
+            chat_msg = None
+
+        await ctx.author.send(
+            f"Play by typing: `{Client.prefix}ms reveal xy [xy]` or `{Client.prefix}ms flag xy [xy]` \n"
+            f"Close the game with `{Client.prefix}ms end`\n"
+        )
+        dm_msg = await ctx.author.send(self.format_for_discord(revealed_board))
```
sco1: Same as the above.

bot/seasons/evergreen/minesweeper.py
vivax3794: Never noticed that, as its only like that on compact mode. I can do it tomorrow as its getting late.

vivax3794:
so somebody suggested this code for the coords: ```
digit, letter = sorted(coordinate.lower())```
 well if it lets say is h10 it does not work........

PR #258, opened 2019-08-15
Add RecommendGame cog and resources

jakeHebert:
Created the RecommendGame cog with `.recommend_game` command as well as a new directory, `bot/resources/evergreen/game_recs`, for game recommendation files in .json to be stored and used by the cog. I added 4 different game recommendations to start with.
Current Embed formatting is as such:
gameRecTest

### DIFF
--- a/bot/resources/evergreen/game_recs/chrono_trigger.json
+++ b/bot/resources/evergreen/game_recs/chrono_trigger.json
@@ -1,7 +1,7 @@
 {
   "title": "Chrono Trigger",
-  "recText": "One of the best games of all time. A brilliant story involving time-travel with loveable characters. It has a brilliant score by Yasonuri Mitsuda and artwork by Akira Toriyama. With over 20 endings and New Game+, there is a huge amount of replay value here.",
-  "wikiLink": "https://rawg.io/games/chrono-trigger-1995",
+  "description": "One of the best games of all time. A brilliant story involving time-travel with loveable characters. It has a brilliant score by Yasonuri Mitsuda and artwork by Akira Toriyama. With over 20 endings and New Game+, there is a huge amount of replay value here.",
+  "link": "https://rawg.io/games/chrono-trigger-1995",
   "image": "https://vignette.wikia.nocookie.net/chrono/images/2/24/Chrono_Trigger_cover.jpg",
   "author": "352635617709916161"
 }
--- a/bot/resources/evergreen/game_recs/chrono_trigger.json
+++ b/bot/resources/evergreen/game_recs/chrono_trigger.json
@@ -1,0 +1,10 @@
+{
+  "title": "Chrono Trigger",
+  "recText": "My personal favorite game of all time. Square + Enix + Akira Toriyama = Pure Time-traveling gold.",
+  "console": [
+    "snes",
+    "gba",
+    "pc"
+  ],
+  "image": "https://vignette.wikia.nocookie.net/chrono/images/2/24/Chrono_Trigger_cover.jpg"
+}
--- a/bot/resources/evergreen/game_recs/chrono_trigger.json
+++ b/bot/resources/evergreen/game_recs/chrono_trigger.json
@@ -1,10 +1,12 @@
 {
   "title": "Chrono Trigger",
-  "recText": "My personal favorite game of all time. Square + Enix + Akira Toriyama = Pure Time-traveling gold.",
+  "recText": "One of the best games of all time. A brilliant story involving time-travel with loveable characters. It has a brilliant score by Yasonuri Mitsuda and artwork by Akira Toriyama. With over 20 endings and New Game+, there is a huge amount of replay value here.",
   "console": [
     "snes",
     "gba",
     "pc"
   ],
-  "image": "https://vignette.wikia.nocookie.net/chrono/images/2/24/Chrono_Trigger_cover.jpg"
+  "wikiLink": "https://chrono.fandom.com/wiki/Chrono_Trigger",
+  "image": "https://vignette.wikia.nocookie.net/chrono/images/2/24/Chrono_Trigger_cover.jpg",
+  "author": "352635617709916161"
 }
--- a/bot/resources/evergreen/game_recs/chrono_trigger.json
+++ b/bot/resources/evergreen/game_recs/chrono_trigger.json
@@ -1,12 +1,7 @@
 {
   "title": "Chrono Trigger",
   "recText": "One of the best games of all time. A brilliant story involving time-travel with loveable characters. It has a brilliant score by Yasonuri Mitsuda and artwork by Akira Toriyama. With over 20 endings and New Game+, there is a huge amount of replay value here.",
-  "console": [
-    "snes",
-    "gba",
-    "pc"
-  ],
-  "wikiLink": "https://chrono.fandom.com/wiki/Chrono_Trigger",
+  "wikiLink": "https://rawg.io/games/chrono-trigger-1995",
   "image": "https://vignette.wikia.nocookie.net/chrono/images/2/24/Chrono_Trigger_cover.jpg",
   "author": "352635617709916161"
 }
--- a/bot/seasons/evergreen/recommend_game.py
+++ b/bot/seasons/evergreen/recommend_game.py
@@ -1,0 +1,56 @@
+import json
+import logging
+import os
+from pathlib import Path
+from random import shuffle
+
+import discord
+from discord.ext import commands
+
+log = logging.getLogger(__name__)
+DIR = "bot/resources/evergreen/game_recs"
+game_recs = []
+
+
+class RecommendGame(commands.Cog):
+    """Commands related to recommending games."""
+
+    def __init__(self, bot):
+        self.bot = bot
+        self.populate_recs()
+
+    @commands.command(name="recommend_game")
+    async def recommend_game(self, ctx):
+        """Sends an Embed of a random game recommendation."""
+        if not game_recs:
+            self.populate_recs()
+            game = game_recs.pop()
+        else:
+            game = game_recs.pop()
+
+        embed = discord.Embed(
+            title=game['title'],
+            url=game['wikiLink'],
+            color=discord.Colour.blue()
+        )
+
+        author = self.bot.get_user(int(game['author']))
+        embed.set_author(name=author.name, icon_url=author.avatar_url)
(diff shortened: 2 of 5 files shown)


bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self, bot):
+        self.bot = bot
+        self.populate_recs()
+
+    @commands.command(name="recommend_game")
+    async def recommend_game(self, ctx):
+        """Sends an Embed of a random game recommendation."""
+        if not game_recs:
+            self.populate_recs()
```
MarkKoz: It's inefficient to have to re-read the JSON files. Instead of popping items from the list, you can keep track of an index. Then, the list would be shuffled once the index reached the end.

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,52 @@
...
+        embed = discord.Embed(color=discord.Colour.blue())
+        embed.set_author(name=author.name, icon_url=author.avatar_url)
+        embed.set_image(url=game['image'])
+        embed.add_field(name='Recommendation: ' + game['title'] + '\n' + game['wikiLink'], value=game['recText'])
+
+        await ctx.send(embed=embed)
+
+    def populate_recs(self):
+        """Populates the list `game_recs` from resources."""
+        for file_url in os.listdir(DIR):
```
MarkKoz: It'd be better to stick with pathlib here. Use `DIR.iterdir()` or, probably even better, do a glob to ensure only JSON files are read: `DIR.glob("*.json")`. `DIR` would of course have to be a `Path` object instead of a string in both cases.

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Creating and formatting Embed
+        embed = discord.Embed(color=discord.Colour.blue())
+        embed.set_author(name=author.name, icon_url=author.avatar_url)
+        embed.set_image(url=game['image'])
+        embed.add_field(name='Recommendation: ' + game['title'] + '\n' + game['wikiLink'], value=game['recText'])
+
+        await ctx.send(embed=embed)
+
+    def populate_recs(self):
```
MarkKoz: Going along with my other suggestion, the files will only need to be read once so this code could be moved to the top of the module.

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,52 @@
...
+            with Path(DIR, file_url).open(encoding='utf-8') as file:
+                data = json.load(file)
+            game_recs.append(data)
+        shuffle(game_recs)
+
+
+def setup(bot):
+    """Loads the RecommendGame cog."""
+    bot.add_cog(RecommendGame(bot))
+    log.info("Recommend_Game cog loaded")
```
MarkKoz:
```suggestion
    log.info("RecommendGame cog loaded")
```

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class RecommendGame(commands.Cog):
+    """Commands related to recommending games."""
+
+    def __init__(self, bot):
+        self.bot = bot
+        self.populate_recs()
+
+    @commands.command(name="recommend_game")
```
MarkKoz: Any ideas for a clear name that doesn't have an underscore?

jakeHebert: gameRec maybe?

bot/seasons/evergreen/recommend_game.py
jakeHebert: Not sure why I ignored using an index, it makes so much more sense.

bot/seasons/evergreen/recommend_game.py
jakeHebert: I'll add `gameRec` and `recommendGame` as aliases for now.

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,54 @@
...
+import discord
+from discord.ext import commands
+
+log = logging.getLogger(__name__)
+DIR = Path("bot/resources/evergreen/game_recs")
+game_recs = []
+
+# Populate the list `game_recs` with resource files
+for file_url in DIR.glob("*.json"):
+    with Path(file_url).open(encoding='utf-8') as file:
```
MarkKoz: `Path.glob()` will yield `Path` objects.

```suggestion
    with rec_path.open(encoding='utf-8') as file:
```

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,54 @@
...
+
+import discord
+from discord.ext import commands
+
+log = logging.getLogger(__name__)
+DIR = Path("bot/resources/evergreen/game_recs")
+game_recs = []
+
+# Populate the list `game_recs` with resource files
+for file_url in DIR.glob("*.json"):
```
MarkKoz: Naming a local file path a "URL" is strange.

```suggestion
for rec_path in DIR.glob("*.json"):
```

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,54 @@
...
+    async def recommend_game(self, ctx):
+        """Sends an Embed of a random game recommendation."""
+        if self.index < len(game_recs):
+            game = game_recs[self.index]
+            self.index += 1
+        else:
+            self.index = 0
+            shuffle(game_recs)
+            game = game_recs[self.index]
+            self.index += 1
```
MarkKoz: If you invert the condition then you can remove the repetition of the last two lines (an `else` wouldn't be needed any more).

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class RecommendGame(commands.Cog):
+    """Commands related to recommending games."""
+
+    def __init__(self, bot):
+        self.bot = bot
+        self.index = 0
+
+    @commands.command(name="recommend_game", aliases=['gameRec', 'recommendGame'])
```
MarkKoz: From what I've seen, none of the other commands use an underscore or camel case for multiple words. Anyone please correct me if I am wrong or you otherwise disagree.

```suggestion
    @commands.command(name="recommendgame", aliases=['gamerec'])
```

bot/resources/evergreen/game_recs/chrono_trigger.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "Chrono Trigger",
+  "recText": "One of the best games of all time. A brilliant story involving time-travel with loveable characters. It has a brilliant score by Yasonuri Mitsuda and artwork by Akira Toriyama. With over 20 endings and New Game+, there is a huge amount of replay value here.",
```
MarkKoz: A better name for these fields would be `description`.

bot/resources/evergreen/game_recs/chrono_trigger.json
```diff
@@ -0,0 +1,7 @@
...
+  "wikiLink": "https://rawg.io/games/chrono-trigger-1995",
```
MarkKoz: Just `link` would suffice for this field. rawg.io isn't really a wiki anyway. Also, let's avoid camel case.

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,54 @@
+import json
+import logging
+from pathlib import Path
+from random import shuffle
+
+import discord
+from discord.ext import commands
+
+log = logging.getLogger(__name__)
+DIR = Path("bot/resources/evergreen/game_recs")
```
MarkKoz: Can just be inlined at this point instead of having it be a constant.

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,50 @@
...
+            self.index = 0
+            shuffle(game_recs)
+        game = game_recs[self.index]
+        self.index += 1
+
+        author = self.bot.get_user(int(game['author']))
+
+        # Creating and formatting Embed
+        embed = discord.Embed(color=discord.Colour.blue())
+        embed.set_author(name=author.name, icon_url=author.avatar_url)
```
MarkKoz: This will fail if the user isn't found (e.g. they left the server). `get_user` will return `None` in such case.

jakeHebert: Cool, ill fix this when i get home

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,51 @@
...
+    with rec_path.open(encoding='utf-8') as file:
+        data = json.load(file)
+    game_recs.append(data)
+shuffle(game_recs)
+
+
+class RecommendGame(commands.Cog):
+    """Commands related to recommending games."""
+
+    def __init__(self, bot):
```
kosayoda: Missing function annotations for the parameter and return value.
```suggestion
    def __init__(self, bot: commands.Bot) -> None:
```

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,51 @@
...
+        self.bot = bot
+        self.index = 0
+
+    @commands.command(name="recommendgame", aliases=['gamerec'])
+    async def recommend_game(self, ctx):
```
kosayoda: Similarly, missing function annotations for the parameter and return value.
```suggestion
    async def recommend_game(self, ctx: commands.Context) -> None:
```

bot/seasons/evergreen/recommend_game.py
```diff
@@ -0,0 +1,51 @@
...
+        embed = discord.Embed(color=discord.Colour.blue())
+        if author is not None:
+            embed.set_author(name=author.name, icon_url=author.avatar_url)
+        embed.set_image(url=game['image'])
+        embed.add_field(name='Recommendation: ' + game['title'] + '\n' + game['link'], value=game['description'])
+
+        await ctx.send(embed=embed)
+
+
+def setup(bot):
```
kosayoda: More missing annotations.
```suggestion
def setup(bot: commands.Bot) -> None:
```

PR #260, opened 2019-08-18
Modify text :-)

kosayoda:
This PR implements fun commands to modify given text.
A utility function was implemented to replace multiple words in a string.

### DIFF
--- a/bot/seasons/evergreen/fun.py
+++ b/bot/seasons/evergreen/fun.py
@@ -44,23 +44,26 @@ class Fun(Cog):
     @commands.command(name="uwu", aliases=("uwuwize", "uwuify",))
     async def uwu_command(self, ctx: Context, *, text: str) -> None:
         """Converts a given `text` into it's uwu equivalent."""
-        text = await Fun.text_to_message(ctx, text)
+        text = await Fun.get_discord_message(ctx, text)
         converted = utils.replace_many(text, UWU_WORDS, ignore_case=True, match_case=True)
         await ctx.send(f">>> {converted}")
 
     @commands.command(name="randomcase", aliases=("rcase", "randomcaps", "rcaps",))
     async def randomcase_command(self, ctx: Context, *, text: str) -> None:
         """Randomly converts the casing of a given `text`."""
-        text = await Fun.text_to_message(ctx, text)
+        text = await Fun.get_discord_message(ctx, text)
         converted = (
             char.upper() if round(random.random()) else char.lower() for char in text
         )
         await ctx.send(f">>> {''.join(converted)}")
 
     @staticmethod
-    async def text_to_message(ctx: Context, text: str) -> str:
+    async def get_discord_message(ctx: Context, text: str) -> str:
         """
-        Attempts to convert a given `text` to a discord Message, then return the contents.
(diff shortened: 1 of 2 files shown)


kosayoda:
Missed a typehint :man_facepalming:, will batch push with review changes.

bot/seasons/evergreen/fun.py
```diff
@@ -27,6 +40,48 @@ def __init__(self, bot):
...
+
+    @uwu_group.command(name="convert")
+    async def convert_uwu(self, ctx: Context, *, text: str) -> None:
+        """Converts a given `text` into it's uwu equivalent."""
+        text = await Fun.text_to_message(ctx, text)
+        converted = utils.replace_many(text, UWU_WORDS, ignore_case=True, match_case=True)
+        await ctx.send(f">>> {converted}")
+
+    @commands.group(name="randomcase", aliases=("rcase", "randomcaps", "rcaps",), invoke_without_command=True)
+    async def randomcase_group(self, ctx: Context, *, text: str) -> None:
```
MarkKoz: I don't see the point in having a group that only has 1 subcommand.

bot/seasons/evergreen/fun.py
```diff
@@ -27,6 +40,48 @@ def __init__(self, bot):
             output += getattr(Emojis, terning, '')
         await ctx.send(output)
 
+    @commands.group(name="uwu", aliases=("uwuwize", "uwuify",), invoke_without_command=True)
+    async def uwu_group(self, ctx: Context, *, text: str) -> None:
```
MarkKoz: I don't see the point in having a group that only has 1 subcommand.

bot/seasons/evergreen/fun.py
```diff
@@ -27,6 +40,48 @@ def __init__(self, bot):
...
+
+    @staticmethod
+    async def text_to_message(ctx: Context, text: str) -> str:
+        """
+        Attempts to convert a given `text` to a discord Message, then return the contents.
+
+        Returns `text` if the conversion fails.
+        """
+        try:
+            message = await MessageConverter().convert(ctx, text)
```
MarkKoz: Why is this conversion needed? How would the contents after the conversion differ from the input and why do we care?

MarkKoz:
Regarding the uwu text, it's a simple approach but it works. However, it got me interested in coming up with a more thorough approach to it, like [mapping graphemes to phonemes](https://github.com/dmort27/epitran#using-the-epitranvector-module) and determining replacements for the graphemes based on the phoneme they map to. This would allow for more thorough "uwuisation" of words while maintaining accuracy since it'd rely on the underlying phonemes rather than on the deep orthography of English and its irregular spelling.

bot/seasons/evergreen/fun.py
sco1: What was the answer to this question?

sco1: Never mind, reading the docstring would help 💯

kosayoda: Whoops, answered it elsewhere. It's so the commands work with a valid Discord message ID or link as an argument, if the conversion suceeds the text from the linked message gets converted.

Issue #248, opened 2019-08-04
Add Linting for Function Definition Type Hinting


sco1:
This will investigate linting our function definitions for type annotations using tools like [flake8-mypy](https://github.com/ambv/flake8-mypy).

See: https://github.com/python-discord/organisation/issues/106 for additional discussion.

PR #269, opened 2019-09-04
Hacktoberfest update & SaL Fix

sco1:
This PR addresses some issues in 2 cogs:

**Hacktoberfest**
Some minor cleanup to get the cog ready for the upcoming event:
* Fix command typo in account linking helper message
* Remove hardcoded values for event year & number of PRs before one gets a shirt, adjust related functionality
* Some type hinting kaizen in advance of #248

**Snakes**
The Snakes and Ladders command has been updated:
* The game can now be played solo
* Add missing `await` to the reaction timeout while waiting for the game to start, preventing the game from being cancelled after the timeout is reached & freeing the channel for another game
* Update cancellation logic as follows (per #253):
  * Game has timed out
  * There are no players left in the game
  * Non-player Moderation staff can cancel the game at any time
  * After a game has started, the initiator will no longer cancel the game if they leave; they can still cancel the game before it has started
* Fix issue with the game continuing after all players have left
  * This was caused by a bug in the round's logic flow. Because the game surrender logic is contained in the reaction block for the round, the round completion coroutine is still invoked after the game is surrendered so a new round is posted to the channel after the game object is deleted.

The alias for the Monster Survey cog has also been adjusted to deconflict with the alias for the Minesweeper cog.

This also adds a fix for our improperly configured flake8-docstring linter (see: https://github.com/python-discord/organisation/issues/131)

Closes: #252

### DIFF
--- a/bot/seasons/evergreen/snakes/utils.py
+++ b/bot/seasons/evergreen/snakes/utils.py
@@ -436,7 +436,7 @@ class SnakeAndLaddersGame:
                 if reaction.emoji == JOIN_EMOJI:
                     await self.player_join(user)
                 elif reaction.emoji == CANCEL_EMOJI:
-                    if user == self.author or not self._mod_in_game_check(user):
+                    if user == self.author or all((self._is_moderator(user), user not in self.players)):
                         # Allow game author or non-playing moderation staff to cancel a waiting game
                         await self.cancel_game()
                         return
@@ -705,10 +705,6 @@ class SnakeAndLaddersGame:
         if is_reversed:
             x_level = 9 - x_level
         return x_level, y_level
-
-    def _mod_in_game_check(self, user: Member) -> bool:
-        """Return True if user is a Moderator and playing the currently running game."""
-        return all((self._is_moderator(user), user in self.players))
 
     @staticmethod
     def _is_moderator(user: Member) -> bool:
--- a/bot/seasons/evergreen/snakes/utils.py
+++ b/bot/seasons/evergreen/snakes/utils.py
@@ -494,7 +494,7 @@ class SnakeAndLaddersGame:
         """
         Handle players leaving the game.
 
-        Leaving is prevented if the user wesn't part of the game.
+        Leaving is prevented if the user wasn't part of the game.
 
         If the number of players reaches 0, the game is terminated. In this case, a sentinel boolean
         is returned True to prevent a game from continuing after it's destroyed.
@@ -616,7 +616,7 @@ class SnakeAndLaddersGame:
                 if reaction.emoji == ROLL_EMOJI:
                     await self.player_roll(user)
                 elif reaction.emoji == CANCEL_EMOJI:
-                    if self._is_moderator(user) and not self._mod_in_game_check(user):
+                    if self._is_moderator(user) and user not in self.players:
                         # Only allow non-playing moderation staff to cancel a running game
                         await self.cancel_game()
                         return
@@ -713,4 +713,4 @@ class SnakeAndLaddersGame:
     @staticmethod
     def _is_moderator(user: Member) -> bool:
         """Return True if the user is a Moderator."""
-        return Roles.moderator in [role.id for role in user.roles]
+        return any(Roles.moderator == role.id for role in user.roles)
--- a/bot/seasons/evergreen/snakes/utils.py
+++ b/bot/seasons/evergreen/snakes/utils.py
@@ -436,7 +436,7 @@ class SnakeAndLaddersGame:
                 if reaction.emoji == JOIN_EMOJI:
                     await self.player_join(user)
                 elif reaction.emoji == CANCEL_EMOJI:
-                    if user == self.author or all((self._is_moderator(user), user not in self.players)):
+                    if user == self.author or (self._is_moderator(user) and user not in self.players):
                         # Allow game author or non-playing moderation staff to cancel a waiting game
                         await self.cancel_game()
                         return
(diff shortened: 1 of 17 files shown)


bot/seasons/halloween/hacktoberstats.py
```diff
@@ -1,17 +1,20 @@
...
 from pathlib import Path
+from typing import List, Tuple
 
 import aiohttp
 import discord
 from discord.ext import commands
 
 log = logging.getLogger(__name__)
 
+CURRENT_YEAR = datetime.now().year  # Used to construct GH API query
```
MarkKoz: If the extension is loaded before the new year, then after the new year it will still show the wrong date until the extension is reloaded. In practice this season will not be loaded during the new year though 🤔

bot/seasons/evergreen/snakes/utils.py
```diff
@@ -708,3 +705,12 @@ def _board_coordinate_from_index(self, index: int):
...
         return x_level, y_level
+
+    def _mod_in_game_check(self, user: Member) -> bool:
+        """Return True if user is a Moderator and playing the currently running game."""
+        return all((self._is_moderator(user), user in self.players))
+
+    @staticmethod
+    def _is_moderator(user: Member) -> bool:
+        """Return True if the user is a Moderator."""
+        return Roles.moderator in [role.id for role in user.roles]
```
MarkKoz:
```suggestion
        return any(Roles.moderator == role.id for role in user.roles)
```

bot/seasons/evergreen/snakes/utils.py
```diff
@@ -708,3 +705,12 @@ def _board_coordinate_from_index(self, index: int):
         if is_reversed:
             x_level = 9 - x_level
         return x_level, y_level
+
+    def _mod_in_game_check(self, user: Member) -> bool:
+        """Return True if user is a Moderator and playing the currently running game."""
+        return all((self._is_moderator(user), user in self.players))
```
MarkKoz:
```suggestion
        return self._is_moderator(user) and user in self.players
```

The suggestion on line 639 would leave this function being used only on line 439. I don't think this simple condition warrants its own function in such case. Furthermore, based on the comments on line 439, the suggested simplification, which is based on the original code, may not exhibit behaviour which was the original intent anyway.

bot/seasons/evergreen/snakes/utils.py
```diff
@@ -488,20 +490,16 @@ def startup_event_check(reaction_: Reaction, user_: Member):
...
         )
 
-    async def player_leave(self, user: Member):
+    async def player_leave(self, user: Member) -> bool:
         """
         Handle players leaving the game.
 
-        Leaving is prevented if the user initiated the game or if they weren't part of it in the
-        first place.
+        Leaving is prevented if the user wesn't part of the game.
```
MarkKoz:
```suggestion
        Leaving is prevented if the user wasn't part of the game.
```

bot/seasons/evergreen/snakes/utils.py
```diff
@@ -623,11 +616,12 @@ def game_event_check(reaction_: Reaction, user_: Member):
                 if reaction.emoji == ROLL_EMOJI:
                     await self.player_roll(user)
                 elif reaction.emoji == CANCEL_EMOJI:
-                    if self.ctx.author == user:
-                        await self.cancel_game(user)
+                    if self._is_moderator(user) and not self._mod_in_game_check(user):
```
MarkKoz: Since the first condition checks that they are a moderator, using `_mod_in_game_check()` has some redundancy.
```suggestion
                    if self._is_moderator(user) and user not in self.players:
```

bot/seasons/evergreen/snakes/utils.py
```diff
@@ -435,8 +436,9 @@ def startup_event_check(reaction_: Reaction, user_: Member):
                 if reaction.emoji == JOIN_EMOJI:
                     await self.player_join(user)
                 elif reaction.emoji == CANCEL_EMOJI:
-                    if self.ctx.author == user:
-                        await self.cancel_game(user)
+                    if user == self.author or not self._mod_in_game_check(user):
```
MarkKoz: `_mod_in_game_check()` returns `False` if the player is not in the game or if they are not a mod. Thus, this allows for anyone not in the game to cancel it.

bot/seasons/evergreen/snakes/utils.py
MarkKoz: This function may be useful after all. The suggested change on 619 I believe is the same condition that would be needed on line 439:

```suggestion
        return self._is_moderator(user) and user not in self.players
```

bot/seasons/halloween/hacktoberstats.py
sco1: Is there a scenario where this wouldn't be loaded at the beginning of October and unloaded at the end?

bot/seasons/evergreen/snakes/utils.py
sco1: Will fix this & the other logic issues tonight when I get home

bot/seasons/halloween/hacktoberstats.py
MarkKoz: I think only through a season override.

bot/seasons/evergreen/snakes/utils.py
```diff
@@ -435,8 +436,9 @@ def startup_event_check(reaction_: Reaction, user_: Member):
                 if reaction.emoji == JOIN_EMOJI:
                     await self.player_join(user)
                 elif reaction.emoji == CANCEL_EMOJI:
-                    if self.ctx.author == user:
-                        await self.cancel_game(user)
+                    if user == self.author or all((self._is_moderator(user), user not in self.players)):
```
MarkKoz: Using `all()` feels more confusing for this when can just be a simple `and` between two objects.
```suggestion
                    if user == self.author or (self._is_moderator(user) and user not in self.players):
```

PR #273, opened 2019-09-19
Docker Refining & Persistent Data Support

scragly:
I've cleaned up a bit the Dockerfile and docker-compose.yml.

After moving the Build and Push CI step to the azure-pipelines.yml itself, I have removed the shell script that was doing it before. Since we only had a single Dockerfile in the docker directory afterwards, I moved it to the project root to keep things clean.

`docker-compose.yml` was not being used for deployment or CI, so I've changed it for local build testing, the same as PyDis site has.

The Dockerfile has changed to the new minideb base image from bitnami. Due to this change a few benefits resulted:
 - All build tools that were previously installed are no longer needed
 - The size has reduced from 329.3MB to 253.4 MB
 - The build time has been reduced from 56.7s to 19.5s

The egg_hunt extension is broken and needs to be rewritten. To avoid issues when it comes time to develop for Easter, I've removed it completely.

I have added a new utility for persistant datafiles under `bot.utils.persist`. This takes advantage of a properly setup bound volume from the host machine. I have successfully tested this on our saltstate environment using the `seasonalbot-dev` state I recently added.

The `persist` module is simple and currently only has three things:
 - The constant `persist.DIRECTORY` which is a Path object for the directory that is to be mapped to the persistent volume. This is for cases where a new file is created rather than relying on an existing datafile to alter from.
 - The function `datafile` which will automatically copy a default datafile that's given to it to the persistent directory in the case it does not already exist, and will return a Path for the datafile in that persistent directory.
 - The function `sqlite` which does the same as `datafile` for an sqlite database, but will return an SQLite3 Connection to the database ready for usage. As we won't have multiple processes trying to access the same SQLite database, it's less expensive to just keep the db connection open the entire time without worries.

When this PR is merged, I will copy the persistent volume settings across to the main seasonalbot saltstate file. If I add the change now, I'm suspect that it will cause a deployment error as the directory won't be ready for it. The change will also fix a bug we have with persistent log files currently as a side-benefit.


scragly:
> I can't check if the persistent storage works, since it's never been a problem locally. I trust you've taken care of that, though.

Yep, I was sure to test it in saltstack on alpha.

> That said, I do think the `persist.datafile` API can use a better docstring ...  it would be good to have that properly documented in the docstring, so contributors can read that back.

Yep sounds right, I've added hopefully a better explanatory section in the docstring for the persist.datafile function.

Thanks for looking it over @SebastiaanZ

bot/utils/persist.py
```diff
@@ -0,0 +1,40 @@
+import sqlite3
+from pathlib import Path
+from shutil import copyfile
+
+DIRECTORY = Path("data")  # directory that has a persistent volume mapped to it
+
+
+def datafile(file_path: Path) -> Path:
```
lemonsaurus: Not sure I'm crazy about this name. Can we consider some alternatives? `persistant_path` or something seems more descriptive.

scragly: `persist.persistan_path` seems redundant.

azure-pipelines.yml
```diff
@@ -33,20 +33,16 @@ jobs:
...
-      command: 'login'
-
-  - task: ShellScript@2
-    displayName: 'Build and deploy containers'
-
+  - task: Docker@2
+    displayName: 'Build & Push Container'
     inputs:
-      scriptPath: scripts/deploy-azure.sh
+      containerRegistry: 'DockerHubV2'
```
MarkKoz: I believe this is OK now.
```suggestion
      containerRegistry: 'DockerHub'
```

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ENV PIP_NO_CACHE_DIR=false \
+    PIPENV_HIDE_EMOJIS=1 \
+    PIPENV_IGNORE_VIRTUALENVS=1 \
+    PIPENV_NOSPIN=1
+
+WORKDIR /bot
+COPY . .
+
+# Update setuptools by removing egg first, add other dependencies
+RUN rm -r /opt/bitnami/python/lib/python3.*/site-packages/setuptools* && \
```
MarkKoz: This can be moved before `COPY` to improve layer caching.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM bitnami/python:3.7-prod
```
MarkKoz: Are you going to move to the python slim image here too or stick with bitnami for now?

scragly: Yes, this entire thing needs to be updated to the python slim image. I was away yesterday so didn't have the time to work on the bot or seasonalbot dockerfiles.

scragly: This is now actioned.

Dockerfile
scragly: The setuptools install is no longer required with the up-to-date official python image, but the pipenv install was moved to before the `COPY` now, thanks.

bot/utils/persist.py
scragly: I've been trying to think of alternative, but I'm not entirely sure I can think of any super amazing approaches, but your concern is reasonable as it's not super descriptive without context and reading the rest of the docstring.

The `persistent_path` suggestion lends the idea that you're just getting the path for the persistent directory, which is actually what `persist.DIRECTORY` is. I'm also really not a fan of the `double.double` reference that is `persist.persistent_path` or similar. I also kinda think it's less accurate than the current one, as this function also copies the base file to the persistent directory if it's not already there. 

`make_file_persistent` is perhaps closer to what the purpose is, but might feel kinda longish.

@MarkKoz would you have any suggestions?

bot/utils/persist.py
```diff
@@ -0,0 +1,40 @@
...
+    the path to this new persistent version of the file.
+
+    If the persistent file already exists, it won't be overwritten with the
+    clean default file, just returning the Path instead to the existing file.
+
+    Example Usage:
+    >>> clean_default_datafile = Path("bot", "resources", "datafile.json")
+    >>> persistent_file_path = datafile(clean_default_datafile)
+    """
+    if not file_path.exists():
```
MarkKoz: Probably a good idea to check it's a file and not a directory... just in case. It also doubles as check for if the file exists. It may be worth revising the exception if this is changed though.
```suggestion
    if not file_path.is_file():
```

bot/utils/persist.py
MarkKoz: I'll give it a shot with some ideas:

* `persist.copy()`
* `persist.make_persistent()`
* `persist.copy_datafile()`
* `persist.add()`

bot/utils/persist.py
```diff
@@ -0,0 +1,40 @@
...
+        raise OSError(f"File not found at {file_path}.")
+
+    persistant_path = Path(DIRECTORY, file_path.name)
+
+    if not persistant_path.exists():
```
MarkKoz: Could this lead to a scenario where the path exists as a directory, but a file with an identical path is trying to be added? Thus it won't be added even though it technically does not exist. Edge case I guess...

scragly: Definitely edge case since filetypes should have an extension present while directories won't. However, I think I'll actually organise these datafiles in the persistent directory properly into subdirectories to avoid the possibility of two features each having a datafile with the same name and conflicting in that manner and to keep things cleaner. At the moment it just dumps all the datafiles in the one directory.

bot/utils/persist.py
```diff
@@ -0,0 +1,68 @@
...
+    >>> import json
+    >>> template_datafile = Path("bot", "resources", "evergreen", "myfile.json")
+    >>> path_to_persistent_file = make_persistent(template_datafile)
+    >>> print(path_to_persistent_file)
+    data/evergreen/myfile.json
+    >>> with path_to_persistent_file.open("w+") as f:
+    >>>     data = json.load(f)
+    """
+    # ensure the persistent data directory exists
+    if not DIRECTORY.exists():
```
MarkKoz: Can skip this step by doing `DIRECTORY.mkdir(exist_ok=True)`

bot/utils/persist.py
```diff
@@ -0,0 +1,68 @@
...
+        DIRECTORY.mkdir()
+
+    if not file_path.is_file():
+        raise OSError(f"File not found at {file_path}.")
+
+    # detect season in datafile path for assigning to subdirectory
+    season = next((s for s in get_seasons() if s in file_path.parts), None)
```
MarkKoz: I foresee some false matches here but I'd prefer something simple like this over a much longer solution that's more robust for little gain.

bot/utils/persist.py
```diff
@@ -0,0 +1,68 @@
...
+
+    if season:
+        # make sure subdirectory exists first
+        subdirectory = Path(DIRECTORY, season)
+        if not subdirectory.exists():
```
MarkKoz: Can skip this step by doing `subdirectory.mkdir(exist_ok=True)`

bot/utils/persist.py
scragly: Yeah, I thought the same haha.

bot/utils/persist.py
scragly: TIL of a new kwarg, nice. Cheers for mentioning, this is cleaner.

bot/utils/persist.py
scragly: ditto

bot/seasons/halloween/hacktoberstats.py
```diff
@@ -21,7 +23,7 @@ class HacktoberStats(commands.Cog):
 
     def __init__(self, bot: commands.Bot):
         self.bot = bot
-        self.link_json = Path("bot/resources/github_links.json")
+        self.link_json = datafile(Path("bot", "resources", "halloween", "github_links.json"))
```
SebastiaanZ: This one as well:

```suggestion
        self.link_json = make_persistent(Path("bot", "resources", "halloween", "github_links.json"))
```

bot/seasons/halloween/hacktoberstats.py
```diff
@@ -10,6 +10,8 @@
 import discord
 from discord.ext import commands
 
+from bot.utils.persist import datafile
```
SebastiaanZ: I thought I had added this to the previous review, but apparently not:

```suggestion
from bot.utils.persist import make_persistent
```

Issue #283, opened 2019-10-01
Update Channel Whitelist to Include the 2019 Hacktoberfest Event Channel


sco1:
Seasonalbot commands are limited by a decorator to a subset of the server's channels, which currently does not include our 2019 Hacktoberfest event channel.

This channel (ID: `628184417646411776`) should be added to Seasonalbot's constants:

https://github.com/python-discord/seasonalbot/blob/526dbc465c2d458c893c72b614282937d15190d7/bot/constants.py#L23

And the whitelist updated:

https://github.com/python-discord/seasonalbot/blob/526dbc465c2d458c893c72b614282937d15190d7/bot/constants.py#L124-L129

Since this is a fairly minimal change, it may be combined with #282

sco1:
Closed by #285

PR #285, opened 2019-10-01
Hacktober date range and channel whitelist fixes

### LINKED ISSUE #282: Hacktoberstats GH Query Timezone Incorrect
New to Hacktoberfest 2019 is a change to the date window used to validate PRs that count towards the goal. Where last year the timezones were based on UTC, for this year's event a PR is valid as long as it was made when somewhere in the world was beyond October 1st.

This will require an adjustment to the query used by the Hacktoberstats command:

https://github.com/python-discord/seasonalbot/blob/526dbc465c2d458c893c72b614282937d15190d7/bot/seasons/halloween/hacktoberstats.py#L219-L232

From [the internet](https://en.wikipedia.org/wiki/UTC%2B14:00), the first timezone is UTC+14, which can be added to the [GH query](https://help.github.com/en/articles/searching-issues-and-pull-requests#search-by-when-an-issue-or-pull-request-was-created-or-last-updated) used to filter PRs.

Since this is a fairly minimal change, it may be combined with #283

Numerlor:
This PR extends the date ranges of the search query to cover the beginning of October in the first timezone and the end in the last timezone;
Adds the 2019 Hacktoberfest channel to the white-list for bot commands.

Closes #282 #283

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -52,6 +52,7 @@ class Channels(NamedTuple):
     python_discussion = 267624335836053506
     show_your_projects = int(environ.get("CHANNEL_SHOW_YOUR_PROJECTS", 303934982764625920))
     show_your_projects_discussion = 360148304664723466
+    hacktoberfest_2019 = 628184417646411776
 
 
 class Client(NamedTuple):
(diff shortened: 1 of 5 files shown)


lemonsaurus:
@Numerlor Hey, thanks for the PR! This is more or less the right idea, but not quite right. I'd like to allow only the `.hacktoberstats` command in this channel, not all seasonalbot commands.

Instead of adding this channel to the whitelist constant that's used everywhere, maybe just append this channel to the whitelist on that specific command

Numerlor:
Changed the white-list behavior but I'm not sure if it's the right way, had some trouble because the command is a group and now it checks 2 times

Numerlor:
Added logging in case the overridden check fails that I didn't notice yesterday

PR #289, opened 2019-10-02
Put role bypass check above channel override check

Numerlor:
In the rewritten in_channel_check, if the command override was used, role bypass wouldn't take effect because the function would exit before it. This PR puts the bypass check above the override check to make sure the roles will bypass it.


Numerlor:
Also added a fix for the last day of October not being completely included in the search query

PR #292, opened 2019-10-03
Pride facts

### LINKED ISSUE #216: [Feature Request] Fact of the Day
What do those letters in LGTBQI+ stand for? Which countries have legalized gay marriage? What's the largest Pride event in the world? There are many interesting and informative facts that are somehow related to Pride and/or the LGTBQI+ community; it's your task to write a command that supplies our users with a new fact every day of our event.

The facts you include do not necessarily have to be the kind of facts you typically find in clickbait articles, but can also contain general information on Pride, LGTBQI+, and related subjects. In fact, having a couple of informational facts in there is preferable. 

## Implementation details

Your command should produce a different fact depending on the day it was called on. To avoid timezone issues, either use the local time of the bot or UTC for midnight. The order of the facts does not have to be randomized, you can just use the order of your static resource (e.g., a json file) as the order of facts.

Since it would be a shame if facts could only be viewed on a single day, think of a way to let users browse the previous facts of the day. You could have the user provide a date argument to the command, an interactive reaction interface beneath the embed, or any other method you think will work. However, make sure that it doesn't expose future facts of the day, just to keep things a bit surprising.

#### Implementation checklist:
- [ ] Research and write down ~31 facts about Pride and the LGBTQI+ community
- [ ] Write a command that shows today's Fact of the Day to the user
- [ ] Think of an interface that lets the user browse previous (but not future) facts of the day
- [ ] Make a pull request including screenshots of your command in action

**Note:** You should aim for having a fact for each day of the event, including the days that have already passed at the time of your pull request. In addition, make sure to calculate the current day relative to the start date of the event, as specified in the `__init__` instead of hard-coding it in your cog. This makes sure your cog will also work if the event gets repeated in the future. (In fact, it's probably a good idea to have 31 facts to support 31-day months.)

## Additional information

- Keep us updated on your progress by leaving comments!
- Make sure to give you commits a descriptive commit message
- Please note the style guide and conventions listed in the various contrib documents
- If you want to include graphics, make sure they have a permissive license

WillDaSilva:
closes #216 

A new fact is made available from the static resource each day during the pride season. Currently only 2020 has facts written for it. When the command is used with no arguments, the daily fact is provided. If you use "random" or "rand" as an argument you get a random fact chosen from previous facts, and the daily fact for the current day. You can also provide a specific date or day of the month, so long as that day is not in the future.

UTC time was used for determining what day it is. I worked under the assumption that the pride season for the bot is only ever June, which has 30 days. None of the logic in the code here deals with months, because so long as the command is available, it is assumed that the current month is June. Likewise, the facts JSON file has facts per year, rather than per month.

In the following screenshot I spoofed the current date, such that the bot thought it was 2020-06-09. Because the date was being spoofed in a simple way, the date parser wasn't working quite right. Normally you wouldn't have to write the full date, but rather just the day of the month (if you're viewing a fact for the current year). 

Selection_011

It would be appreciated if someone could fact check the facts I've collected for 2020.


RohanJnr:
maybe you can also add a background task that sends a fact every day/every 12 hours or so :thinking:

Akarys42:
BTW instead of writing “issue: #216”, you can write “closes #216” and the issue will automatically closed when the PR is merged (see https://help.github.com/en/articles/closing-issues-using-keywords) 😀

WillDaSilva:
> BTW instead of writing “issue: #216”, you can write “closes #216” and the issue will automatically closed when the PR is merged (see https://help.github.com/en/articles/closing-issues-using-keywords) grinning

I've updated it. Hopefully that works too.

bot/seasons/pride/pride_facts.py
```diff
@@ -0,0 +1,106 @@
...
+            return json.load(f)
+
+    async def send_pride_fact_daily(self) -> None:
+        """Background task to post the daily pride fact every day."""
+        channel = self.bot.get_channel(Channels.seasonalbot_chat)
+        while True:
+            await self.send_select_fact(channel, datetime.utcnow())
+            await asyncio.sleep(24 * 60 * 60)
+
+    async def send_random_fact(self, ctx: commands.Context) -> None:
```
sco1: I think it would be better for random facts to use all of the facts as the pool from which to draw from rather than limiting it to the previous + current years.

WillDaSilva: > I think it would be better for random facts to use all of the facts as the pool from which to draw from rather than limiting it to the previous + current years.

That would go against what was requested in the original issue:
> However, make sure that it doesn't expose future facts of the day, just to keep things a bit surprising.

@SebastiaanZ Thoughts?

I don't have strong feelings one way or the other.

sco1: That's fine then, I hadn't read the issue itself.

Issue #293, opened 2019-10-03
Limit the AoC and issue commands to certain channels instead of being allowed everywhere


Numerlor:
Currently, the AoC and issue commands override the global channel whitelist which allows them to be used in any channel.
The decorator was rewritten to accept a new channel whitelist that will be used in place of the global whitelist, this makes the implementation quite simple:

## Implementation details
- [ ] Construct a new whitelist and pass it to `override_in_channel` in issues
- [ ] Construct a new whitelist and pass it to `override_in_channel` in adventofcode

PR #294, opened 2019-10-03
Limit the issue and AoC commands to a superset of the global whitelist instead of bypassing it entirely

Numerlor:
Limits the issue and AoC command groups to the global whitelist + the seasonal bot chat and the AoC channel respectively

Closes: #293


RohanRadia:
Don't include the issues command in this for now. I am working on an updated issues command. Wait for it to be merged then update your PR please.

sco1:
There's no reason you can't merge the changes in if this PR is merged before yours.

RohanRadia:
> There's no reason you can't merge the changes in if this PR is merged before yours.

Was just saying as I could sort it out in that PR but that is fair enough 👌

Numerlor:
> The in-channel overrides are not inherited by subcommands so they need to be explicitly defined.
> 
> This is also an issue with the hacktoberstats cog, it can be fixed with this PR

I was going to change the decorator to work on the cog level instead of per-command basis in a second PR after this one was merged, would it be alright to include it in here?

sco1:
The hackstats subcommands are currently broken in the whitelisted channels, so a fix should be included even if it is temporary.

Numerlor:
added the decorators to the whole hacktober command group

bot/seasons/christmas/adventofcode.py
```diff
@@ -126,7 +128,7 @@ def __init__(self, bot: commands.Bot):
         self.status_task = asyncio.ensure_future(self.bot.loop.create_task(status_coro))
 
     @commands.group(name="adventofcode", aliases=("aoc",), invoke_without_command=True)
-    @override_in_channel()
+    @override_in_channel(AOC_WHITELIST)
```
lemonsaurus: didn't we discover that overriding groups didn't work? isn't that why you're overriding individual commands in the hacktoberfest cog instead?

Numerlor: the decorator only works for individual commands with how it is now, I've only added it in on every hacktober command because it's an out of scope hotfix and the AoC commands aren't available.
I was planning to do a PR that would change the decorator to modify the cog class instead of the commands coros after this

lemonsaurus: I think maybe it's overengineering to change the decorator to have it modify the cog class, besides, we'd lose precision. I mean, we still want a command decorator in case we only want one or two commands in a cog to be available, or perhaps that different commands in the same cog need different whitelists.

Ultimately I kind of think we should just apply the current decorator to all the commands in the AoC cog as well and then leave it like that.

Numerlor: Fair enough, just thought it'd help with the repeating decorators when it isn't used individually anywhere yet. I'll look if there's a good way to do it by groups but don't have much hope in that

lemonsaurus: I'm not that bothered by a few extra decorators, to be honest. one could argue that this has better readability anyway, since it's more explicit.

lemonsaurus: ..But shouldn't this line now be removed, since you've added the decorators to the individual commands instead?

Numerlor: it's still a command coro since it has `invoke_without_command`, although the help it provides may not be needed to be allowed in the aoc channel

lemonsaurus: oh yeah that's true. no, then it's fine.

PR #298, opened 2019-10-07
Quiz command fix

RohanJnr:
The quiz command was facing some issues regarding the scoreboard and a little confusion with the arguments. This PR will fix all that as I have implemented the following:
- Every game gets a new scoreboard
- Updated docstrings.
- Storing all player's points in a dict and it refreshes when the bot restarts.

Additional Info:
- Added a bit of documentation for the code(comments I mean).
- To view the scoreboard which stores everyone's score, I have made a command for it.


bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -139,36 +149,56 @@ def check(m: discord.Message) -> bool:
...
                 hint_no = 0
-                unanswered = 0
+
                 await ctx.send(f"{msg.author.mention} got the correct answer :tada: {points} points for ya.")
                 await self.send_answer(ctx.channel, question_dict)
-                await self.send_score(ctx.channel, self.player_dict[ctx.channel.id])
+                await self.send_score(ctx.channel, self.game_player_scores[ctx.channel.id])
                 await asyncio.sleep(2)
 
+    @commands.command(name="scoreboard")
```
kosayoda: The scoreboard command shouldn't be a top level command, but in a `quiz` command group.

I am also not sure if the command should be called `scoreboard`, as it is actually a bot session leaderboard. The scoreboard would be the current ranking of players in a current game. `.quiz leaderboard` would be a better name.

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -139,36 +149,56 @@ def check(m: discord.Message) -> bool:
...
+    async def overall_scoreboard(self, ctx: commands.Context) -> None:
+        """View everyone's score for this bot session."""
+        await self.send_score(ctx.channel, self.player_scores)
+
     @staticmethod
     async def send_score(channel: discord.TextChannel, player_data: dict) -> None:
         """A function which sends the score."""
         embed = discord.Embed(colour=discord.Colour.blue())
         embed.title = "Score Board"
```
kosayoda: Similarly, `Leaderboard` would be a better term.

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -139,36 +149,56 @@ def check(m: discord.Message) -> bool:
...
         embed.description = ""
+        if len(player_data) == 0:
+            await channel.send("No one has made it to the scoreboard yet.")
```
kosayoda:
```suggestion
            await channel.send("No one has made it to the leaderboard yet.")
```

RohanJnr:
updated.

RohanJnr:
Working on it, thanks for the review 😀

RohanJnr:
updated

RohanJnr:
will work on it. thanks! and sorry for the problem :sweat:

RohanJnr:
There is one way to stop that error and that is cool down, maybe around 10s cooldown could be a good one

RohanJnr:
i will add some additional checks split the big functions into function and will also add a cooldown.

RohanJnr:
btw are the doc strings and comments all good?

RohanJnr:
updated

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -45,71 +48,66 @@ def load_questions() -> dict:
             questions = json.load(json_data)
             return questions
 
-    @commands.command(name="quiz", aliases=["trivia"])
+    @cooldown(1, 20, BucketType.channel)
```
SebastiaanZ: This `cooldown` also prevents moderators from stopping a quiz as long as it's active. I'm not even sure if the user who started the quiz should be unable to stop it due to a cooldown, but moderators should certainly be able to stop it without having to wait for the cooldown to finish.

I think this ties into this command method being one monolithic callable that tries to perform multiple actions, including both starting and stopping the quiz. A separate command to stop the quiz, like `!quiz stop` would solve that.

RohanJnr: Well, it would not make sense to stop the game right after starting it, so that was my plan for adding the cooldown.
I switched to using the single command instead of a command group having 2 commands `start` and `stop` as it was suggested to me.
I can add it back or would it be better to just remove the cooldown?

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -14,10 +16,8 @@
 logger = logging.getLogger(__name__)
 
 
-ANNOYED_EXPRESSIONS = ["-_-", "-.-"]
-
 WRONG_ANS_RESPONSE = [
-    "No one gave the correct answer",
+    "No answered correctly!",
```
kwzrd: Looks like there's a word missing here.
```suggestion
    "No one answered correctly!",
```

RohanJnr: got it

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -28,88 +26,87 @@ class TriviaQuiz(commands.Cog):
...
+                f"Game is already running..."
+                f"do `{self.bot.command_prefix}quiz stop`"
+            )
+            await self.stop_quiz(ctx.author, ctx.channel)
+            return
 
+        category = category.lower()
+        # Send embed showing available categories if inputted category is invalid.
         if category not in self.categories:
             embed = self.category_embed
```
MarkKoz: It's weird that `category_embed` is a property. Being in-line with `make_start_embed`, I think it's better as a function. Or, you could just make them both attributes set only once in `__init__` since their content is not dynamic as far as I can tell.

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -121,56 +118,120 @@ def load_questions() -> dict:
...
+
+                # if number of hints sent or time alerts sent is less than 2, then send one.
                 if hint_no < 2:
                     hint_no += 1
                     if "hints" in question_dict:
                         hints = question_dict["hints"]
                         await ctx.send(f"**Hint #{hint_no+1}\n**{hints[hint_no]}")
                     else:
-                        await ctx.send(f"Cmon guys, {30-hint_no*10}s left!")
+                        await ctx.send(f"{30-hint_no*10}s left!")
```
MarkKoz:
```suggestion
                        await ctx.send(f"{30 - hint_no * 10}s left!")
```

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -121,56 +118,120 @@ def load_questions() -> dict:
...
                 embed.title = f"Question #{len(done_question)}"
                 embed.description = q
-                await ctx.send(embed=embed)
+                await ctx.send(embed=embed)  # Send question embed.
 
+            # A function to check whether user input is the correct answer(close to the right answer)
             def check(m: discord.Message) -> bool:
                 ratio = fuzz.ratio(answer.lower(), m.content.lower())
                 return ratio > 85 and m.channel == ctx.channel
             try:
```
MarkKoz:
```suggestion

            try:
```

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -121,56 +118,120 @@ def load_questions() -> dict:
...
+            await self.stop(ctx.author, ctx.channel)
+        else:
+            await ctx.send("No quiz running.")
+
+    async def stop(self, author: discord.Member, channel: discord.TextChannel) -> None:
+        """Stop the quiz."""
+        # Check if the author is the game starter or a moderator.
+        if (
+                author == self.game_owners[channel.id]
+                or any(Roles.moderator == role.id for role in author.roles)
```
MarkKoz: These are indented too far.

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -121,56 +118,120 @@ def load_questions() -> dict:
...
+
+    @quiz_game.command(name="stop")
+    async def stop_quiz(self, ctx: commands.Context) -> None:
+        """Stop a quiz game if its running in the channel."""
+        if self.game_status[ctx.channel.id] is True:
+            await self.stop(ctx.author, ctx.channel)
+        else:
+            await ctx.send("No quiz running.")
+
+    async def stop(self, author: discord.Member, channel: discord.TextChannel) -> None:
```
MarkKoz: Why is this a separate function? If it is kept as part of the command, decorators for checks can also be used (though maybe they shouldn't be if it adds more complexity, especially given that there is a custom error message).

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -121,56 +118,120 @@ def load_questions() -> dict:
...
         embed.title = "Score Board"
         embed.description = ""
-        for k, v in player_data.items():
-            embed.description += f"{k} : {v}\n"
+        if len(player_data) == 0:
+            await channel.send("No one has made it onto the leaderboard yet.")
+            return
+        sorted_dict = sorted(player_data.items(), key=lambda a: a[1], reverse=True)
+        for item in sorted_dict:
+
```
MarkKoz:
```suggestion
```

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -121,56 +118,120 @@ def load_questions() -> dict:
...
     @staticmethod
     async def send_score(channel: discord.TextChannel, player_data: dict) -> None:
         """A function which sends the score."""
         embed = discord.Embed(colour=discord.Colour.blue())
         embed.title = "Score Board"
         embed.description = ""
-        for k, v in player_data.items():
-            embed.description += f"{k} : {v}\n"
+        if len(player_data) == 0:
```
MarkKoz: This check could be moved to the start of the function.

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -28,88 +26,87 @@ class TriviaQuiz(commands.Cog):
     def __init__(self, bot: commands.Bot) -> None:
         self.bot = bot
         self.questions = self.load_questions()
-        self.game_status = {}
-        self.game_owners = {}
+        self.game_status = {}  # A variable to store the game status: either running or not running.
```
MarkKoz: It strikes me as simpler to maintain a single object (probably a dataclass) to represent game state instead of 4 separate dictionaries. However, I acknowledge this has been in review for a while already so this change can be held off for now considering the current approach is functional.

RohanJnr: The old version of this quiz used a data class as there were many dictionaries but I had to ditch that code as I could not fix errors such as to reset the quiz game in a channel.
I tried a new solution given by Kosa which is working really good now and it does not require many dictionaries so thats the reason I am not using a data class.

bot/seasons/evergreen/trivia_quiz.py
RohanJnr: I think I will put the Stop function code inside the stop quiz command.

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -28,149 +26,214 @@ class TriviaQuiz(commands.Cog):
...
-            else:
-                await ctx.send(f"{ctx.author.mention}, you are not authorised to stop this game :ghost: !")
+        # Stop game if running.
+        if self.game_status[ctx.channel.id] is True:
+            return await ctx.send(
+                f"Game is already running..."
+                f"do `{self.bot.command_prefix}quiz stop`"
+            )
+            await self.stop_quiz(ctx.author, ctx.channel)
+            return
```
MarkKoz: This doesn't make sense. The message is sent with a return, so the quiz is never stopped; lines 71 and 72 are never executed. This also contradicts the behaviour documented:

> If the quiz game is running, then the owner or a mod can stop it by using this command without providing any arguments or vice versa.

And if it did stop the quiz, then there'd be no need to put 

```py
f"do `{self.bot.command_prefix}quiz stop`"
```

So what do you really want to happen here? In my opinion, stopping the quiz should be explicit. Therefore, the only command that should stop the quiz is the stop command.

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -185,33 +248,34 @@ def check(m: discord.Message) -> bool:
...
+                f"Congratulations {winners_mention} :tada: "
+                f"{word} have won this quiz game with a grand total of {highest_points} points!"
             )
 
-    @property
     def category_embed(self) -> discord.Embed:
         """Build an embed showing all available trivia categories."""
         embed = discord.Embed(colour=discord.Colour.blue())
         embed.title = "The available question categories are:"
+        embed.set_footer(text="If a category is not chosen, a random one will be selected.")
```
MarkKoz: I recognise there is currently only one category. However, technically speaking this is not true. If no category is given, it will default to the general category; no randomness is involved. Were you not planning on implementing that now?

RohanJnr: I don't have time to add other categories, maybe later I will.
This PR was not to add more categories, It was to fix a problem mentioned in the PR description.

bot/seasons/evergreen/trivia_quiz.py
RohanJnr: i forgot to clean that mess up, thanks

bot/seasons/evergreen/trivia_quiz.py
MarkKoz: I am asking about implementing randomness not adding more categories.

RohanJnr: There is only 1 category available so cant implement random now.

RohanJnr: Initially, I had a plan to add categories for every season(Halloween, Christmas, valentines, easter, pride, etc) but I guess that has to wait.

MarkKoz: You can still implement the randomness now even if there is currently only one category. It'll just always pick the same category until more are added. However, if you don't want to, that's OK.

RohanJnr: oooh like that, I think its gonna be better, I will do it.

bot/seasons/evergreen/trivia_quiz.py
```diff
@@ -28,149 +26,215 @@ class TriviaQuiz(commands.Cog):
...
+        if self.game_status[ctx.channel.id] is True:
+            return await ctx.send(
+                f"Game is already running..."
+                f"do `{self.bot.command_prefix}quiz stop`"
+            )
+
+        # Send embed showing available categori es if inputted category is invalid.
+        if category is None:
+            category = random.choice(list(self.categories))
+        if category.lower() not in self.categories:
```
MarkKoz:
```suggestion

        category = category.lower()
        if category not in self.categories:
```

MarkKoz:
@RohanJnr Did you uncheck "allow edits from maintainers" or is it some GitHub problem that is not allowing me to commit the suggestions myself?

bot/seasons/evergreen/trivia_quiz.py
RohanJnr: This wont work cuz if category is None, then it will throw error

MarkKoz: The category will never be none because that is checked right above this line.

RohanJnr:
allow edits from maintainers is checked, I dint touch that, maybe its some problem with github.

bot/seasons/evergreen/trivia_quiz.py
RohanJnr: nvm  i get it, its coming after the 1st if , so i guess it should be fine.

RohanJnr: what difference will it make btw?

MarkKoz: > Fix KeyError if quiz category is not lowercase

MarkKoz:
Alright then you try committing them.

Issue #213, opened 2019-05-28
[Feature Request] Pride Movie Generator


SebastiaanZ:
What a SeasonalBot event without a movie generator? There are plenty of movies out there that showcase, celebrate, or are otherwise associated with the LGBTQI+ community. For this feature, it's your job to write a command that suggests such a movie to the user.

## Implementation details

Instead of suggesting a random movie, it would be nice if the suggestion was based on the movie preferences of the user. So, use your creativity and come up with a system to suggest movies to the user in desperate use of a movie. You could think of having the user provide a movie genre, an interactive yes/no/choices question/decision tree ("Scary? yes/no", "Romcom or Action"...) or any other method you think of.

- [ ] Think of a unique command name (suggestion: `.pridemovie`)
- [ ] Make some sort of static resource with movies
- [ ] Code the command following the contributors docs
- [ ] Make a Pull Request detailing your implementation, preferably including screenshots as well 

## Additional information

- Keep us updated on your progress by leaving comments!
- Make sure to give you commits a descriptive commit message
- Please note the style guide and conventions listed in the various contrib documents
- There have been other movie generators for previous events, so feel free to check to out for inspiration
- If you want to include graphics, make sure they have a permissive license

RohanJnr:
This can be merged with generalized movie command #345

Akarys42:
Closed by #350

PR #306, opened 2019-10-13
Generic Movie Embed Generator

sgh304:
---
Generic Movie Embed Generator
---
Adds a generic method for generating random movie embeds.

## Pull Request Details
This abstracts out the logic from the Valentine's Day movie generation command into a util function.

This will make creating new movie generation commands easier, making them just a function call with different parameters. I've gone ahead and converted the existing `romancemovie` command to use this new function. I also plan to use this framework in a future PR to tackle issue #213.

### DIFF
--- a/bot/seasons/valentines/movie_generator.py
+++ b/bot/seasons/valentines/movie_generator.py
@@ -1,12 +1,6 @@
 import logging
-import random
-from os import environ
-from urllib import parse
-
-import discord
 from discord.ext import commands
-
-TMDB_API_KEY = environ.get("TMDB_API_KEY")
+from bot.utils.movies import send_random_movie_embed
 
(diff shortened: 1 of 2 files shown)


sco1:
I'm a fan of this approach with one slight tweak: rather than having a utility function, can you move this into an movie recommendation Evergreen cog so we can get movie recommendations year-round? The seasonal-specific commands should still be able to call it with their specific genre flags.

With the above in mind, it should be fine to include #213 with this PR.

ks129:
Can I take this development over due current currently this development is pretty dead?

SebastiaanZ:
This PR does look stale and I think it's okay for someone else to take over the development at this point.

sco1:
More or less superseded by #350

Issue #310, opened 2019-11-01
Advent of Codes configuration need to change.


eivl:
This is from [constants.py](https://github.com/python-discord/seasonalbot/blob/master/bot/constants.py) and it has the year hard-coded.

```python
class AdventOfCode:
    leaderboard_cache_age_threshold_seconds = 3600
    leaderboard_id = 363275
    leaderboard_join_code = str(environ.get("AOC_JOIN_CODE", None))
    leaderboard_max_displayed_members = 10
    year = 2018
    role_id = int(environ.get("AOC_ROLE_ID", 518565788744024082))
```
This should be changed to the 2019 and maybe we need to do something else and use the constant as a fallback.

We can get the year from datetime, but then you could not access the commands on the first of January. 

The API have access to the event before it starts, so we might just get it from there.

kosayoda:
Not sure whether it should be constant, one way I see would be to use the current year if the month is December, else the previous year.

SebastiaanZ:
Do we want the commands to be accessible in January? The Advent of Code ends with Christmas, on the 25th of December and "our" seasonalbot" season ends at midnight UTC on the 31st. That means extracting it from `datetime.utcnow()` should work well. 

By the way, this class does not control the availability of commands; the start and end date of the seasonalbot event itself are specified in the [`__init__` file for Christmas](https://github.com/python-discord/seasonalbot/blob/master/bot/seasons/christmas/__init__.py) and it's handled by the season manager.

PR #313, opened 2019-11-18
Advent of Code year binds to datetime instead of a magic number

eivl:
---
name: AoC year fix
issue: Issue #310 
---
 

## Pull Request Details
* Previous the year was a magic number. This PR adds datetime.utcnow() to fix this.
  * Support for an `AOC_YEAR` env var is also added for local debugging
* Leaderboard ID updated

Closes: #310

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -17,7 +17,7 @@ class AdventOfCode:
     leaderboard_id = 631135
     leaderboard_join_code = str(environ.get("AOC_JOIN_CODE", None))
     leaderboard_max_displayed_members = 10
-    year = datetime.utcnow().year
+    year = int(environ.get("AOC_YEAR", datetime.utcnow().year))
     role_id = int(environ.get("AOC_ROLE_ID", 518565788744024082))
 
 
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -1,6 +1,7 @@
 import logging
 from os import environ
 from typing import NamedTuple
+from datetime import datetime
 
 __all__ = (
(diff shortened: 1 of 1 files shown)


bot/constants.py
```diff
@@ -16,7 +17,7 @@ class AdventOfCode:
     leaderboard_id = 363275
     leaderboard_join_code = str(environ.get("AOC_JOIN_CODE", None))
     leaderboard_max_displayed_members = 10
-    year = 2018
+    year = datetime.utcnow().year
```
sco1: More for SA than anything else since it doesn't functionally affect the functionality of the PR: the TZ of AOC mirrors that of its maintainer (EST, UTC-5).

sco1: For local debugging it might be nice to have an env var to override this if it's being worked on after the New Year but before December 1st.

eivl: this is because seasonal bot uses this TZ. @SebastiaanZ do you have a comment?

eivl: but isnt the command linked to the season that ends at new year?

SebastiaanZ: Yes, but we have commands to manually load seasons so you can easily activate one outside of its regular time period. An optional environment variable with the current year as the default value would work well for that since you'd be able to work with last year's edition in lieu of the not-yet-available edition of the current year.

SebastiaanZ: I don't really have an opinion on this. 

This value is equal the whole year round and AoC does not overlap with a change of its value, so I don't think it's really that relevant for us if our year change syncs perfectly with the year change of the AoC servers.

sco1: I needed this change for debugging so I went ahead and made it

PR #317, opened 2019-11-30
Fix AOC countdown logic

sco1:
The current time delta until the next AOC event assumes that the next event is next year's. While this is was a safe assumption when written, since the command would not be available until the season is loaded on December 1st, it provides an incorrect answer if the season is loaded prior: 

image

The logic has been adjusted to return the closest December 1st that is not in the past. The feedback string has also been adjusted to give hours remaining if we're less than a day away from the event starting:

image
(The second invocation uses November 1st for testing purposes)

The information embed has also been updated to remove the hardcoded year from the link to the global leaderboard. The hardcoded year is unnecessary & the site should redirect to the current event so we don't need to update this value every year.

### DIFF
--- a/bot/seasons/christmas/adventofcode.py
+++ b/bot/seasons/christmas/adventofcode.py
@@ -170,10 +170,21 @@ class AdventOfCode(commands.Cog):
         """Return time left until next day."""
         if not is_in_advent():
             datetime_now = datetime.now(EST)
-            december_first = datetime(datetime_now.year + 1, 12, 1, tzinfo=EST)
-            delta = december_first - datetime_now
+
+            # Calculate the delta to this & next year's December 1st to see which one is closest and not in the past
+            this_year = datetime(datetime_now.year, 12, 1, tzinfo=EST)
+            next_year = datetime(datetime_now.year + 1, 12, 1, tzinfo=EST)
+            deltas = (dec_first - datetime_now for dec_first in (this_year, next_year))
+            delta = min(delta for delta in deltas if delta >= timedelta())  # timedelta() gives 0 duration delta
+
+            # Add a finer timedelta if there's less than a day left
+            if delta.days == 0:
+                delta_str = f"approximately {delta.seconds // 3600} hours"
+            else:
+                delta_str = f"{delta.days} days"
+
             await ctx.send(f"The Advent of Code event is not currently running. "
-                           f"The next event will start in {delta.days} days.")
+                           f"The next event will start in {delta_str}.")
             return
 
         tomorrow, time_left = time_left_to_aoc_midnight()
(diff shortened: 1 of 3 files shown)


bot/seasons/christmas/adventofcode.py
```diff
@@ -170,10 +170,21 @@ def __init__(self, bot: commands.Bot):
...
         if not is_in_advent():
             datetime_now = datetime.now(EST)
-            december_first = datetime(datetime_now.year + 1, 12, 1, tzinfo=EST)
-            delta = december_first - datetime_now
+
+            # Calculate the delta to this & next year's December 1st to see which one is closest and not in the past
+            this_year = datetime(datetime_now.year, 12, 1, tzinfo=EST)
+            next_year = datetime(datetime_now.year + 1, 12, 1, tzinfo=EST)
+            deltas = (dec_first - datetime_now for dec_first in (this_year, next_year))
+            delta = min(delta for delta in deltas if delta >= timedelta())  # timedelta() gives 0 duration delta
```
MarkKoz: Why is `if delta >= timedelta()` needed? Is this for the edge case that the delta is calculated on the exact moment the event starts?

sco1: To filter negative timedeltas

MarkKoz: Never mind. I thought the event lasted all of December so negatives would not be possible in practice. But it only goes until the 25th.

sco1:
I've added a fix to the season check failing for the Christmas season. The `SeasonBase` logic for checking whether the current date falls in a season assumes that the season's start and end dates fall within the same year:

```py
    @classmethod
    def start(cls) -> datetime.datetime:
        if not cls.start_date:
            return datetime.datetime.min
        return datetime.datetime.strptime(f"{cls.start_date}/{cls.current_year()}", cls.date_format)

    @classmethod
    def end(cls) -> datetime.datetime:
        if not cls.end_date:
            return datetime.datetime.max
        return datetime.datetime.strptime(f"{cls.end_date}/{cls.current_year()}", cls.date_format)

    @classmethod
    def is_between_dates(cls, date: datetime.datetime) -> bool:
        return cls.start() <= date < cls.end()
```

This is a fine assumption for every season except Christmas, whose end date falls on New Year's Day, which is the next year:

```
now:             2019-12-03 00:16:36.337083
christmas start: 2019-12-01 00:00:00
end:             2019-01-01 00:00:00
```

This fails the `.is_between_dates()` check, so the Christmas season can never be loaded.

To fix, the Christmas season's `.end()` method is overloaded to add a year to the end date, allowing the in-season check to be performed successfully.

sco1:
Merging this in so we don't have to keep resetting the season

Issue #335, opened 2019-12-15
.aoc leaderboard #deleted-channel


Lvl4Sword:
When ``.aoc leaderboard`` is ran in a channel that it isn't allowed in, the alert that people get:

> Not gonna happen.
Sorry, but you may only use this command within #bot-commands, #seasonalbot-commands, #ot0-everyone-is-lemon-except-you, #ot1-import-that, #ot2-change-the-colours-back, #deleted-channel, #advent-of-code.

references a ``#deleted-channel``

kwzrd:
Since the outdated constant still hasn't been removed I will do it in #329.

PR #342, opened 2020-01-31
Check if channel exists before referencing it

neonsea:
---
name: Check if channel exists before referencing it
about: Avoid referencing #invalid-channel
closes: #335 

---

## Pull Request Details

Please ensure your PR fulfills the following criteria:

- [x] Have you joined the [PythonDiscord Community](https://pythondiscord.com/invite)?
- [x] Were your changes made in a Pipenv environment?
- [x] Does flake8 pass (```pipenv run lint```)

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -36,7 +36,6 @@ class Channels(NamedTuple):
     checkpoint_test = 422077681434099723
     devalerts = 460181980097675264
     devlog = int(environ.get("CHANNEL_DEVLOG", 548438471685963776))
-    devtest = 414574275865870337
     help_0 = 303906576991780866
     help_1 = 303906556754395136
     help_2 = 303906514266226689
@@ -142,7 +141,6 @@ STAFF_ROLES = Roles.helpers, Roles.moderator, Roles.admin, Roles.owner
 WHITELISTED_CHANNELS = (
     Channels.bot, Channels.seasonalbot_commands,
     Channels.off_topic_0, Channels.off_topic_1, Channels.off_topic_2,
-    Channels.devtest,
 )
 
 # Bot replies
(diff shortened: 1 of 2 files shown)


neonsea:
> The constant which provides the channel list should be changed here instead of checking it against the server every time
> https://github.com/python-discord/seasonalbot/blob/2679849160f160d8f70c597636798df185873523/bot/constants.py#L141-L146
> 
> 
> The invalid channel could also be removed from the `Channels` constants class

Aye, that was my initial reaction as well, however I opted to also check every time as an extra precaution as it seems like this would prevent the issue from popping up in the future (should the channel list change). I've removed the invalid channel \#devtest from both `Channels` and `WHITELISTED_CHANNELS`.

Numerlor:
Not sure what the consensus here is about checking it, because the channels don't really change much.
But if the check remains some sort of logging to notify about invalid channels so they can be removed could also prove to be useful.

kwzrd:
I think I like the idea of just removing the invalid channel (i.e. ba69450).

> ... it seems like this would prevent the issue from popping up in the future (should the channel list change) ...

The underlying problem is a misconfigured / outdated constant. The `#deleted-channel` showing up is a symptom, and dd6063f silences it rather than fixes it. If there exists a problem like this in the code base, it should be observable so that it can be fixed.

sco1:
> The underlying problem is a misconfigured / outdated constant. The #deleted-channel showing up is a symptom, and dd6063f silences it rather than fixes it. If there exists a problem like this in the code base, it should be observable so that it can be fixed.

We've come to an agreement on this, and rather than flagging this randomly when users bounce off of a whitelist, it would be better to solve this issue by validating the channel constants on bot startup & logging a warning if any of them fail to resolve so we can identify & update the problematic constant(s).

sco1:
Hi @neonsea, any interest in updating the implementation here?

neonsea:
I will implement functionality to check for invalid channels on startup and warn if there are any.

PR #347, opened 2020-02-09
Revert "Reddit command."

ikuyarihS:
Reverts python-discord/seasonalbot#346

The command is failing because of missing emojis - an action to add these emojis as well as changing the ids of the emojis in the code is needed before this PR is ready to be merged again.


AG4lyf:
hey can we have them in constants ? 
maybe we can use them later in future .

SebastiaanZ:
Instead of reverting this, we could also PR in the emojis required in a tidy manner.

We already have emojis for user`, upvotes, and comments icons because of our Reddit webhook, so I've added the other four (reddit logo + text, video, photo post) to our Emoji server.

All the IDs and emoji names:
```
# These exist on our main server (also used by webhook)
:user:              638729835442602003
:upvotes:           638729835245731840
:comments:          638729835073765387

# These now exist on PyDis Emojis II
:reddit_logo:       676030265734332427
:reddit_post_photo: 676030265734201344
:reddit_post_text:  676030265910493204
:reddit_post_video: 676030265839190047
```

SebastiaanZ:
I've changed emoji IDs with a commit to master so there's no need to revert the PR anymore. I think it would be a good idea to factor out these IDs to a constants file in the future. It's probably also a good idea to stylize the text, image, and video emojis to match the existing webhook emojis we're using. (I think the design is really good, but they could be slightly smaller to match the webhook emojis.)

PR #349, opened 2020-02-12
Bookmark reaction bookmark

### LINKED ISSUE #338: Blank .bm bookmarking request message
As an alternative from explicitly giving the message ID to bookmark when using the `.bm [message id]` command, add a shorthand version with the entire command being `.bm`. This will work by simply getting the message saying "`.bm`"'s ID and bookmarking that. For example:

```none
USER1 [m_id 334]: This is a cool post
USER2 [m_id 942]: .bm
```

In this example, instead of USER2 writing `.bm 334` to bookmark the message ID, they can simply type `.bm` and bookmark their own message (message ID of `942`). This can be useful for easier bookmarking with developer/debug mode disabled.

### LINKED ISSUE #341: Allow Users to React to a Bookmark to Recieve a Copy
Spun off of conversations in https://github.com/python-discord/seasonalbot/issues/338#issuecomment-578322144 and https://github.com/python-discord/seasonalbot/pull/340#pullrequestreview-348784550:

@kwzrd:
> I've also seen a suggestion that the bot could listen for further clicks on its 📨 reaction, and dispatch the bookmark to those who click as well. This would be a super clean way to ask to be given the bm as well, although it may not be very obvious that it does that.

@AG4lyf made a good initial implementation in #340 that exposed some UX concerns:

@sco1:
> Regarding the reaction by others to receive the same bookmark: the implementation details need to be discussed more before the feature is merged.
> 
> I don't think it makes sense from a UX perspective to utilize the same reaction for multiple purposes when one of those purposes silently disappears after a timeout. This is already barely discoverable functionality, and with this implementation a user that does know of its existence doesn't know if it has timed out or if it's broken. A second emoji solves half of the problem, which might be sufficient.

Since this implementation discussion shouldn't block the fix being implemented by #340, it has been removed from the scope of that PR.

Per discussions in Discord, the initial feeling is to provide a separate reaction to expose this functionality & keep it distinct from the bot's feedback that the DM has been sent. :pushpin: or :round_pushpin: have been suggested.

Also to be discussed is whether or not this should be implemented as either:
* A `wait_for` reaction loop, with a timeout that removes the reaction
* An `on_reaction_add` event handler with a persistent cache that links the bookmark invocation message and the target message.

The former is simpler to implement & is what I'm leaning towards. The latter is more generic but will be trickier to implement in a way that prevents the file from growing too large & limit vectors for trolling.

The reaction emoji and reaction handling approach should be decided on prior to work being done.

AG4lyf:
## Relevant Issues
Closes #341 , Closes #338 

## Description
it adds these features mainly :- 
1. Ability to react to a message that contain usage of bookmark command and get a copy of same bookmark
2. Ability to use blank bookmarks, i.e `.bm` will choose the message itself.

i tried to include it previously with a bug related update
but it needed some planning and improvement so we doing it now 
it basically create a temp list of the people who already got the bookmark , 
when someone react with pin emoji it check if user in that list if not it sends the 
bookmark to prevent spamming. and after 5 minutes it removes all the reaction.

## Reasoning
People requested this
Also it provide less keyboard typing to do same thing which can be done with a mouse click.
>>> This can be useful for easier bookmarking with developer/debug mode disabled.

## Screenshots
Now bookmarking from other will add a field , to remember the context of the message and reason to see what was happening when this message got bookmarked.
image

## Did you:


SebastiaanZ:
Uhm, I can't quite make out what this PR is about from the PR description. Could you maybe add a general description of the PR and a listing of the features it adds?

AG4lyf:
> Uhm, I can't quite make out what this PR is about from the PR description. Could you maybe add a general description of the PR and a listing of the features it adds?

I guess it's better now

gdude2002:
I think your `Closes` line is formatted incorrectly - you have the following:

> Closes #341 and #338

I think it should be:

> Closes #341, closes #338

bot/seasons/evergreen/bookmark.py
```diff
@@ -57,6 +60,30 @@ def __init__(self, bot: commands.Bot):
...
+            try:
+                reaction, user = await self.bot.wait_for("reaction_add", timeout=360.0, check=check)
+                if user != self.bot.user and (user not in sent_person):
+                    log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from {ctx.author}")
+                    await user.send(embed=embed)
+                    sent_person.append(user)
+            except asyncio.TimeoutError:
+                for reaction in ctx.message.reactions:
+                    async for user in reaction.users():
+                        await reaction.remove(user)
```
scragly: Don't iterate through all reactions, individually removing all single instances belonging to all users, it's a waste of API requests.

Change this to use [ctx.message.clear_reactions](https://discordpy.readthedocs.io/en/latest/api.html#discord.Message.clear_reactions) instead, as it's a single API request with the same result.

bot/seasons/evergreen/bookmark.py
```diff
@@ -57,6 +61,28 @@ def __init__(self, bot: commands.Bot):
...
+        )
+
+        def check(reaction: discord.Reaction, user: discord.User) -> bool:
+            return reaction.emoji == Emojis.pin
+
+        sent_person = [ctx.author]  # list of id who got the message
+        while True:
+            try:
+                reaction, user = await self.bot.wait_for("reaction_add", timeout=60.0, check=check)
+                if user != self.bot.user and user not in sent_person:
```
ikuyarihS: Since `Member` and `User` are both hashable, you can safely use a set here, as it will tremendously improve the membership check.

With that logic in mind, you can move all of them together into the `check` since it is also part of checking who reacted, like so:
```py
def check(reaction: discord.Reaction, user: discord.User) -> bool:
    return (
        user != self.bot.user and
        reaction.emoji == Emojis.pin and
        user not in sent_person
    )

sent_person = {ctx.author}
```

Then your loop will not have any extra indentation, simply
```py
while True:
    try:
        reaction, user = await self.bot.wait_for("reaction_add", timeout=60.0, check=check)
        log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from {ctx.author}")
        await user.send(embed=embed)
        sent_person.append(user)
    except asyncio.TimeoutError:
        ....
```

bot/seasons/evergreen/bookmark.py
```diff
@@ -57,6 +61,28 @@ def __init__(self, bot: commands.Bot):
...
+                    log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from {ctx.author}")
+                    await user.send(embed=embed)
+                    sent_person.append(user)
+            except asyncio.TimeoutError:
+                await ctx.message.clear_reactions()
+                return
```
ikuyarihS: It may be a rare case, but in case something else happen ( user is a bot spamming the reaction / bot cannot send DM to user ) you should include a broad except here, or better, do specific except for when the bot failed to send the DM, since it will effectively crash the entire loop.

bot/seasons/evergreen/bookmark.py
ikuyarihS: Also, there should also be a check where the message with the reaction is the same message as `ctx.message` - Otherwise the bot will simply check for reaction everywhere.

bot/seasons/evergreen/bookmark.py
scragly: Broad exceptions should be avoided. If there's another exception type to take into consideration, then it's better to add that specific one in.

In this case, the scope is a bit incorrect. You'd want to ignore those who have DMs disabled, rather than end the loop entirely and clear reactions. So the sending of the DM should be wrapped in it's own try except block catching a `discord.Forbidden`.

Good thought though to consider it; I didn't think of that case.

scragly: In fact, if the DM sent raises and it just ends up being `pass`ed then instead it's better to use [`contextlib.suppress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress):
```py
import contextlib
...
with contextlib.suppress(discord.Forbidden):
    await user.send(embed=embed)
sent_person.append(user)
```

bot/seasons/evergreen/bookmark.py
ikuyarihS: Please don't auto resolve, you can certainly ping me in the discord channel for discussion / a review over it as well.

As it stands, you can certainly change `sent_person` from being a list to a set since `Member` and `User` are hashable, do you have plans for that as well?

AG4lyf: no , list are way better and make it easy to read , so I am gonna keep that as list only.

ikuyarihS: You are not reading it anywhere at all, you only keep it to do a membership checking - avoid sending to the same person twice, is that correct?

AG4lyf: We are reading it in the check ,also i feel like someone in future ever read this code , he is gonna get confused with sets.

ikuyarihS: If that is a case, I strongly recommend going from a list to a set, as set has a constant, close to O(1) for membership checking as opposed to a list with a O(n) to do the same thing - unless we are doing something else with the list.

bot/seasons/evergreen/bookmark.py
SebastiaanZ: Why was this conversation resolved? I don't see any changes to code.

bot/seasons/evergreen/bookmark.py
SebastiaanZ: I doubt anyone will get confused by sets; they read just as easy as lists and make absolute sense in this context.

AG4lyf: Commited

AG4lyf:
Waiting for reviews ;-;

bot/seasons/evergreen/bookmark.py
```diff
@@ -50,13 +55,40 @@ def __init__(self, bot: commands.Bot):
...
+        while True:
+            try:
+                reaction, user = await self.bot.wait_for("reaction_add", timeout=60.0, check=check)
+                log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from '{ctx.author}'.")
+                with contextlib.suppress(discord.Forbidden):
+                    await user.send(embed=embed)
+                sent_person.add(user)
+            except asyncio.TimeoutError:
+                await ctx.message.clear_reactions()
+                return
```
ikuyarihS: Since only `wait_for()` can raise a `asyncio.TimeoutError` in this block of code, it is much better to only put that in the try block, and everything else outside to reduce the level of indentation / make it clear what to do, something like this
```py
try:
    reaction, user = await self.bot.wait_for("reaction_add", timeout=60.0, check=check)
except asyncio.TimeoutError:
    # handle timeout

log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from '{ctx.author}'.")
...
```
Also, as a quality of life, in case the bot cannot send the message to the user, remove the user's reaction and send a self-deleting message mentioning them that they need to open their DM - it will be more user-friendly than silently dismiss the error and let the user wonder why they did not receive any DM, similar to what you did in line 55.

AG4lyf: but if the chat proceed and someone randomly comes and put the reaction
it will feel weird a random error message out of nowhere.

ikuyarihS: The bot will only send such a message if it fails to DM, and you can send a message with `delete_after` and so it will not swarm the chat at one bit.
A message like "@ someone you tried to bookmark this bookmark, but your DM is blocked, reopen it and try again" or something similar will be clear enough in my opinion.

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,18 +49,48 @@ def __init__(self, bot: commands.Bot):
...
+                and reaction.message == ctx.message
+                and user not in sent_person
+            )
+
+        sent_person = {ctx.author}  # set of id who got the message
+        while True:
+            reaction, user = await self.bot.wait_for("reaction_add", timeout=10.0, check=check)
+            log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from '{ctx.author}'.")
+            try:
+                await user.send(embed=embed)
```
ikuyarihS: `wait_for` is the one that will raise `asyncio.TimeoutError` and not `await user.send(embed=embed)` while `user.send()` will raise `discord.Forbidden` - so timeout will effectively raise an error.

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,18 +49,48 @@ def __init__(self, bot: commands.Bot):
...
+            except asyncio.TimeoutError:
+                await ctx.message.clear_reactions()
+                return
+            except discord.Forbidden:  # Mention so that we can differentiate who it was for
+                await ctx.send(user.mention, embed=error_embed, delete_after=7.5)
```
ikuyarihS: Sending an embed along with a single mention will waste a lot of space in the channel, maybe just the mention and a short text suggesting them to open the DM.

Also consider remove their reaction here as well, just the reaction said user added.

bot/seasons/evergreen/bookmark.py
AG4lyf: done

bot/seasons/evergreen/bookmark.py
AG4lyf: Yeah forgot about removing the reaction tho , done in this commit

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,19 +49,54 @@ def __init__(self, bot: commands.Bot):
...
+
+        def check(reaction: discord.Reaction, user: discord.User) -> bool:
+            return (
+                user != self.bot.user
+                and reaction.emoji == Emojis.pin
+                and reaction.message == ctx.message
+                and user not in sent_person
+            )
+
+        sent_person = {ctx.author}  # set of id who got the message
```
ikuyarihS: There is a rare case when `ctx.author` first disabled their DM, but the bot still accepts reaction to add bookmarks for others, the `ctx.author` cannot do it even after turning on their DM because by default the set `sent_person` already have them. There should be a check here if the first DM went through or not.

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,19 +49,54 @@ def __init__(self, bot: commands.Bot):
...
+
         try:
             await ctx.author.send(embed=embed)
         except discord.Forbidden:
-            error_embed = discord.Embed(
-                title=random.choice(ERROR_REPLIES),
-                description=f"{ctx.author.mention}, please enable your DMs to receive the bookmark",
-                colour=Colours.soft_red
-            )
             await ctx.send(embed=error_embed)
```
ikuyarihS: This should be self-deleting as well to avoid clogging the channel. Maybe change this to a normal message as well so the bot can mention the `ctx.author`
This can be done in a similar manner in line 95.

AG4lyf: It is been talked about and the decision was made to follow the theme of error message.

bot/seasons/evergreen/bookmark.py
AG4lyf: I did that on purpose , As you can see it don't add the envelope reaction but only the pin emoji so that other can get the message it self even if it failed . cuz why not ?

AG4lyf: and yeah i am making some changes so that person now not in that set.

ikuyarihS: Be careful when you handle it, if it has already bookmarked for the `ctx.author` you do not want that author to react and got another copy of the bookmark.

AG4lyf: Done :)

bot/seasons/evergreen/bookmark.py
AG4lyf: Forgot to mention . added self deletion after 7.5 seconds. same as in error_handler

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,18 +50,61 @@ def __init__(self, bot: commands.Bot):
...
+            log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from '{ctx.author}'.")
+
+            try:
+                if user == ctx.author:
+                    await user.send(embed=embed)
+                else:
+                    await user.send(embed=copied_embed)
+
+            except discord.Forbidden:  # Mention so that we can differentiate who it was for
+                await ctx.send(f"{user.mention} Please enable you're DM to receive the message.", delete_after=7.5)
```
ikuyarihS:
```suggestion
                await ctx.send(f"{user.mention} Please enable your DM to receive the message.", delete_after=7.5)
```
Minor grammar mistake

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,18 +50,61 @@ def __init__(self, bot: commands.Bot):
...
+
+            log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from '{ctx.author}'.")
+
+            try:
+                if user == ctx.author:
+                    await user.send(embed=embed)
+                else:
+                    await user.send(embed=copied_embed)
+
+            except discord.Forbidden:  # Mention so that we can differentiate who it was for
```
ikuyarihS: The comment here should be put above the `await ctx.send` instead, since it is for the send, not for the exception.
For the exception, you can put a comment similar to 
```py
# if failed to DM the user, most likely because they are blocking DM
```

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,18 +50,61 @@ def __init__(self, bot: commands.Bot):
...
+                await ctx.send(f"{user.mention} Please enable you're DM to receive the message.", delete_after=7.5)
+                await reaction.remove(user)
```
ikuyarihS: This can raise 3 exceptions:

Raises:
- HTTPException – Removing the reaction failed.
- Forbidden – You do not have the proper permissions to remove the reaction.
- NotFound – The user you specified, or the reaction’s message was not found.
You can read more about it here - [discord.Reaction.remove](https://discordpy.readthedocs.io/en/latest/api.html#discord.Reaction.remove)

While this can be rare, if this happens just once the entire loop will end, leaving all reactions on the message and may confuse others ( "I reacted to it, but it did not send me any bookmark" ) because they cannot see the log, so consider try except this part as well.

ikuyarihS: As I've asked, please do not automatically resolve the conversation.
As of right now, you are adding more exception to the try, while the line that can cause exception is inside your `except discord.Forbidden` block, adding more `except` outside will **not** catch the exception of `reaction.remove(user)`

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,18 +50,65 @@ def __init__(self, bot: commands.Bot):
...
+
+            try:
+                if user == ctx.author:
+                    await user.send(embed=embed)
+                else:
+                    await user.send(embed=copied_embed)
+
+            except discord.Forbidden:
+                await ctx.send(f"{user.mention} Please enable your DM to receive the message.", delete_after=7.5)
+                await reaction.remove(user)
```
ks129:
```suggestion
                try:
                    await reaction.remove(user)
                except discord.Forbidden:
                    await ctx.send("I don't have permissions to remove reaction.")
                except discord.HTTPException:
                    await ctx.send("Unknown error while removing reaction.")
                except discord.NotFound:
                    await ctx.send("Message or reaction author not found, can't remove reaction.")

```

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,18 +50,65 @@ def __init__(self, bot: commands.Bot):
...
+            except discord.HTTPException:
+                pass
+            except discord.NotFound:
+                pass
```
ks129:
```suggestion
            except discord.HTTPException:
                await ctx.send("Unknown error while sending bookmark to user.")

```

bot/seasons/evergreen/bookmark.py
```diff
@@ -1,65 +1,124 @@
...
+                    await reaction.remove(user)
+                except discord.Forbidden:
+                    await ctx.send("I don't have permissions to remove reaction.")
+                except discord.HTTPException:
+                    await ctx.send("Unknown error while removing reaction.")
+                except discord.NotFound:
+                    await ctx.send("Message or reaction author not found, can't remove reaction.")
+
+            except discord.HTTPException:
+                await ctx.send("Unknown error while sending bookmark to user.")
```
ikuyarihS: This looks better now, and it won't crash the loop.
But should these messages be automatically deleted after a delay? Or should they be simple logging only?

ks129:
```suggestion
                try:
                    await reaction.remove(user)
                except discord.Forbidden:
                    await ctx.send("I don't have permissions to remove reaction.", delete_after=7.5)
                except discord.HTTPException:
                    await ctx.send("Unknown error while removing reaction.", delete_after=7.5)
                except discord.NotFound:
                    await ctx.send("Message or reaction author not found, can't remove reaction.", delete_after=7.5)

            except discord.HTTPException:
                await ctx.send("Unknown error while sending bookmark to user.", delete_after=7.5)

```
I think it's better to delete this. Delay may different, but just I think user need too some information. When this only log, there is no response by bot.

AG4lyf:
I am well aware that my lint is failing . I am gonna debug it right now . as i now got some time after whole this time.

AG4lyf:
Since this PR is such a mess now and so much outdated , i am closing this and following it up with a new PR cuz i can't fork a new until this fork is deleted.

Issue #348, opened 2020-02-09
Problems with Reddit command.


RohanJnr:
Recently, the Reddit command PR (#346) was merged with a number of problems being overlooked. This issue is to outline the required fixes and any additional enhancements to the feature.
## Current Problems

- When navigating using reactions, the embed is edited twice in succession, first making the embed blank and then with the new content. This sometimes results in slow updating and the embed staying blank.
![here](https://cdn.discordapp.com/attachments/542272993192050698/676055841220067338/Screenshot_20200209-131812.jpg)

- The user-agent in the header dict is currently set to `Iceman` but in fact it should be set the app/client that is sending the request.
So it needs to be changed to : python3:python-discord/seasonalbot:1.0.0 (by /u/PythonDiscord)

- Currently, the API does not use an authenticated request. We should make sure to use an app token from Reddit the same as the Python Bot does also. This bot will need one created by an admin specific to Seasonalbot.
Use the PyDis Bot's Reddit cog implementation [here](https://github.com/python-discord/bot/blob/master/bot/cogs/reddit.py) as reference.

- The word `meant` is misspelt as `ment`([on line 60](https://github.com/python-discord/seasonalbot/blob/master/bot/seasons/evergreen/reddit.py#L60)).

Suhail6inkling:
also - this is a minor error but

* “meant” is misspelt incorrectly ([shown on L60](https://github.com/python-discord/seasonalbot/blob/master/bot/seasons/evergreen/reddit.py#L60))

scragly:
Definitely worth mentioning, thanks @Suhail6inkling.

AG4lyf:
Bot doesn't check if the reddit have any post in it.
Such reddit can be `/r/makemyideas`
Bot fails silently 
I found this on private testing 
```
03/01/20 09:29:29 - bot.seasons.evergreen.error_handler ERROR: Unhandled command error: list index out of range
Traceback (most recent call last):
  File "/home/ag/.local/share/virtualenvs/seasonalbot-0Ba-uPuk/lib/python3.8/site-packages/discord/ext/commands/core.py", line 79, in wrapped
    ret = await coro(*args, **kwargs)
  File "/home/ag/PycharmProjects/bookmark-re-discord/seasonalbot/bot/seasons/evergreen/reddit.py", line 58, in get_reddit
    if posts[1]["data"]["over_18"] is True:
IndexError: list index out of range
```
It happened today 
[...]

PR #353, opened 2020-02-16
Fix the Pagination cog

### LINKED ISSUE #352: pagination does not work when clicking LAST_EMOJI
## Description
Right now, when using the pagination, clicking the LAST_EMOJI will raise an exception

```py
  File "/bot/bot/pagination.py", line 392, in paginate
    current_page = len(paginator.pages - 1)
TypeError: unsupported operand type(s) for -: 'list' and 'int'
```

## Proposed Implementation
This is because `paginator.pages` is a list. I believe it should be
```py
current_page = len(paginator.pages) - 1
```

This line of code can be found at [bot/paginator.py](https://github.com/python-discord/seasonalbot/blob/7ef23b4d42716883dd8d51d0b2ffab36e45b6f59/bot/pagination.py#L392)

F4zii:
When using pagination in the Reddit cog, clicking the LAST_EMOJI would
raise an exception like this:
TypeError: unsupported operand type(s) for -: 'list' and 'int'

This was resolved by taking the subtraction out of the len() function.

## Additional Changes
Swapping to the trash can emoji with this PR to align with the changes made to the bot.
https://github.com/python-discord/bot/pull/625

## Relevant Issues
Closes #352 

## Did you:


F4zii:
@sco1 Migrated, waiting for your approval

bot/pagination.py
```diff
@@ -261,8 +261,8 @@ def event_check(reaction_: Reaction, user_: Member) -> bool:
 
                 await message.edit(embed=embed)
 
-        log.debug("Ending pagination and removing all reactions...")
-        await message.clear_reactions()
+        log.debug("Ending pagination and deleting the message")
+        await message.delete()
```
ikuyarihS: When `asyncio.TimeoutError` happened ( No more reaction from user in set amount of time ) then the default behavior will be to keep the result but remove all the reactions, if you change it to delete, the result from the query will also be gone.

In this case, we want to keep the message and only clear all of its reactions.

bot/pagination.py
```diff
@@ -424,5 +424,5 @@ def check_event(reaction_: Reaction, member: Member) -> bool:
 
             await message.edit(embed=embed)
 
-        log.debug("Ending pagination and removing all reactions...")
-        await message.clear_reactions()
+        log.debug("Ending pagination and deleting the message")
+        await message.delete()
```
ikuyarihS: Same with the above, we want to keep the message and only clear the reactions.

bot/pagination.py
```diff
@@ -10,7 +10,7 @@
 LEFT_EMOJI = "\u2B05"    # [:arrow_left:]
 RIGHT_EMOJI = "\u27A1"   # [:arrow_right:]
 LAST_EMOJI = "\u23ED"    # [:track_next:]
-DELETE_EMOJI = "\u274c"  # [:x:]
+DELETE_EMOJI = "<:trashcan:637136429717389331>"  # [:trashcan:]
 
 PAGINATION_EMOJI = [FIRST_EMOJI, LEFT_EMOJI, RIGHT_EMOJI, LAST_EMOJI, DELETE_EMOJI]
```
ikuyarihS: A minor changes & considerations. Otherwise the functionality works well.
This can be changed from a list to a tuple instead.

F4zii: Sure, will do. I just saw that this is the implementation at https://github.com/python-discord/bot
Should I also apply this migration in the /python-discord/bot?

ikuyarihS: Should, yes, but doing a whole PR just to change a line, I don't think it's worth it, unless you want to refactor it, then yes, please do!

F4zii: Done, waiting for your approval on both bot repos 👍

F4zii:
@sco1 Is everything ready or there are still changes needed?

bot/pagination.py
```diff
@@ -10,9 +10,9 @@
 LEFT_EMOJI = "\u2B05"    # [:arrow_left:]
 RIGHT_EMOJI = "\u27A1"   # [:arrow_right:]
 LAST_EMOJI = "\u23ED"    # [:track_next:]
-DELETE_EMOJI = "\u274c"  # [:x:]
+DELETE_EMOJI = "<:trashcan:637136429717389331>"  # [:trashcan:]
```
ikuyarihS: The PR is good to be merged, but before that, let's move this to `constants.py` so we can easily change it later and will make sure that the trashcan used elsewhere will be the same as this one.

F4zii: Yeah, as I said, I will add it to the constants module.
Will it be a good idea to try to clear things out and try to solve #348?

PR #356, opened 2020-02-23
Icon switch frequency

kwzrd:
Adds a configurable constant, which determines the amount of days to wait between cycling icons within a single season. See my conversation with lemon: https://github.com/python-discord/organisation/issues/191#issuecomment-590086579 and below.

@lemonsaurus I liked your suggestion with using `itertools.cycle`, but I decided to just keep internal state as an int. The reason here is that we can easily reset it back to 0 if we switch seasons.

I tried to keep the implementation as simple as possible. It can be tested by manually setting `sleep_time` to a few seconds. Running this loop with a `sleep_time` forced to 3 seconds and logging set to `DEBUG` produces the following output.

```
02/23/20 20:03:41 - bot.seasons.season DEBUG: Days since last icon change: 1
02/23/20 20:03:41 - bot.seasons.season DEBUG: Waiting 2 more days to cycle icon
02/23/20 20:03:44 - bot.seasons.season DEBUG: Days since last icon change: 2
02/23/20 20:03:44 - bot.seasons.season DEBUG: Waiting 1 more days to cycle icon
02/23/20 20:03:47 - bot.seasons.season DEBUG: Days since last icon change: 3
02/23/20 20:03:50 - bot.seasons.season DEBUG: Days since last icon change: 1
02/23/20 20:03:50 - bot.seasons.season DEBUG: Waiting 2 more days to cycle icon
02/23/20 20:03:53 - bot.seasons.season DEBUG: Days since last icon change: 2
02/23/20 20:03:53 - bot.seasons.season DEBUG: Waiting 1 more days to cycle icon
02/23/20 20:03:56 - bot.seasons.season DEBUG: Days since last icon change: 3
02/23/20 20:03:59 - bot.seasons.season DEBUG: Days since last icon change: 1
02/23/20 20:03:59 - bot.seasons.season DEBUG: Waiting 2 more days to cycle icon
```

## Considerations
Have I chosen a good place for the constant? Should it be configurable via an environmental var, like the other constants in `Client`?

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -69,6 +69,7 @@ class Client(NamedTuple):
     token = environ.get("SEASONALBOT_TOKEN")
     debug = environ.get("SEASONALBOT_DEBUG", "").lower() == "true"
     season_override = environ.get("SEASON_OVERRIDE")
+    icon_cycle_frequency = 3  # N days to wait between cycling server icons within a single season
 
 
 class Colours:
--- a/bot/seasons/season.py
+++ b/bot/seasons/season.py
@@ -386,7 +386,7 @@ class SeasonManager(commands.Cog):
 
         while True:
             await asyncio.sleep(self.sleep_time)  # Sleep until midnight
-            self.sleep_time = 86400  # Next time, sleep for 24 hours.
+            self.sleep_time = 24 * 3600  # Next time, sleep for 24 hours
 
             # If the season has changed, load it.
             new_season = get_season(date=datetime.datetime.utcnow())
--- a/bot/seasons/season.py
+++ b/bot/seasons/season.py
@@ -383,18 +383,29 @@ class SeasonManager(commands.Cog):
         """Asynchronous timer loop to check for a new season every midnight."""
         await self.bot.wait_until_ready()
         await self.season.load()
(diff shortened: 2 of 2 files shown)


sco1:
Do we care that this counter is going to be reset every time the bot restarts?

kwzrd:
@sco1 I don't think it's a tragedy. The bot already doesn't persist state across sessions w.r.t. the `index` variable, which controls which icon to choose when cycling. So the cycle is reset to `index=0` on every restart. I'm not trying to use this as an excuse, of course, ideally we'd do something smarter for that as well, but I see this fix as a stop-gap solution until #329 is ready.

A logical solution would be taking current day in month, and moduloing it by the configured delay:
```python
>>> delay = 3
>>>
>>> for day in range(1, 31):
...     print(day, "cycle" if day % delay == 0 else "wait")
...
1 wait
2 wait
3 cycle
4 wait
5 wait
6 cycle
7 wait
8 wait
9 cycle
...
```
I don't think this really improves the situation that much without also changing the way icons are drawn though. That should probably apply a similar mechanism. Or we could always just draw randomly. I'm open to suggestions.

PR #361, opened 2020-02-24
Added .games command

### LINKED ISSUE #278: [Feature Request] Horror Game Recommendations
It's that spooky time of the year, and there's no better time to turn off the lights and cower in fear in front of a screen! To also keep up with the gaming spirit, what game should you play? Ask Spookybot!

## Implementation details
- [ ] Return an embed containing information on a random popular horror game

## Information in the embed
- Summary of the game
- Developer of the game
- Publisher of the game
- Rating
- Release Date
- IGN score

### LINKED ISSUE #354: [Feature Request] .games Command
## Description

There should `.games` command where you must provide genre to get random games from this genre (like my added `.movies` command) and also have following subcommands:
- `top <amount>` - Most popular games
- `company` - Get random game company

## Reasoning

I'd like to see this in SeasonalBot, and there is already something like this in Halloween Season, so this mean other peoples want this too.

ks129:
## Relevant Issues

Closes #354
Closes #278 

## Description

I added `.games <genre> <amount (default 5)>` command + all it's subcommands:
- `.games top <amount, default 10>` - Get current top games
- `.games search <search string>` - Search games by title
- `.games genres` - Get all current available genres
- `.games companies <amount default 5>` - Get random game companies

I used IGDB API for this.

## Reasoning

I added this due I lately added `.movies` command and current game recommendation command is pretty limited.

## Screenshots

Screenshot 2020-02-24 at 18 59 48
Screenshot 2020-02-24 at 19 00 11
Screenshot 2020-02-24 at 19 00 43
Screenshot 2020-02-24 at 19 00 57

## Did you:

### DIFF
--- a/bot/seasons/evergreen/game.py
+++ b/bot/seasons/evergreen/game.py
@@ -1,10 +1,8 @@
 import difflib
 import logging
 import random
-import textwrap
 from datetime import datetime as dt
 from enum import IntEnum
-from string import Template
 from typing import Any, Dict, List, Optional, Tuple
 
 from aiohttp import ClientSession
@@ -33,65 +31,53 @@ logger = logging.getLogger(__name__)
 
 # Body templates
 # Request body template for get_games_list
-GAMES_LIST_BODY = Template(
-    textwrap.dedent("""
-        fields cover.image_id, first_release_date, total_rating, name, storyline, url, platforms.name, status,
-        involved_companies.company.name, summary, age_ratings.category, age_ratings.rating, total_rating_count;
-        ${sort} ${limit} ${offset} ${genre} ${additional}
-    """)
+GAMES_LIST_BODY = (
+    "fields cover.image_id, first_release_date, total_rating, name, storyline, url, platforms.name, status,"
+    "involved_companies.company.name, summary, age_ratings.category, age_ratings.rating, total_rating_count;"
+    "{sort} {limit} {offset} {genre} {additional}"
 )
 
 # Request body template for get_companies_list
-COMPANIES_LIST_BODY = Template(
-    textwrap.dedent("""
-        fields name, url, start_date, logo.image_id, developed.name, published.name, description;
-        offset ${offset};
-        limit ${limit};
-    """)
+COMPANIES_LIST_BODY = (
+    "fields name, url, start_date, logo.image_id, developed.name, published.name, description;"
+    "offset {offset}; limit {limit};"
 )
 
 # Request body template for games search
-SEARCH_BODY = Template('fields name, url, storyline, total_rating, total_rating_count; limit 50; search "${term}";')
+SEARCH_BODY = 'fields name, url, storyline, total_rating, total_rating_count; limit 50; search "{term}";'
 
 # Pages templates
 # Game embed layout
-GAME_PAGE = Template(
-    textwrap.dedent("""
-        **[${name}](${url})**
-        ${description}
-        **Release Date:** ${release_date}
-        **Rating:** ${rating}/100 :star: (based on ${rating_count} ratings)
-        **Platforms:** ${platforms}
-        **Status:** ${status}
-        **Age Ratings:** ${age_ratings}
-        **Made by:** ${made_by}
-
-        ${storyline}
-    """)
+GAME_PAGE = (
+    "**[{name}]({url})**\n"
+    "{description}"
+    "**Release Date:** {release_date}\n"
+    "**Rating:** {rating}/100 :star: (based on {rating_count} ratings)\n"
+    "**Platforms:** {platforms}\n"
+    "**Status:** {status}\n"
+    "**Age Ratings:** {age_ratings}\n"
+    "**Made by:** {made_by}\n\n"
+    "{storyline}"
 )
 
 # .games company command page layout
-COMPANY_PAGE = Template(
-    textwrap.dedent("""
-        **[${name}](${url})**
-        ${description}
-        **Founded:** ${founded}
-        **Developed:** ${developed}
-        **Published:** ${published}
-    """)
+COMPANY_PAGE = (
+    "**[{name}]({url})**\n"
+    "{description}"
+    "**Founded:** {founded}\n"
+    "**Developed:** {developed}\n"
+    "**Published:** {published}"
 )
 
 # For .games search command line layout
-GAME_SEARCH_LINE = Template(
-    textwrap.dedent("""
-        **[${name}](${url})**
-        ${rating}/100 :star: (based on ${rating_count} ratings)
-    """)
+GAME_SEARCH_LINE = (
+    "**[{name}]({url})**\n"
+    "{rating}/100 :star: (based on {rating_count} ratings)"
 )
 
 # URL templates
-COVER_URL = Template("https://images.igdb.com/igdb/image/upload/t_cover_big/${image_id}.jpg")
-LOGO_URL = Template("https://images.igdb.com/igdb/image/upload/t_logo_med/${image_id}.png")
+COVER_URL = "https://images.igdb.com/igdb/image/upload/t_cover_big/{image_id}.jpg"
+LOGO_URL = "https://images.igdb.com/igdb/image/upload/t_logo_med/{image_id}.png"
 
 # Create aliases for complex genre names
 ALIASES = {
@@ -298,7 +284,7 @@ class Games(Cog):
             "genre": f"where genres = ({genre});" if genre else "",
             "additional": additional_body
         }
-        body = GAMES_LIST_BODY.substitute(params)
+        body = GAMES_LIST_BODY.format(**params)
 
         # Do request to IGDB API, create headers, URL, define body, return result
         async with self.http_session.get(url=f"{BASE_URL}/games", data=body, headers=HEADERS) as resp:
@@ -307,7 +293,7 @@ class Games(Cog):
     async def create_page(self, data: Dict[str, Any]) -> Tuple[str, str]:
         """Create content of Game Page."""
         # Create cover image URL from template
-        url = COVER_URL.substitute({"image_id": data["cover"]["image_id"] if "cover" in data else ""})
+        url = COVER_URL.format(**{"image_id": data["cover"]["image_id"] if "cover" in data else ""})
 
         # Get release date separately with checking
         release_date = dt.utcfromtimestamp(data["first_release_date"]).date() if "first_release_date" in data else "?"
@@ -332,7 +318,7 @@ class Games(Cog):
             "made_by": ", ".join(companies),
             "storyline": data["storyline"] if "storyline" in data else ""
         }
-        page = GAME_PAGE.substitute(formatting)
+        page = GAME_PAGE.format(**formatting)
 
         return page, url
 
@@ -341,7 +327,7 @@ class Games(Cog):
         lines = []
 
         # Define request body of IGDB API request and do request
-        body = SEARCH_BODY.substitute({"term": search_term})
+        body = SEARCH_BODY.format(**{"term": search_term})
 
         async with self.http_session.get(url=f"{BASE_URL}/games", data=body, headers=HEADERS) as resp:
             data = await resp.json()
@@ -354,7 +340,7 @@ class Games(Cog):
                 "rating": round(game["total_rating"] if "total_rating" in game else 0, 2),
                 "rating_count": game["total_rating_count"] if "total_rating" in game else "?"
             }
-            line = GAME_SEARCH_LINE.substitute(formatting)
+            line = GAME_SEARCH_LINE.format(**formatting)
             lines.append(line)
 
         return lines
@@ -367,7 +353,7 @@ class Games(Cog):
         returning results.
         """
         # Create request body from template
-        body = COMPANIES_LIST_BODY.substitute({
+        body = COMPANIES_LIST_BODY.format(**{
             "limit": limit,
             "offset": offset
         })
@@ -378,7 +364,7 @@ class Games(Cog):
     async def create_company_page(self, data: Dict[str, Any]) -> Tuple[str, str]:
         """Create good formatted Game Company page."""
         # Generate URL of company logo
-        url = LOGO_URL.substitute({"image_id": data["logo"]["image_id"] if "logo" in data else ""})
+        url = LOGO_URL.format(**{"image_id": data["logo"]["image_id"] if "logo" in data else ""})
 
         # Try to get found date of company
         founded = dt.utcfromtimestamp(data["start_date"]).date() if "start_date" in data else "?"
@@ -395,7 +381,7 @@ class Games(Cog):
             "developed": developed,
             "published": published
         }
-        page = COMPANY_PAGE.substitute(formatting)
+        page = COMPANY_PAGE.format(**formatting)
 
         return page, url
 
--- a/bot/seasons/evergreen/game.py
+++ b/bot/seasons/evergreen/game.py
@@ -1,23 +1,102 @@
-import asyncio
+import difflib
+import logging
 import random
-from datetime import datetime
+import textwrap
+from datetime import datetime as dt
 from enum import IntEnum
-from typing import Any, Dict, List, Tuple
+from string import Template
+from typing import Any, Dict, List, Optional, Tuple
 
 from aiohttp import ClientSession
 from discord import Embed
-from discord.ext.commands import Bot, Cog, Context, group
+from discord.ext.commands import Cog, Context, group
 
+from bot.bot import SeasonalBot
 from bot.constants import Tokens
 from bot.pagination import ImagePaginator, LinePaginator
 
 # Base URL of IGDB API
-BASE_URL = "https://api-v3.igdb.com/"
-IMAGE_BASE_URL = "https://images.igdb.com/igdb/image/upload/"
+BASE_URL = "https://api-v3.igdb.com"
 
 HEADERS = {
     "user-key": Tokens.igdb,
     "Accept": "application/json"
+}
+
+logger = logging.getLogger(__name__)
+
+# ---------
+# TEMPLATES
+# ---------
+
+# Body templates
+# Request body template for get_games_list
+GAMES_LIST_BODY = Template(
+    textwrap.dedent("""
+        fields cover.image_id, first_release_date, total_rating, name, storyline, url, platforms.name, status,
+        involved_companies.company.name, summary, age_ratings.category, age_ratings.rating, total_rating_count;
+        ${sort} ${limit} ${offset} ${genre} ${additional}
+    """)
+)
+
+# Request body template for get_companies_list
+COMPANIES_LIST_BODY = Template(
+    textwrap.dedent("""
+        fields name, url, start_date, logo.image_id, developed.name, published.name, description;
+        offset ${offset};
+        limit ${limit};
+    """)
+)
+
+# Request body template for games search
+SEARCH_BODY = Template('fields name, url, storyline, total_rating, total_rating_count; limit 50; search "${term}";')
+
+# Pages templates
+# Game embed layout
+GAME_PAGE = Template(
+    textwrap.dedent("""
+        **[${name}](${url})**
+        ${description}
+        **Release Date:** ${release_date}
+        **Rating:** ${rating}/100 :star: (based on ${rating_count} ratings)
+        **Platforms:** ${platforms}
+        **Status:** ${status}
+        **Age Ratings:** ${age_ratings}
+        **Made by:** ${made_by}
+
+        ${storyline}
+    """)
+)
+
+# .games company command page layout
+COMPANY_PAGE = Template(
+    textwrap.dedent("""
+        **[${name}](${url})**
+        ${description}
+        **Founded:** ${founded}
+        **Developed:** ${developed}
+        **Published:** ${published}
+    """)
+)
+
+# For .games search command line layout
+GAME_SEARCH_LINE = Template(
+    textwrap.dedent("""
+        **[${name}](${url})**
+        ${rating}/100 :star: (based on ${rating_count} ratings)
+    """)
+)
+
+# URL templates
+COVER_URL = Template("https://images.igdb.com/igdb/image/upload/t_cover_big/${image_id}.jpg")
+LOGO_URL = Template("https://images.igdb.com/igdb/image/upload/t_logo_med/${image_id}.png")
+
+# Create aliases for complex genre names
+ALIASES = {
+    "Role-playing (rpg)": ["Role-playing", "Rpg"],
+    "Turn-based strategy (tbs)": ["Turn-based-strategy", "Tbs"],
+    "Real time strategy (rts)": ["Real-time-strategy", "Rts"],
+    "Hack and slash/beat 'em up": ["Hack-and-slash"]
 }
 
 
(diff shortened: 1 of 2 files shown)


bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,333 @@
...
+            pages.append(await self.create_page(game))
+        return pages
+
+    async def create_page(self, data: Dict[str, Any]) -> Tuple[str, str]:
+        """Create content of Game Page."""
+        # Create page content variable, what will be returned
+        page = ""
+
+        # Get keys to one variable. Used for code style
+        keys = data.keys()
```
tagptroll1: Why is this needed?  The use of it seems inconsistent with for instance L236 and L224. Might as well use data in those places?

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,333 @@
...
+            data = await resp.json()
+
+        # Loop over games, format them to good format, make line and append this to total lines
+        for game in data:
+            line = ""
+
+            # Add games name and rating, also attach URL to title
+            line += f"**[{game['name']}]({game['url']})**\n"
+            line += f"""{'{0:.2f}'.format(game['total_rating'] if 'total_rating' in game.keys() else 0)}/100 :star: (by {
+            game['total_rating_count'] if 'total_rating_count' in game.keys() else '?'} users)"""
```
tagptroll1:
```suggestion
            game['total_rating_count'] if 'total_rating_count' in game else '?'} users)"""
```

tagptroll1: You get the idea

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,333 @@
...
+
+    async def get_company_pages(self, data: List[Dict[str, Any]]) -> List[Tuple[str, str]]:
+        """Get all good formatted pages about Game Companies."""
+        pages = []
+
+        # Loop over companies, add them to pages listing and return pages
+        for co in data:
+            pages.append(await self.create_company_page(co))
+
+        return pages
```
tagptroll1: Couldn't this be a list comprehension?

bot/seasons/evergreen/game.py
```diff
@@ -198,42 +208,38 @@ def __init__(self, bot: Bot):
     async def get_pages(self, data: List[Dict[str, Any]]) -> List[Tuple[str, str]]:
         """Generate all game pages, do additional requests to IGDB API."""
         pages = []
-        for game in data:
-            pages.append(await self.create_page(game))
+        [pages.append(await self.create_page(game)) for game in data]
```
tagptroll1: What about 
```suggestion
        pages = [await self.create_page(game) for game in data]
```
Doesn't that work?

bot/seasons/evergreen/game.py
```diff
@@ -198,42 +208,38 @@ def __init__(self, bot: Bot):
     async def get_pages(self, data: List[Dict[str, Any]]) -> List[Tuple[str, str]]:
         """Generate all game pages, do additional requests to IGDB API."""
         pages = []
```
tagptroll1:
```suggestion
```
with suggestion below

bot/seasons/evergreen/game.py
```diff
@@ -296,8 +302,7 @@ def __init__(self, bot: Bot):
...
 
         # Loop over companies, add them to pages listing and return pages
-        for co in data:
-            pages.append(await self.create_company_page(co))
+        [pages.append(await self.create_company_page(co)) for co in data]
```
tagptroll1: See comment on L211

bot/seasons/evergreen/game.py
```diff
@@ -198,42 +208,38 @@ def __init__(self, bot: Bot):
...
         page += f"**Platforms:** "
-        page += f"{', '.join(pf['name'] for pf in data['platforms']) if 'platforms' in keys else '?'}\n"
+        page += f"{', '.join(pf['name'] for pf in data['platforms']) if 'platforms' in data else '?'}\n"
 
-        page += f"**Status:** {GameStatus(data['status']).name if 'status' in keys else '?'}\n"
+        page += f"**Status:** {GameStatus(data['status']).name if 'status' in data else '?'}\n"
 
-        if 'age_ratings' in keys:
+        if 'age_ratings' in data:
             rating = f"""{', '.join(AgeRatingCategories(age['category']).name + ' ' + AgeRatings(age['rating']).name
```
tagptroll1: Why does this need an f-string wrapped around it?  join returns a string. I'd use an f-string inside the join, but not around it.
```
rating = ', '.join(
    f'{AgeRatingCategories(age["category"]).name} {AgeRatings(age["rating"]).name}'
    for age
    in date['age_ratings']
)
```

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+            elif genre == "Turn-based strategy (tbs)":
+                self.genres["Turn-based-strategy"] = genres[genre]
+                self.genres["Tbs"] = genres[genre]
+            elif genre == "Real time strategy (rts)":
+                self.genres["Real-time-strategy"] = genres[genre]
+                self.genres["Rts"] = genres[genre]
+            elif genre == "Hack and slash/beat 'em up":
+                self.genres["Hack-and-slash"] = genres[genre]
+            else:
+                self.genres[genre] = genres[genre]
```
kwzrd: Looking over this is very confusing. Correct me if I'm wrong, but you're creating aliases for long genre names.

It would be a lot more maintainable to handle the aliasing logic in a generic loop, test it to make sure it works properly, and then simply define the aliases on the side.
```python
aliases = {
    "Role-playing (rpg)": ["Role-playing", "Rpg"],
}

for original, aliases in aliases.items():
    for alias in aliases:
        genres[alias] = original[original]
```
In this particular case, both aliases are case-less sub-strings of the original. There's room for improvement here in terms of how much the cog could aid the user with look-ups.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+        pages = await self.get_pages(games)
+        await ImagePaginator.paginate(pages, ctx, Embed(title=f'Top {amount} Games'))
+
+    @games.command(name='genres', aliases=['genre', 'g'])
+    async def genres(self, ctx: Context) -> None:
+        """Get all available genres."""
+        await ctx.send(f"Currently available genres: {', '.join(f'`{genre}`' for genre in self.genres)}")
+
+    @games.command(name='search', aliases=['s'])
+    async def search(self, ctx: Context, *, search: str) -> None:
```
kwzrd: Consider using a better name for the param, such as `search_term`.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+        Get random Game Companies companies from IGDB API.
+
+        Support amount parameter. Max is 25, min is 1.
+        """
+        if amount > 25:
+            await ctx.send("You can't get more than 25 companies at once.")
+            return
+        elif amount < 1:
+            await ctx.send("You can't get less than 1 company.")
+            return
```
kwzrd: I think just checking if `1 <= amount <= 25` would be sufficient. You can then explain to the user that `amount` escapes the allowed range, and show what the range is.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+        await ImagePaginator.paginate(pages, ctx, Embed(title='Random Game Companies'))
+
+    async def get_games_list(self,
+                             client: ClientSession,
+                             limit: int,
+                             genre: str = None,
+                             sort: str = None,
+                             additional_body: str = "",
+                             offset: int = 0) \
+            -> List[Dict[str, Any]]:
```
kwzrd: The prevalent style for long function signatures is as follows:
```python
def my_function(
    a: type,
    b: type = default
) -> return_type:
```
Please try to avoid line continuation with `\`. It is *very* rarely necessary.

Additionally:
- Does `client` need to be a param? You always pass in `self.http_session`. This function can simply grab the session from `self`.
- I'd prefer using `amount` rather than `limit`, especially if the parameter isn't explained anywhere. When calling the function, it is referred to as `amount`.
- What are allowed values for `sort`? What does it do?
- What am I supposed to pass into `additional_body`?
- What is `offset`?

A docstring of `Get Games List from IGDB API.` may not cut it for this function signature.

Regarding the type annotations - if a param may be of type `T` or `None`, annotate it as `Optional[T]`. Some of these string params default to None, whereas some default to an empty string. It's up to you to choose which one makes more sense: if the query can be safely constructed with empty strings, you can save yourself a conditional. If you'll have to check whether a non-default param was provided, using `None` to denote "no param was passed" in a conditional may look cleaner.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+        body = "fields cover.image_id, first_release_date, total_rating, name, storyline, url, platforms.name, "
+        body += "status, involved_companies.company.name, summary, age_ratings.category, age_ratings.rating, "
+        body += "total_rating_count;\n"
+
+        body += f"sort {sort};\n" if sort else ''
+        body += f"offset {offset};\n"
+        body += f"limit {limit};\n"
+
+        body += f"where genres = ({genre});" if genre else ''
+        body += additional_body
```
kwzrd: I find the repeated concatenation unnecessary. A string that won't fit into one line can be constructed as such:
```python
body = (
    "a very long "
    "message."
)
```
This won't distract from the content. Using f-strings remains allowed. If the conditionals are necessary, it may be better to assign the desired values to variables with descriptive names, and then constructing the string from them. Since this is an API query, you may want to consider using a template string at the top of the module, and then simply passing in these optional parameters to it as necessary. It's probably ok to keep it here though.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+            page += f"**Age Ratings:** {rating}\n"
+
+        if 'involved_companies' in data:
+            companies = ', '.join(co['company']['name'] for co in data['involved_companies'])
+        else:
+            companies = "?"
+        page += f"**Made by:** {companies}\n"
+
+        page += "\n"
+        page += data['storyline'] if 'storyline' in data else ''
```
kwzrd: This is beyond unreadable. Please consider using a template string and calling `.format` on it. Intermediate parts should be built separately and then joined together.

See my other comment regarding repeated concatenations.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+
+        # Define request body of IGDB API request and do request
+        body = f"""fields name, url, storyline, total_rating, total_rating_count;
+search "{search}";
+limit 50;"""
+
+        async with client.get(
+                url=BASE_URL + "games",
+                data=body,
+                headers=HEADERS) as resp:
```
kwzrd: On line 297, you style a similar expression in a preferred manner:
```python
async with client.get(
    url=BASE_URL + "companies",
    data=body,
    headers=HEADERS
) as resp:
```
Without getting into a discussion of which one is more "correct", it's good to be consistent.

By the way, this whole thing will comfortably fit into one line:
```python
async with client.get(url=BASE_URL + "companies", data=body, headers=HEADERS) as resp:
```

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+            lines.append(line)
+
+        return lines
+
+    async def get_companies_list(self, client: ClientSession, limit: int, offset: int = 0) -> List[Dict[str, Any]]:
+        """Get random Game Companies from IGDB API."""
+        # Create request body, define included fields, limit and offset, do request
+        body = f"""fields name, url, start_date, logo.image_id, developed.name, published.name, description;
+limit {limit};
+offset {offset};"""
```
kwzrd: Please see my other comments regarding long string construction. Is it necessary that the query contains newlines?

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+
+class Games(Cog):
+    """Games Cog contains commands that collect data from IGDB."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.http_session: ClientSession = bot.http_session
+
+        # Initialize genres
+        asyncio.get_event_loop().create_task(self._get_genres())
```
kwzrd: You can simply use the bot's own event loop: `bot.loop`. It's good to assign the returned task to an instance variable. That way, you can check on it later, if necessary for debug or other functionality, e.g. to check whether it has already finished.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+                self.genres["Rts"] = genres[genre]
+            elif genre == "Hack and slash/beat 'em up":
+                self.genres["Hack-and-slash"] = genres[genre]
+            else:
+                self.genres[genre] = genres[genre]
+
+    @group(name='games', aliases=['game'], invoke_without_command=True)
+    async def games(self, ctx: Context, genre: str = "", amount: int = 5) -> None:
+        """
+        Get random game(s) by genre from IGDB. Use .movies genres command to get all available genres.
```
kwzrd: You advise the user here to call `.movies`. I assume this was copy-pasted from the `Movie` cog.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+        self.bot = bot
+        self.http_session: ClientSession = bot.http_session
+
+        # Initialize genres
+        asyncio.get_event_loop().create_task(self._get_genres())
+
+    async def _get_genres(self) -> None:
+        """Create genres variable for games command."""
+        body = "fields name; limit 100;"
+        async with self.http_session.get(BASE_URL + "genres", data=body, headers=HEADERS) as resp:
```
sco1: Let's use an fstring here

```suggestion
        async with self.http_session.get(f"{BASE_URL}genres", data=body, headers=HEADERS) as resp:
```

It's probably more readable to strip the trailing `/` from `BASE_URL` so these concatenations would end up being `f"{BASE_URL}/genres"` instead.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+        elif amount < 1:
+            await ctx.send("You can't get less than 1 game.")
+            return
+
+        # Get games listing, if genre don't exist, show help.
+        try:
+            games = await self.get_games_list(self.http_session, amount, self.genres[genre],
+                                              offset=random.randint(0, 150))
+        except KeyError:
+            await ctx.send_help('games')
```
sco1: I don't think just sending the command's help here is useful. There should be feedback to the user that there's something wrong with the genre they provided.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,323 @@
...
+            data = await resp.json()
+
+        # Loop over games, format them to good format, make line and append this to total lines
+        for game in data:
+            line = ""
+
+            # Add games name and rating, also attach URL to title
+            line += f"**[{game['name']}]({game['url']})**\n"
+            line += f"""{'{0:.2f}'.format(game['total_rating'] if 'total_rating' in game.keys() else 0)}/100 :star: (by {
+            game['total_rating_count'] if 'total_rating_count' in game else '?'} users)"""
```
sco1: Having to mix so many things and use a multiline string on top of it is generally an indication that this should be broken up into discrete chunks.

ks129:
@sco1 @kwzrd After 6 hours of work, this is done.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,374 @@
...
+                             offset: int = 0
+                             ) -> List[Dict[str, Any]]:
+        """
+        Get list of games from IGDB API by parameters that is provided.
+
+        Amount param show how much movies this get, genre is genre ID and at least one genre in game must this when
+        provided. Sort is sorting by specific field and direction, ex. total_rating desc/asc (total_rating is field,
+        desc/asc is direction). Additional_body is field where you can pass extra search parameters. Offset show start
+        position in API.
+        """
```
eivl: The comment refers to movies. should be changed to games.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,374 @@
...
+
+        # Check for amounts, max is 25 and min 1
+        if not 1 <= amount <= 25:
+            await ctx.send("Your provided amount is out of range. Our minimum is 1 and maximum 25.")
+            return
+
+        # Get games listing, if genre don't exist, show error message with possibilities.
+        try:
+            games = await self.get_games_list(amount, self.genres[genre],
+                                              offset=random.randint(0, 150))
```
eivl: I do not find the offset, in this case, to be self-documented. Had to track down the `get_games_list` function to understand what it does. Maybe a clarification of why the offset is picked at random could be written as a comment?

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,374 @@
...
+        """
+        Get random Game Companies companies from IGDB API.
+
+        Support amount parameter. Max is 25, min is 1.
+        """
+        if not 1 <= amount <= 25:
+            await ctx.send("Your provided amount is out of range. Our minimum is 1 and maximum 25.")
+            return
+
+        companies = await self.get_companies_list(amount, random.randint(0, 150))
```
eivl: I would love keyword arguments so that I would know that this was a random offset. Same as the other ones
It's not self-documented so I have to look up `get_companies_list` to see what it is. As the other comments for similar methods, I still do not know why the offset is generated randomly.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,374 @@
...
+        # Get release date separately with checking
+        release_date = dt.utcfromtimestamp(data["first_release_date"]).date() if "first_release_date" in data else "?"
+
+        # Create Age Ratings value
+        rating = ", ".join(f"{AgeRatingCategories(age['category']).name} {AgeRatings(age['rating']).name}"
+                           for age in data["age_ratings"]) if "age_ratings" in data else "?"
+
+        companies = ", ".join(comp["company"]["name"] for comp in data["involved_companies"]) \
+            if "involved_companies" in data else "?"
+
```
eivl: I do not like the way the multi-line has been handled. they are inconsistent and overly messy. 
I do not have a suggestion atm, but we can have a discussion about it at your leisure. The contribution guide might have a suggestion about what to do.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,374 @@
...
+        """Create good formatted Game Company page."""
+        # Generate URL of company logo
+        url = LOGO_URL.substitute({"image_id": data["logo"]["image_id"] if "logo" in data else ""})
+
+        # Try to get found date of company
+        founded = dt.utcfromtimestamp(data["start_date"]).date() if "start_date" in data else "?"
+
+        # Generate list of games, that company have developed or published
+        developed = ", ".join(game["name"] for game in data["developed"]) if "developed" in data else "?"
+        published = ", ".join(game["name"] for game in data["published"]) if "published" in data else "?"
```
eivl: this is a great example of a readable, well-written comprehension and ternary expression.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,382 @@
...
+        if not 1 <= amount <= 25:
+            await ctx.send("Your provided amount is out of range. Our minimum is 1 and maximum 25.")
+            return
+
+        # Get games listing, if genre don't exist, show error message with possibilities.
+        # Offset must be random, due otherwise we will get always same result (offset show in which position should
+        # API start returning result)
+        try:
+            games = await self.get_games_list(amount, self.genres[genre],
+                                              offset=random.randint(0, 150))
```
ikuyarihS:
```suggestion
            games = await self.get_games_list(amount, self.genres[genre], offset=random.randint(0, 150))
```

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,382 @@
...
+
+        self.genres = {}
+
+        # Replace complex names with names from ALIASES
+        for genre in genres:
+            if genre in ALIASES:
+                for alias in ALIASES[genre]:
+                    self.genres[alias] = genres[genre]
+            else:
+                self.genres[genre] = genres[genre]
```
ikuyarihS: Since we only do `_get_genres()` **once**, if this fails we have to reload the entire cog everytime, so I suggest turning `self.genres` into a `property` instead like so
```py
@property
def gernes(self):
    ...
```
This way, we can also add another command to refresh `genres` whenever we want to.

Or, you can create a `discord.ext.tasks` that loop every x amount that will fetch the genres, and of course, a command to refresh `self.genres`.

ikuyarihS: since `genres` is a dictionary, you can loop over it's `.items()` to avoid having to do `genres[genre]` every time, something like
```py
for genre_name, genre in genres.items():
    ....
        self.genres[alias] = genre
```

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,403 @@
...
+            await ctx.send("Your provided amount is out of range. Our minimum is 1 and maximum 25.")
+            return
+
+        # Get games listing, if genre don't exist, show error message with possibilities.
+        # Offset must be random, due otherwise we will get always same result (offset show in which position should
+        # API start returning result)
+        try:
+            games = await self.get_games_list(amount, self.genres[genre], offset=random.randint(0, 150))
+        except KeyError:
+            possibilities = "`, `".join(difflib.get_close_matches(genre, self.genres))
```
ikuyarihS: Right now `difflib.get_close_matches()` is having inconsistent results, for example it cannot find `hack` and `slash` in `hack-and-slashes` so maybe you need a different way to suggest it?

image

image

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,403 @@
...
+            await ctx.send(f"Invalid genre `{genre}`. {f'Maybe you meant `{possibilities}`?' if possibilities else ''}")
+            return
+
+        # Create pages and paginate
+        pages = [await self.create_page(game) for game in games]
+
+        await ImagePaginator.paginate(pages, ctx, Embed(title=f"Random {genre} Games"))
```
ikuyarihS: Right now there are way too many blank lines between games, consider using the keyword `empty`

image

Thanks @sco1 for finding out about that!

ikuyarihS:
Also right now it is impossible to find genres with spaces between them, consider using `Greedy` or parse the arguments manually instead

image

bot/seasons/evergreen/game.py
eivl: interesting... what would fuzzy-matching return in this case?

ks129:
@ikuyarihS I don't added auto recalling fetch for games when 1 item in possibilities, due this have conflict with `cutoff`.

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,395 @@
...
+            await ctx.send("Your provided amount is out of range. Our minimum is 1 and maximum 25.")
+            return
+
+        # Get games listing, if genre don't exist, show error message with possibilities.
+        # Offset must be random, due otherwise we will get always same result (offset show in which position should
+        # API start returning result)
+        try:
+            games = await self.get_games_list(amount, self.genres[genre], offset=random.randint(0, 150))
+        except KeyError:
+            possibilities = "`, `".join(difflib.get_close_matches(genre, self.genres, cutoff=0.4))
```
ikuyarihS: WIth `cutoff=0.4` the results seems a bit wild, so maybe increasing this back?
image

bot/seasons/evergreen/game.py
```diff
@@ -0,0 +1,395 @@
...
+    async def genres(self, ctx: Context) -> None:
+        """Get all available genres."""
+        await ctx.send(f"Currently available genres: {', '.join(f'`{genre}`' for genre in self.genres)}")
+
+    @games.command(name="search", aliases=["s"])
+    async def search(self, ctx: Context, *, search_term: str) -> None:
+        """Find games by name."""
+        lines = await self.search_games(search_term)
+
+        await LinePaginator.paginate(lines, ctx, Embed(title=f"Game Search Results: {search_term}"), empty=False)
```
ikuyarihS: This is a minor issue, can be ignored and addressed later - right now there is no space between each results, so either move the rating up to the name or add a space in between.

image

PR #373, opened 2020-03-08
Bump Python version to 3.8

### LINKED ISSUE #372: Upgrade Seasonalbot to Python 3.8
With `bot` and `snekbox` now being bumped to 3.8, we should also look into bumping Seasonalbot.

This is a fairly self-explanatory issue.

As explained in https://github.com/python-discord/bot/issues/758, one of the motivations for upgrading were the improvements to unittesting asynchronous code. We do not have a testing suite for Seasonalbot, which means that we won't benefit from those improvements here, but it also means that we can more or less expect everything to work out-of-the-box, perhaps with minimal adjustments.

I would be happy to look into this myself.

Although I'm currently working on #329, I think this should probably be a separate PR.

kwzrd:
This PR essentially replicates the efforts of https://github.com/python-discord/bot/pull/783.

Bumps Python to 3.8 in Pipfile & lock, Azure and Docker.
Bumps the Ubuntu image in CI to 18.04.

So far it appears that Seasonalbot is fully compatible with 3.8 (tested on 3.8.2) as is, without any required adjustments. I will test some more tomorrow before taking this out of draft status.

Closes #372

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.7-slim
+FROM python:3.8-slim
 
 # Set pip to have cleaner logs and no saved cache
 ENV PIP_NO_CACHE_DIR=false \
--- a/Pipfile
+++ b/Pipfile
@@ -25,7 +25,7 @@ pep8-naming = "~=0.9"
 pre-commit = "~=2.1"
 
 [requires]
-python_version = "3.7"
+python_version = "3.8"
 
 [scripts]
 start = "python -m bot"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,11 +1,11 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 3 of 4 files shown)


kwzrd:
All seasons load fine, works under Docker too.

Can't really think of anything else to test, so I'm opening this for review.

kwzrd:
@SebastiaanZ I requested your review specifically because you handled the migration for `bot`, so there may still be some things fresh in memory that had to be adjusted, but I'm fairly sure they're more or less irrelevant for Seasonalbot.

SebastiaanZ:
> @SebastiaanZ I requested your review specifically because you handled the migration for `bot`, so there may still be some things fresh in memory that had to be adjusted, but I'm fairly sure they're more or less irrelevant for Seasonalbot.

I can look at it tomorrow. I don't think there's anything particularly interesting that applies to Seasonal Bot, except maybe the handling of those asyncio exceptions.

kwzrd:
@SebastiaanZ That's perfectly fine, there's no need to rush it.

PR #382, opened 2020-03-23
Game genre matching fix

### LINKED ISSUE #369: .games not accurately get genre
## Description
Right now when searching for a genre, the genre has to be 100% similar to the genre listed, otherwise it will be regarded as not correct. The code responsible for getting the searched genre and suggesting genres can be seen here:
https://github.com/python-discord/seasonalbot/blob/ce8fb0f8f541ade993d6c93bdcefa33a63287311/bot/seasons/evergreen/game.py#L194-L199

Some example of not being able to find the genre can be seen here:
- All the genres:
image
- Finding for `Hack and slash`
image
- Finding for `Point-and-click`
image

## Expected Behaviour
When searching for `Hack` the cog should try to find and score the query against the genres and give back matching scores - if there is only one fitting match ( score above 60 for example ) then return the only match, that way, `.game hack` will work as if we do `.game Hack and slash` / `.game point click` will work like `.game point-and-click`

## Actual Behaviour
Unless the query is exactly the same as the genre ( with dashes, no missing letter, case-insensitive ) the command will disregard the query and try to suggest instead.

## Possible Solutions
Since the cog is using `difflib.get_close_matches()` we can examine some snippets to see what it actually is doing:

Using this as the base:
```py
import difflib

string = "Simulator, Tactical, Quiz/trivia, Fighting, Strategy, Adventure, Role playing, Rpg, Shooter, Music, Indie, Turn based strategy, Tbs, Pinball, Puzzle, Real time strategy, Rts, Hack and slash, Visual novel, Platform, Racing, Sport, Arcade, Point-and-click"
genres = string.split(', ')
```
Here are results of using `get_close_matches()`
```py
print('point', difflib.get_close_matches('point', genres), sep='\n')
print('point-and', difflib.get_close_matches('point-and', genres), sep='\n')
```
```py
point
['Sport']
point-and
['Point-and-click']
```
and using `SequenceMatcher().ratio()`
```py
def get_top(query: str, number: int = 4) -> list:
    results = []
    for genre in genres:
        sm = difflib.SequenceMatcher(None, query, genre)
        results.append((round(sm.ratio(), 2), genre))
    return sorted(results, reverse=True)[:number]

print('point', *get_top('point'), sep='\n')
print('point-and', *get_top('point-and'), sep='\n')
```
```py
point
(0.6, 'Sport')
(0.4, 'Point-and-click')
(0.36, 'Racing')
(0.35, 'Role playing')
point-and
(0.67, 'Point-and-click')
(0.43, 'Sport')
(0.38, 'Pinball')
(0.3, 'Quiz/trivia')
```
Now, if we look at `get_close_matches()` we will see a `cutoff=0.6` being the default, so in both cases, we see only the gernes with ratio above `0.6` get returned, which is consistent.

A suggested solution is to modify this process of getting ratios, here is a very crude example of what we can do:
```py
import re

regex_non_alphabet = re.compile(r"[^a-z0-9]", re.IGNORECASE)

def get_top(query: str, number: int = 4) -> list:
    results = []
    for genre in genres:
        ratios = [difflib.SequenceMatcher(None, query, genre).ratio()]
        for word in regex_non_alphabet.split(genre):
            ratios.append(difflib.SequenceMatcher(None, query, word).ratio())
        results.append((round(max(ratios), 2), genre))
    return sorted(results, reverse=True)[:number]

print('point', *get_top('point'), sep='\n')
print('point-and', *get_top('point-and'), sep='\n')
```
```py
point
(0.8, 'Point-and-click')
(0.6, 'Sport')
(0.5, 'Role playing')
(0.36, 'Racing')
point-and
(0.67, 'Point-and-click')
(0.5, 'Hack and slash')
(0.43, 'Sport')
(0.38, 'Role playing')
```
In this case, we increased the ratio for `point-and-click` up to `0.8` for `point` making it better.
But of course, this approach has its downfall as well, so I strongly suggest trying different approach until we see a preferred outcome ( `point` and `point click` will both only yield `point-and-click` )

ks129:
## Relevant Issues

Closes #369 

## Description

I added matching function (thanks to @ikuyarihS for code).

## Reasoning

Due currently when you type genre wrong, this only suggest genres, but this will auto use correct genre.

## Did you:

### DIFF
--- a/bot/seasons/evergreen/game.py
+++ b/bot/seasons/evergreen/game.py
@@ -1,6 +1,7 @@
 import difflib
 import logging
 import random
+import re
 from datetime import datetime as dt
 from enum import IntEnum
 from typing import Any, Dict, List, Optional, Tuple
@@ -24,6 +25,8 @@ HEADERS = {
 }
 
 logger = logging.getLogger(__name__)
+
+REGEX_NON_ALPHABET = re.compile(r"[^a-z0-9]", re.IGNORECASE)
 
 # ---------
 # TEMPLATES
@@ -384,6 +387,16 @@ class Games(Cog):
         page = COMPANY_PAGE.format(**formatting)
 
         return page, url
+
+    async def get_best_results(self, query: str, nr: int = 4) -> List[Tuple[float, str]]:
+        """Get best match result of genre when original genre is invalid."""
+        results = []
+        for genre in self.genres:
+            ratios = [difflib.SequenceMatcher(None, query, genre).ratio()]
+            for word in REGEX_NON_ALPHABET.split(genre):
+                ratios.append(difflib.SequenceMatcher(None, query, word).ratio())
+            results.append((round(max(ratios), 2), genre))
+        return sorted(results, reverse=True)[:nr]
 
 
 def setup(bot: SeasonalBot) -> None:
--- a/bot/seasons/evergreen/game.py
+++ b/bot/seasons/evergreen/game.py
@@ -400,7 +400,7 @@ class Games(Cog):
 
         return page, url
 
-    async def get_best_results(self, query: str, nr: int = 4) -> List[Tuple[float, str]]:
+    async def get_best_results(self, query: str) -> List[Tuple[float, str]]:
         """Get best match result of genre when original genre is invalid."""
         results = []
         for genre in self.genres:
@@ -408,7 +408,7 @@ class Games(Cog):
             for word in REGEX_NON_ALPHABET.split(genre):
                 ratios.append(difflib.SequenceMatcher(None, query, word).ratio())
             results.append((round(max(ratios), 2), genre))
-        return sorted(results, reverse=True)[:nr]
+        return sorted(results, reverse=True)[:4]
 
 
 def setup(bot: SeasonalBot) -> None:
--- a/bot/seasons/evergreen/game.py
+++ b/bot/seasons/evergreen/game.py
@@ -201,6 +201,9 @@ class Games(Cog):
             possibility = None
             # Check is there any possibility that ratio is higher than 0.60
             for p in possibilities:
+                # If there is more than 1 genre that have higher than 0.60 ratio, then show list of possibilities
+                if sum(1 for po in possibilities if po[0] >= 0.60) > 1:
+                    break
                 if p[0] >= 0.60:
                     possibility = await self.get_games_list(amount, self.genres[p[1]], offset=random.randint(0, 150))
                     genre = p[1]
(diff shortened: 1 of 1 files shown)


ikuyarihS:
Everything is working great!

bot/seasons/evergreen/game.py
```diff
@@ -194,9 +197,23 @@ def cog_unload(self) -> None:
...
-            await ctx.send(f"Invalid genre `{genre}`. {f'Maybe you meant `{possibilities}`?' if possibilities else ''}")
-            return
+            possibilities = await self.get_best_results(genre)
+            possibility = None
+            # Check is there any possibility that ratio is higher than 0.60
+            for p in possibilities:
+                if p[0] >= 0.60:
+                    possibility = await self.get_games_list(amount, self.genres[p[1]], offset=random.randint(0, 150))
+                    genre = p[1]
+                    break
```
ikuyarihS: This part can use extra logics check, checking for the highest score might not always work.

However, currently it is working pretty well with all the test cases I provided, so you can argue otherwise that being practical is better in this case. Still, I am keen on seeing a suggestion on this part,

ks129: Maybe one thing that this should check is the first letter?

ikuyarihS: It will be more complicated than that, for example when you do `.games and` you will actually get `Point and click` and you will not see `hack-and-slash` because the check grab the first result already.

Granted that no one should ever search for a genre called `and` this is something we can do to to improve user's experience.

An example of what we can do, is that if all returned results have scores greater than 0.6 we can list all of them instead, like when searching for tags and there are multiple matching suggestions.

ks129: I added now this check. Is now better?

ikuyarihS: It does sound better in theory, I suggest modifying `get_best_results()` to accept an optional argument for score, then you only have to check for the length of the list it returns, that'll simplify the checks outside!

ks129: I don't see reason for the additional argument, so I hardcoded it, due this is used only in one place. I also simplified checking in `.games` command side.

ikuyarihS: That's true.

In your new code, 
```py
display_possibilities = "`, `".join(p[1] for p in possibilities if p[0] >= 0.40)
```
Will practically be the same as
```py
display_possibilities  = "`, `".join(p[1] for p in possibilities)
```
Because the function only returns genres with ratio >= 0.6 - the check for >= 0.4 is redundant.

ks129: Okay, fixed it.

ikuyarihS: Working great!

ks129:
@sco1 Okay, I changed this now.

PR #337, opened 2020-01-06
Add exception where relayed message.content length exceed 2048 charac…

AG4lyf:
…ters

---
name: Exception of character limits
about: Resolve a issue where bookmarked message content is near 2048
issue: Issue #336 

---

## Pull Request Details

Please ensure your PR fulfills the following criteria:

- [check] Have you joined the [PythonDiscord Community](https://pythondiscord.com/invite)?
- [check] Were your changes made in a Pipenv environment?
- [check] Does flake8 pass (```pipenv run lint```)

## Additional information

New message should look like this

An '.....' is pretty much enough to indicate user that there is more.
Screenshot_88

PR #386, opened 2020-03-29
Bookmark

### LINKED ISSUE #338: Blank .bm bookmarking request message

### LINKED ISSUE #341: Allow Users to React to a Bookmark to Recieve a Copy

AG4lyf:
This PR is addition to #349 
I had to destroy that PR because it got old and due to some reasons(me not knowing how to use github CLI tool) I couldn't update the branch and i can't fork until old fork is deleted so here we go.

Closes #341  , Closes #338 

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -89,6 +89,7 @@ class Emojis:
     christmas_tree = "\U0001F384"
     check = "\u2611"
     envelope = "\U0001F4E8"
+    pin = "\U0001f4cc"
     trashcan = "<:trashcan:637136429717389331>"
 
     terning1 = "<:terning1:431249668983488527>"
--- a/bot/seasons/evergreen/bookmark.py
+++ b/bot/seasons/evergreen/bookmark.py
@@ -92,7 +92,8 @@ class Bookmark(commands.Cog):
                 await ctx.message.clear_reactions()
                 return
 
-            log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from '{ctx.author}'.")
+            log.info(f"{user.name} tried bookmarking {target_message.jump_url} with title '{title}' from "
+                     f"'{ctx.author}'.")
 
             try:
                 if user == ctx.author:
--- a/bot/seasons/evergreen/bookmark.py
+++ b/bot/seasons/evergreen/bookmark.py
@@ -37,6 +37,11 @@ class Bookmark(commands.Cog):
                     target = message
                     break
 
+        if type(target) == discord.User:  # target is still User that means message not found
+            log.info(f'{ctx.author.name} Tried to bookmark {target} message but no message was found.')
+            await ctx.send(f'None of the previous 20 message is from {target.name}.', delete_after=7)
+            return
+
         # Prevent users from bookmarking a message in a channel they don't have access to
         permissions = ctx.author.permissions_in(target.channel)
         if not permissions.read_messages:
--- a/bot/seasons/evergreen/bookmark.py
+++ b/bot/seasons/evergreen/bookmark.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 import random
 from copy import deepcopy
+from typing import Union
 
 import discord
 from discord.ext import commands
@@ -19,20 +20,23 @@ class Bookmark(commands.Cog):
 
     @commands.command(name="bookmark", aliases=("bm", "pin"))
     async def bookmark(
-        self,
-        ctx: commands.Context,
-        target_message: discord.Message = None,
-        *,
-        title: str = "Bookmark"
+            self,
+            ctx: commands.Context,
+            target: Union[discord.Message, discord.User],
+            *,
+            title: str = "Bookmark"
     ) -> None:
         """Send the author a link to `target_message` via DMs. If Nothing is provided it take the commanding message."""
-        if target_message is None:
-            target_message = ctx.message
+        if type(target) == discord.User:
+            async for message in ctx.channel.history(limit=20):  # Takes last 2 messages from the channel.
+                if message.author == target:
+                    target = message
+                    break
 
         # Prevent users from bookmarking a message in a channel they don't have access to
-        permissions = ctx.author.permissions_in(target_message.channel)
+        permissions = ctx.author.permissions_in(target.channel)
         if not permissions.read_messages:
-            log.info(f"{ctx.author} tried to bookmark a message in #{target_message.channel} but has no permissions.")
+            log.info(f"{ctx.author} tried to bookmark a message in #{target.channel} but has no permissions.")
             embed = discord.Embed(
                 title=random.choice(ERROR_REPLIES),
                 color=Colours.soft_red,
@@ -44,10 +48,10 @@ class Bookmark(commands.Cog):
         embed = discord.Embed(
             title=title,
             colour=Colours.soft_green,
-            description=target_message.content
+            description=target.content
         )
-        embed.add_field(name="Wanna give it a visit?", value=f"[Visit original message]({target_message.jump_url})")
-        embed.set_author(name=target_message.author, icon_url=target_message.author.avatar_url)
+        embed.add_field(name="Wanna give it a visit?", value=f"[Visit original message]({target.jump_url})")
+        embed.set_author(name=target.author, icon_url=target.author.avatar_url)
         embed.set_thumbnail(url=bookmark_icon_url)
 
         error_embed = discord.Embed(
@@ -66,7 +70,7 @@ class Bookmark(commands.Cog):
             await ctx.message.add_reaction(Emojis.envelope)
             sent_person = {ctx.author}  # set of id who got the message
 
-        log.info(f"{ctx.author} bookmarked {target_message.jump_url} with title '{title}'.")
+        log.info(f"{ctx.author} bookmarked {target.jump_url} with title '{title}'.")
         await ctx.message.add_reaction(Emojis.pin)
 
         copied_embed = deepcopy(embed)
@@ -92,7 +96,7 @@ class Bookmark(commands.Cog):
                 await ctx.message.clear_reactions()
                 return
 
-            log.info(f"{user.name} tried bookmarking {target_message.jump_url} with title '{title}' from "
+            log.info(f"{user.name} tried bookmarking {target.jump_url} with title '{title}' from "
                      f"'{ctx.author}'.")
 
             try:
@@ -111,6 +115,7 @@ class Bookmark(commands.Cog):
 
             except discord.HTTPException:
                 await ctx.send("Unknown error while sending bookmark to user.", delete_after=7.5)
+                await reaction.remove(user)
 
             else:
                 sent_person.add(user)
--- a/bot/seasons/evergreen/bookmark.py
+++ b/bot/seasons/evergreen/bookmark.py
@@ -1,5 +1,7 @@
+import asyncio
 import logging
 import random
+from copy import deepcopy
 
 import discord
 from discord.ext import commands
@@ -19,15 +21,18 @@ class Bookmark(commands.Cog):
     async def bookmark(
         self,
         ctx: commands.Context,
-        target_message: discord.Message,
+        target_message: discord.Message = None,
         *,
         title: str = "Bookmark"
     ) -> None:
-        """Send the author a link to `target_message` via DMs."""
+        """Send the author a link to `target_message` via DMs. If Nothing is provided it take the commanding message."""
+        if target_message is None:
+            target_message = ctx.message
+
         # Prevent users from bookmarking a message in a channel they don't have access to
         permissions = ctx.author.permissions_in(target_message.channel)
         if not permissions.read_messages:
-            log.info(f"{ctx.author} tried to bookmark a message in #{target_message.channel} but has no permissions")
+            log.info(f"{ctx.author} tried to bookmark a message in #{target_message.channel} but has no permissions.")
             embed = discord.Embed(
                 title=random.choice(ERROR_REPLIES),
                 color=Colours.soft_red,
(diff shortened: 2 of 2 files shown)


AG4lyf:
About this
image
I tried making a branch named `bookmark` like how we normally make from switching branch menu, turned out it was not possible ;-; weird.

ks129:
Overall looks good, but one thing: I really don't suggest making commit messages like `initial commit`. This can be much more descriptive, like `Bookmark: Added thing 1, thing 2 ...` when you want to make big commit messages, but better will be ever split these to small commits. @AG4lyf

AG4lyf:
@ks129 this is just the same commits from old PR . there is nothing changed actually. so same commit messages of #349 applies here

bot/constants.py
```diff
@@ -89,6 +89,7 @@ class Emojis:
     christmas_tree = "\U0001F384"
     check = "\u2611"
     envelope = "\U0001F4E8"
+    pin = "\U0001f4cc"
```
ikuyarihS: Is there a reason why you introduce an entirely new constant for this feature?

After invoking `.bm` now every message will have 2 emojis instead of one. Other user when reacting with `envelope` will not receive anything at all, and that is very confusing.

I do not think that we need to introduce a new constant and add a totally new emoji to do it, `envelope` will do just fine.

bot/seasons/evergreen/bookmark.py
```diff
@@ -19,15 +21,18 @@ def __init__(self, bot: commands.Bot):
...
         ctx: commands.Context,
-        target_message: discord.Message,
+        target_message: discord.Message = None,
         *,
         title: str = "Bookmark"
     ) -> None:
-        """Send the author a link to `target_message` via DMs."""
+        """Send the author a link to `target_message` via DMs. If Nothing is provided it take the commanding message."""
+        if target_message is None:
+            target_message = ctx.message
```
ikuyarihS: By doing this, now when someone do `.bm` instead of getting a help message, the bot will bookmark the message with `.bm` instead like here

image

image

There need to be a stricter check here to prevent this to happen.

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,21 +50,75 @@ def __init__(self, bot: commands.Bot):
...
+            )
+
+        while True:
+            try:
+                reaction, user = await self.bot.wait_for("reaction_add", timeout=60.0, check=check)
+            except asyncio.TimeoutError:
+                await ctx.message.clear_reactions()
+                return
+
+            log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from '{ctx.author}'.")
```
ikuyarihS: This should be moved much further down, so we only log when we successfully help someone bookmark the message.

A failing case example will be when we cannot send the DM to user, even after logging that "user bookmarked message". Either change the logging message or move this line further down.

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,21 +50,75 @@ def __init__(self, bot: commands.Bot):
...
+            except discord.Forbidden:
+                await ctx.send(f"{user.mention} Please enable your DM to receive the message.", delete_after=7.5)
+                try:
+                    await reaction.remove(user)
+                except discord.Forbidden:
+                    await ctx.send("I don't have permissions to remove reaction.", delete_after=7.5)
+                except discord.HTTPException:
+                    await ctx.send("Unknown error while removing reaction.", delete_after=7.5)
+                except discord.NotFound:
+                    await ctx.send("Message or reaction author not found, can't remove reaction.", delete_after=7.5)
```
ikuyarihS: This is caught by pylint:

image

Which makes sense, because `NotFound` is a subclass of `HTTPException` - Consider changing the order of catching.

bot/seasons/evergreen/bookmark.py
```diff
@@ -45,21 +50,75 @@ def __init__(self, bot: commands.Bot):
...
+                return
+
+            log.info(f"{user.name} bookmarked {target_message.jump_url} with title '{title}' from '{ctx.author}'.")
+
+            try:
+                if user == ctx.author:
+                    await user.send(embed=embed)
+                else:
+                    await user.send(embed=copied_embed)
+
```
ikuyarihS: This blank line should be removed to make it much easier to see which except belong to what try in my opinion, separating `except` with a blank line does not make it more readable at all.

ikuyarihS:
Also I do not think that `initial commit` is a good commit message. If we have 400 PRs and each one of them have an `initial commit` then we will have 400 commits with that same messages, which is definitely not helpful at all.

bot/constants.py
AG4lyf: As @sco1 said 
`I don't think it makes sense from a UX perspective to utilize the same reaction for multiple purposes when one of those purposes silently disappears after a timeout. This is already barely discoverable functionality, and with this implementation a user that does know of its existence doesn't know if it has timed out or if it's broken. A second emoji solves half of the problem, which might be sufficient.`

And i agree on this rather than using same emoji.

ikuyarihS: If that's the case, you might want to think about how to avoid the confusion that is having 2 emojis, and one does not do anything at all, from the user's perspective.

If the envelope is purely for visual confirmation that the command worked, it might not be needed anymore since it is not adding anything else.

AG4lyf: I mean the envelope emoji doesn't get sent when it fails to dm . But pin still work and it give author an another chance to bookmark it. 

I like how it is now myself. 
An unknowing person will press both.
Or we can make it so it remove react when someone react with envelope.

bot/seasons/evergreen/bookmark.py
AG4lyf: I mean anyone can use .help bm to get the usage. 
For now it looks clean and easy to remember. One cannot remember a special word for just a command.

Or maybe you can suggest some better checks ?

ikuyarihS: The problem is not that we have a `.help bm` - the problem is that `.bm` by itself, without any arguments, will work.

So going by that, you can definitely check the part where there is no message being passed to it - when it is `None`

AG4lyf: That is the scope of #338 and that is why it is there.

ikuyarihS: Right now it is not doing what is said in #338 at all, like in the screenshot, it is bookmarking the **current message invoking the command** which is the message with `.bm`

bot/seasons/evergreen/bookmark.py
AG4lyf: I was told in my first contribution to log even though we fail to send DM . In that case I should use `Tried to bookmark`.

AG4lyf:
Since git F it up 
so commit message for this ^ are this
`Added functionality to bookmark using user mention , basically search for the last message by that user in that channel 
renamed target_message to target`

bot/seasons/evergreen/bookmark.py
AG4lyf: Latest commit make it better . And .bm
Now invoke its usage error

bot/seasons/evergreen/bookmark.py
```diff
@@ -17,17 +20,27 @@ def __init__(self, bot: commands.Bot):
...
+        """
+        if type(target) == discord.User:
+            async for message in ctx.channel.history(limit=20):  # Takes last 2 messages from the channel.
+                if message.author == target:
+                    target = message
+                    break
+
         # Prevent users from bookmarking a message in a channel they don't have access to
-        permissions = ctx.author.permissions_in(target_message.channel)
+        permissions = ctx.author.permissions_in(target.channel)
```
ikuyarihS, line 41: This will fail when doing a `.bm @seasonalbot` raising an error `AttributeError: 'ClientUser' object has no attribute 'channel'` - You will need to check if target is a bot as well.

bot/seasons/evergreen/bookmark.py
```diff
@@ -39,27 +52,80 @@ def __init__(self, bot: commands.Bot):
         embed = discord.Embed(
             title=title,
```
ikuyarihS, line 53: setting `title` as the title of the embed will break discord's special format, such as when bookmarking using a channel as the name, like this

image
image

You may want to put it in the description instead.

bot/seasons/evergreen/bookmark.py
```diff
@@ -17,17 +20,27 @@ def __init__(self, bot: commands.Bot):
...
+        """
+        Send the author a link to `target_message` via DMs.
+
+        A user mention can also be passed to get the user's last message in channel where the command is invoked.
+        """
+        if type(target) == discord.User:
+            async for message in ctx.channel.history(limit=20):  # Takes last 2 messages from the channel.
+                if message.author == target:
+                    target = message
+                    break
```
ikuyarihS, line 38: There should be another special check here, otherwise trying to bookmarking your last message will result in the command message being bookmarked, like this:

image
image

In practice user may want to bookmark their "other" latest message instead.

bot/seasons/evergreen/bookmark.py
AG4lyf: You want me to break the symmetry how it used to work previously
if yes What you think title will be so that it is easy to find ?

ikuyarihS: Actually, you can definitely use `.clean_content` of the message, that should escape and show the name of the channel properly.

If you need to escape markdown as well, you will need to look at [discord.utils.escape_markdown](https://discordpy.readthedocs.io/en/latest/api.html#discord.utils.escape_markdown)

AG4lyf: I think you mean
https://discordpy.readthedocs.io/en/latest/api.html#discord.utils.escape_mentions
But that does not contain channel mention.
Guess we can't make everything work.

ikuyarihS: It is linked in the comment above.

And as I've said, you can certainly use `.clean_content` to escape mentions for user / channels.

AG4lyf: clean_content can only be applied on message object not string I can do that on target.clean_content not title.

bot/seasons/evergreen/bookmark.py
```diff
@@ -17,17 +20,27 @@ def __init__(self, bot: commands.Bot):
...
     ) -> None:
-        """Send the author a link to `target_message` via DMs."""
+        """
+        Send the author a link to `target_message` via DMs.
+
+        A user mention can also be passed to get the user's last message in channel where the command is invoked.
+        """
+        if type(target) == discord.User:
+            async for message in ctx.channel.history(limit=20):  # Takes last 2 messages from the channel.
+                if message.author == target and message != ctx.message:
```
AG4lyf, line 36: Should compare message.id rather than message object.
IDK is that an issue comparing message object not working as it should be.

bot/seasons/evergreen/bookmark.py
ikuyarihS: I think you can, you are forgetting `ctx.message` in this case. Either way, the format on the title should be addressed, as `<#267659945086812160>` does not make a great title.

AG4lyf: ctx.message.content will return whole command string including `.bm @user1`

AG4lyf: I can trick discord.py actually but i don't feel like that is a good way to do that 
basically change ctx.message.content = title
that way I can use clean_content

AG4lyf: image
image
I mean why not ?

ks129: Maybe should better remove title as `title` and use:
```
**Title:** {title}
{content}
```
as description?

AG4lyf: Ohh this got a good story
If a message is exactly discord message length
this will crash the command as now description is higher than limit

ks129: Hmm. Yes. For this you should make length check, like 1900 characters (2000 is Discord max).

AG4lyf: #337 you should read this ,
And i don't think there is need to change how people are familiar with bookmark, for example if you see all the bm at once and they are different it can be hard to understand,

AG4lyf: image
image

This is perfect.

bot/seasons/evergreen/bookmark.py
ikuyarihS: You will have to check the docs if the class supports equality checking, for example [discord.TextChannel](https://discordpy.readthedocs.io/en/latest/api.html#textchannel) supports it
image
But [discord.Message](https://discordpy.readthedocs.io/en/latest/api.html#message) does not, so yes you need to check via `.id`

bot/seasons/evergreen/bookmark.py
AG4lyf: Added a check. so resolving this convo

ikuyarihS: This has not been solved yet, when I do `.bm @seasonalbot hi` I got this traceback
```py
  File "/usr/local/lib/python3.8/site-packages/discord/ext/commands/core.py", line 83, in wrapped
  File "/bot/bot/seasons/evergreen/bookmark.py", line 46, in bookmark
    permissions = ctx.author.permissions_in(target.channel)
AttributeError: 'ClientUser' object has no attribute 'channel'
```

kwzrd:
@AG4lyf This has been stale for a while now, has all feedback been addressed? Do you need help resolving the conflicts here?

kwzrd:
I will close this due to inactivity, @Akarys42 is now assigned on both linked issues.

PR #389, opened 2020-03-31
Wikipedia command

### LINKED ISSUE #331: Addition of a new command !wiki
Hi

So I have a suggestion, not sure if here's where I should post. If it's the wrong place, let me know so I could relocate it

I was thinking, we could have a !wiki command
where !wiki 'xyz' would make the bot fetch the wikipedia article on xyz if it exists.

Just a thought

vivax3794:
## Relevant Issues
Closes #331 

## Description
Added a new command `.wikipedia` (with alias `.wiki`) that uses the Wikipedia api to get a search result.
I decided to make the bot just print the wikipedia link instead of making a custom embed, since discord comes with a nice embed for wikipedia nativly. This can of course be changed if we think it is needed.

## Screenshots
image
image

## Did you:


ks129:
@vivax3794 First, I think this should have more features than just link showing. This should use its own embed with parsed HTML (to MD). Additionally, this should have `.wiki search <search>` command, that uses `LinePaginator` and shows all results that Wikipedia returns. Also, one feature is supporting custom wikis, because all/most wikis that use MediaWiki have open API. So like `.wiki custom <base_url> <article>` and `.wiki custom search <base_url> <search_term>`.

One thing that I want to note is your commit messages. `Got it working`. When you see it in the commit history, what this says to you? Make small and well-written commit messages, instead of big and bad quality messages.

bot/seasons/evergreen/wikipedia.py
```diff
@@ -0,0 +1,52 @@
...
+        search_result = data["query"]["search"]
+        if len(search_result) == 0:
+            return None
+
+        page = search_result[0]
+
+        # we dont like "may refere to" pages.
+        if "may refer to" in page["snippet"]:
+            log.info(f"hit a 'may refer to' page, taking result 2 (index 1)")
+            page = search_result[1]
```
ikuyarihS: I believe this can use more logic to check the results we got, otherwise we can have this

image

In case we got too many like when we hit a `may refer to` as the first page, maybe wikipedia does not have it and so we can list the results instead?

bot/seasons/evergreen/wikipedia.py
```diff
@@ -0,0 +1,52 @@
...
+        self.http_session = bot.http_session
+
+    async def search_wikipedia(self, search_term: str) -> Optional[str]:
+        """Search wikipedia and return the first page found."""
+        async with self.http_session.get(SEARCH_API.format(search_term=search_term)) as response:
+            data = await response.json()
+
+        search_result = data["query"]["search"]
+        if len(search_result) == 0:
+            return None
```
ikuyarihS: The returned API does contain the amount of hits, and using this is much more accurate, specially when the API can cut off the amount of results it return, like in this case:

image

I think showing this number is cool as well, as well as the suggestion that wikipedia suggested!

bot/seasons/evergreen/wikipedia.py
ikuyarihS: We can definitely use some logic here instead of relying on the very first result, specially when they are sorted alphabetically and not by how close the match is, for example when searching for `Avenger` I got this:

image

And the first hit is this:
image

I mean, when searching wikipedia for Marvel's Avenger, I definitely am not searching for a `Deer Avenger` right? :D

vivax3794:
@ks129 about parsing that html, as far as I see it looks like all the html returned (at least in the snippets) are just `<span class=\"searchmatch\">something</span>`  so maybe it would just be easieste to use `str.replace` to remove those?

ks129:
I suggest using [markdownify](https://pypi.org/project/markdownify/). It's a very good tool for this. @vivax3794

vivax3794:
sorry for not working on this, something has happen in my life and dont have the motivation to code right now, and this quarantine is not helping... so sorry.

PR #391, opened 2020-04-03
Send greeting message on bot start-up

### LINKED ISSUE #390: Send greeting message on start-up
In https://github.com/python-discord/seasonalbot/pull/329, the `SeasonManager` cog was removed. It was responsible for dispatching a greeting message to the `#dev-log` channel, much like Python does:

image

SeasonalBot does therefore not currently dispatch any such message. I think this is a regression, as the message lets us know that the bot has successfully connected after deployment, and we're already used to seeing it.

I think this should be added. However, I would like to ask for best practice advice. Python's greeting is dispatched via a task in its [`bot.cogs.logging`](https://github.com/python-discord/bot/blob/master/bot/cogs/logging.py#L21) cog. The advantage of this is that:
* It is guaranteed to only dispatch once
* Setting up a cog for sending messages to the `#dev-log` channel could be useful in the future

An alternative would be to simply register an `on_ready` listener, and dispatch the message there. The difference is that an `on_ready` can technically fire more than once ([docs](https://discordpy.readthedocs.io/en/latest/api.html#discord.on_ready)), e.g. upon reconnection. Is this necessarily a bad thing though? Wouldn't it be interesting to know that the bot had to reconnect? Does this potentially happen too often and could be spammy?

Once we agree on which approach is better, I would like to implement this myself.

kwzrd:
Closes #390 

The issue gives a fairly though explanation. Implementation is simple, as we already have a method for dispatching embeds to the `#dev-log` channel. In 6af47f8, we first ensure that the cache is ready before grabbing the devlog channel instance.

image

The author field is always populated with `SeasonalBot` - the seasonal nickname isn't used.

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -44,6 +44,7 @@ class SeasonalBot(commands.Bot):
         self.http_session = ClientSession(
             connector=TCPConnector(resolver=AsyncResolver(), family=socket.AF_INET)
         )
+        self.greeting_task = self.loop.create_task(self.send_log("SeasonalBot", "Connected!"))
 
     @property
     def member(self) -> Optional[discord.Member]:
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -44,7 +44,7 @@ class SeasonalBot(commands.Bot):
         self.http_session = ClientSession(
             connector=TCPConnector(resolver=AsyncResolver(), family=socket.AF_INET)
         )
-        self.greeting_task = self.loop.create_task(self.send_log("SeasonalBot", "Connected!"))
+        self.loop.create_task(self.send_log("SeasonalBot", "Connected!"))
 
     @property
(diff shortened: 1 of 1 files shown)


bot/bot.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, **kwargs):
         self.http_session = ClientSession(
             connector=TCPConnector(resolver=AsyncResolver(), family=socket.AF_INET)
         )
+        self.greeting_task = self.loop.create_task(self.send_log("SeasonalBot", "Connected!"))
```
scragly, line 47: There's no need to store the task; it's never referenced later and it's a fire and forget type of thing.

kwzrd: I have a tendency to assign in case something wants to check in on the task later on, but I guess in this case it really isn't necessary. I will remove it.

PR #396, opened 2020-04-06
APOD and Minesweeper command fixes

### LINKED ISSUE #387: Minesweeper: KeyError in reveal command when game no longer exists
Sentry Issue: [SEASONALBOT-A](https://sentry.io/organizations/python-discord/issues/1588105645/?referrer=github_integration)

```
KeyError: <redacted user ID>
  File "discord/ext/commands/core.py", line 83, in wrapped
  File "bot/seasons/evergreen/minesweeper.py", line 256, in reveal_command
    game = self.games[ctx.author.id]

Unhandled command error: <redacted user ID>
```

### LINKED ISSUE #388: KeyError: 'date' in apod command
When passed a date not in the range of dates supported by the NASA Astronomy picture of the day a KeyError is raised.

We should confirm that the date passed is between June 16th 1995 and the current date.

Additionally, to be more fault tolerant we should handle API errors completely, I'll attach an example of an error response for people to work with should they want to take this issue on.

Sentry Issue: [SEASONALBOT-B](https://sentry.io/organizations/python-discord/issues/1589477958/?referrer=github_integration)

```
KeyError: 'date'
  File "discord/ext/commands/core.py", line 83, in wrapped
  File "bot/seasons/evergreen/space.py", line 86, in apod
    f"Astronomy Picture of the Day - {result['date']}",

Unhandled command error: 'date'
```

ks129:
## Relevant Issues

Closes #388 
Closes #387 

## Description

Added a try-except block to Minesweeper and date check to APOD command.

## Did you:

### DIFF
--- a/bot/exts/evergreen/minesweeper.py
+++ b/bot/exts/evergreen/minesweeper.py
@@ -253,7 +253,11 @@ class Minesweeper(commands.Cog):
     @minesweeper_group.command(name="reveal")
     async def reveal_command(self, ctx: commands.Context, *coordinates: CoordinateConverter) -> None:
         """Reveal multiple cells."""
-        game = self.games[ctx.author.id]
+        try:
+            game = self.games[ctx.author.id]
+        except KeyError:
+            await ctx.author.send("Game don't exist. Can't reveal.")
+            return
         revealed: GameBoard = game.revealed
         board: GameBoard = game.board
 
--- a/bot/exts/evergreen/minesweeper.py
+++ b/bot/exts/evergreen/minesweeper.py
@@ -29,6 +29,12 @@ MESSAGE_MAPPING = {
 log = logging.getLogger(__name__)
 
 
+class UserNotPlayingError(Exception):
+    """Will raised when user try to use game commands when not playing."""
+
+    pass
+
+
 class CoordinateConverter(commands.Converter):
     """Converter for Coordinates."""
 
--- a/bot/exts/evergreen/minesweeper.py
+++ b/bot/exts/evergreen/minesweeper.py
@@ -253,11 +253,7 @@ class Minesweeper(commands.Cog):
     @minesweeper_group.command(name="reveal")
     async def reveal_command(self, ctx: commands.Context, *coordinates: CoordinateConverter) -> None:
         """Reveal multiple cells."""
-        try:
-            game = self.games[ctx.author.id]
-        except KeyError:
-            await ctx.author.send("Game don't exist. Can't reveal.")
-            return
+        game = self.games[ctx.author.id]
         revealed: GameBoard = game.revealed
         board: GameBoard = game.board
 
@@ -281,6 +277,13 @@ class Minesweeper(commands.Cog):
         if game.activated_on_server:
             await game.chat_msg.edit(content=new_msg)
         del self.games[ctx.author.id]
+
+    @reveal_command.error
+    @end_command.error
+    async def keyerror_handler(self, ctx: commands.Context, error: typing.Any) -> None:
+        """Handle `KeyError` that is raised in reveal or end command when getting game."""
+        if isinstance(error, KeyError):
+            await ctx.send("Game don't exist.")
 
 
 def setup(bot: commands.Bot) -> None:
--- a/bot/exts/evergreen/minesweeper.py
+++ b/bot/exts/evergreen/minesweeper.py
@@ -182,6 +182,8 @@ class Minesweeper(commands.Cog):
     @minesweeper_group.command(name="flag")
     async def flag_command(self, ctx: commands.Context, *coordinates: CoordinateConverter) -> None:
         """Place multiple flags on the board."""
+        if ctx.author.id not in self.games:
+            raise UserNotPlayingError
         board: GameBoard = self.games[ctx.author.id].revealed
         for x, y in coordinates:
             if board[y][x] == "hidden":
@@ -259,6 +261,8 @@ class Minesweeper(commands.Cog):
     @minesweeper_group.command(name="reveal")
     async def reveal_command(self, ctx: commands.Context, *coordinates: CoordinateConverter) -> None:
         """Reveal multiple cells."""
+        if ctx.author.id not in self.games:
+            raise UserNotPlayingError
         game = self.games[ctx.author.id]
         revealed: GameBoard = game.revealed
         board: GameBoard = game.board
@@ -275,6 +279,8 @@ class Minesweeper(commands.Cog):
     @minesweeper_group.command(name="end")
     async def end_command(self, ctx: commands.Context) -> None:
         """End your current game."""
+        if ctx.author.id not in self.games:
+            raise UserNotPlayingError
         game = self.games[ctx.author.id]
         game.revealed = game.board
         await self.update_boards(ctx)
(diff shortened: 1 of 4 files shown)


bot/exts/evergreen/space.py
```diff
@@ -74,7 +74,12 @@ def cog_unload(self) -> None:
...
-                params["date"] = datetime.strptime(date, "%Y-%m-%d").date().isoformat()
+                now = datetime.now().date()
+                apod_date = datetime.strptime(date, "%Y-%m-%d").date()
+                if datetime(1995, 6, 16).date() > apod_date or now < apod_date:
+                    await ctx.send(f"Date must be in range 16th June 1995 and {now.strftime('%dth %B %Y')}.")
+                    return
+                params["date"] = apod_date.isoformat()
             except ValueError:
                 await ctx.send(f"Invalid date {date}. Please make sure your date is in format YYYY-MM-DD.")
                 return
```
ikuyarihS, line 85: The `ValueError` can only be raised by
```py
apod_date = datetime.strptime(date, "%Y-%m-%d").date()
```
right? In that case, you can move the try block up and put it there only, since the rest will not raise `ValueError`, so something like this
```py
try:
    apod_date = datetime.strptime(date, "%Y-%m-%d").date()
except ValueError:
    await ctx.send(f"Invalid date {date}. Please make sure your date is in format YYYY-MM-DD.")
    return

now = datetime.now().date()

if datetime(1995, 6, 16).date() > apod_date or now < apod_date:
    await ctx.send(f"Date must be in range 16th June 1995 and {now.strftime('%dth %B %Y')}.")
    return

params["date"] = apod_date.isoformat()
```

bot/exts/evergreen/space.py
```diff
@@ -74,11 +74,18 @@ def cog_unload(self) -> None:
...
         if date:
             try:
-                params["date"] = datetime.strptime(date, "%Y-%m-%d").date().isoformat()
+                apod_date = datetime.strptime(date, "%Y-%m-%d").date()
             except ValueError:
                 await ctx.send(f"Invalid date {date}. Please make sure your date is in format YYYY-MM-DD.")
                 return
 
+            now = datetime.now().date()
+            if datetime(1995, 6, 16).date() > apod_date or now < apod_date:
```
kwzrd, line 83: Why not just build a date object directly?
```python
date(1995, 6, 16)
```
I'd probably make this specific date a constant and perhaps explain why it's needed, but it's not necessary.

bot/exts/evergreen/space.py
```diff
@@ -74,11 +74,18 @@ def cog_unload(self) -> None:
...
+                await ctx.send(f"Date must be in range 16th June 1995 and {now.strftime('%dth %B %Y')}.")
```
kwzrd, line 84: * If I'm not mistaken, this instructs the user to provide the date in an unsupported format.
* The current format string produces `07th April 2020` today.
* I think `Date must be between 16th June 1995 and today` would be sufficient.

bot/exts/evergreen/space.py
kwzrd: If this is made a constant (or at least a local variable), then it can be referenced in the error message for requests exceeding the allowed range.

ks129:
@kwzrd I used d.py error handler now.

bot/exts/evergreen/minesweeper.py
```diff
@@ -278,6 +278,13 @@ def reveal_zeros(self, revealed: GameBoard, board: GameBoard, x: int, y: int) ->
             await game.chat_msg.edit(content=new_msg)
         del self.games[ctx.author.id]
 
+    @reveal_command.error
+    @end_command.error
+    async def keyerror_handler(self, ctx: commands.Context, error: typing.Any) -> None:
+        """Handle `KeyError` that is raised in reveal or end command when getting game."""
+        if isinstance(error, KeyError):
+            await ctx.send("Game don't exist.")
```
kwzrd, line 286: This is an interesting approach that I did not consider, but I like it a lot. There are, however, a few changes needed.
* There are still places in this cog where the look-up is unsafe and fails on a `KeyError`. The `flag_command` is an example. Please review the **entire** cog to ensure that this PR addresses **all** unsafe look-ups.
* I'm a little bit on the edge about blanket-catching *any* `KeyError` that falls out of the commands. While it's probably ok, my personal approach would have been to define a custom `UserNotPlayingError` exception subclass, and handle specifically only that:
```python
if ctx.author.id not in self.games:
    raise UserNotPlayingError
```
* The type annotation for `error` is unnecessarily lax. At the very least, it should be annotated as an `Exception`.

bot/exts/evergreen/space.py
```diff
@@ -74,11 +76,18 @@ def cog_unload(self) -> None:
...
-                params["date"] = datetime.strptime(date, "%Y-%m-%d").date().isoformat()
+                apod_date = datetime.strptime(date, "%Y-%m-%d").date()
             except ValueError:
                 await ctx.send(f"Invalid date {date}. Please make sure your date is in format YYYY-MM-DD.")
                 return
 
+            now = datetime.now().date()
+            if APOD_MIN_DATE > apod_date or now < apod_date:
+                await ctx.send(f"Date must be between {APOD_MIN_DATE.isoformat()} and {now.isoformat()} (today).")
+                return
```
kwzrd, line 87: This is much better!

ks129:
@kwzrd I changed all unsafe lookups in commands. I didn't change it in helper functions, because commands already check these and don't let reach code to there.

bot/exts/evergreen/minesweeper.py
```diff
@@ -278,6 +290,14 @@ def reveal_zeros(self, revealed: GameBoard, board: GameBoard, x: int, y: int) ->
             await game.chat_msg.edit(content=new_msg)
         del self.games[ctx.author.id]
 
+    @reveal_command.error
+    @end_command.error
+    @flag_command.error
+    async def user_not_playing_handler(self, ctx: commands.Context, error: Exception) -> None:
+        """Handle `KeyError` that is raised in reveal or end command when getting game."""
+        if isinstance(error, UserNotPlayingError):
+            await ctx.send("Game don't exist.")
```
kwzrd, line 299: Exceptions raised within a `d.py` command will be wrapped in [`CommandInvokeError`](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.CommandInvokeError) which carries on `original` attribute with the specific exception. As a result, this doesn't catch the `UserNotPlayingError`, as `error` isn't an instance thereof - `error.original` is.

The error handler cog gets around this via [safely getting the original](https://github.com/python-discord/seasonalbot/blob/38d0098b0f2dce38e032be235b6dd2f1e2109b4a/bot/exts/evergreen/error_handler.py#L49) before processing it.

Additionally, as *all* errors from the commands will end up here, we should also make sure they aren't silenced. Let's [log the exception](https://github.com/python-discord/seasonalbot/blob/38d0098b0f2dce38e032be235b6dd2f1e2109b4a/bot/exts/evergreen/error_handler.py#L124) just like the error handler does.

Finally, since `user_not_playing_handler` will receive all errors, not just `UserNotPlayingError`, it would be better to just name it `error_handler`, or similar. The docstring still mentions `KeyError` which no longer holds.

---

An alternative to re-implementing the error handler cog's logic here, we could also consider moving our logic there. If you decide to do this, please move the `UserNotPlayingError` definition to `bot.utils.exceptions` - importing Exceptions classes across `d.py` extension modules is dangerous and can lead to unexpected behaviour, as the definition may be executed multiple times.

PR #403, opened 2020-05-04
Handle staff users and DMs in the halloweenify cog

### LINKED ISSUE #402: Handle potential errors in halloweenify command
The `halloweenify` cog will generate a Halloween-themed nickname for the invoking user, and then attempt to set it as a guild nickname.

https://github.com/python-discord/seasonalbot/blob/b45272aa62859abedc1ccc1836f0a61935394e87/bot/exts/halloween/halloweenify.py#L44

It does not currently handle the following situations.

## Lacking permissions

When invoked by a staff member, the bot lacks the required permissions, and a `discord.Forbidden` exception silently kills the command.

The exception should be caught & handled by informing the user that their nickname cannot be changed by the bot, but they can still change it themselves - the bot already also shows an embed with the new name (if it doesn't die before the line is reached).

## No nicknames in DMs

When invoked in DMs, the cog will also fail as `ctx.author` is not an instance of `discord.Member` and has no `edit` method. In such a case, we can still show the generated nickname and inform the user that if they invoke the command in-server, they can have their nickname changed automatically.

While writing the code to handle both situations shouldn't be difficult, a good solution will incorporate the needed information into the response embed in a natural way, with UX in mind.

---

Akarys42:
## Relevant Issues

Closes #402

## Description
 
This PR adds proper fallback messages when the halloweenify command is invoked by a member staff or through DM.

## Reasoning

The bot doesn't have enough rights to modify staff nicknames, and can't change it when it is invoked through DM neither.

## Screenshots

image
In-server usage.

image
Staff uage.

image
In DM usage.

## Did you:

### DIFF
--- a/bot/exts/halloween/halloweenify.py
+++ b/bot/exts/halloween/halloweenify.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from random import choice
 
 import discord
+from discord.errors import Forbidden
 from discord.ext import commands
 from discord.ext.commands.cooldowns import BucketType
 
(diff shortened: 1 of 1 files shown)


bot/exts/halloween/halloweenify.py
```diff
@@ -37,11 +38,25 @@ async def halloweenify(self, ctx: commands.Context) -> None:
...
+                else:   # The command has been invoked in DM
+                    embed.description += (
+                        f"Your new nickname should be: \n :ghost: **{nickname}** :jack_o_lantern: \n\n"
+                        f"Feel free to change it yourself, or invoke the command again inside the server."
+                    )
+
+            except Forbidden:   # The bot doesn't have enough permission
+                embed.description += (
+                    f"Your new nickname should be: \n :ghost: **{nickname}** :jack_o_lantern: \n\n"
+                    f"Although it looks like I can't change it myself, but feel free to change it yourself."
```
kwzrd, line 58: `Although` or `but` should be removed to make this read well. Overall this could be written better, e.g.:
```
It looks like I cannot change your name, but feel free to change it yourself.
```

bot/exts/halloween/halloweenify.py
```diff
@@ -37,11 +38,25 @@ async def halloweenify(self, ctx: commands.Context) -> None:
...
+            try:
+                if isinstance(ctx.author, discord.Member):
+                    await ctx.author.edit(nick=nickname)
+                    embed.description += f"Your new nickname will be: \n:ghost: **{nickname}** :jack_o_lantern:"
+
+                else:   # The command has been invoked in DM
+                    embed.description += (
+                        f"Your new nickname should be: \n :ghost: **{nickname}** :jack_o_lantern: \n\n"
+                        f"Feel free to change it yourself, or invoke the command again inside the server."
+                    )
```
kwzrd, line 53: I'd prefer if only the line that we expect to raise was in the `try` block, e.g.:
```python
if isinstance(...):
    try:
        await ctx.author.edit(...)
    except Forbidden:
        # append no permissions message
    else:
        # append success message
else:
    # append DM message
```
This way the logic for both cases (guild / DM) is clearly separated, and we see exactly which lines may raise.

I like the comment on line 49 - helps see why we're branching there.

PR #408, opened 2020-05-17
Use input hashing instead of randomness inside 8ball

Akarys42:
## Description

This PR make the `.8ball` command use hashing instead of `random.choice()`

## Reasoning

This makes the outcome always the same for the same question, so a user cannot brute-force the 8ball until they obtain the answer they want :grin:. 

## Did you:

### DIFF
--- a/bot/exts/evergreen/magic_8ball.py
+++ b/bot/exts/evergreen/magic_8ball.py
@@ -1,6 +1,6 @@
 import json
 import logging
-import random
+from hashlib import sha1
 from pathlib import Path
 
 from discord.ext import commands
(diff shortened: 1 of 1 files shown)


sco1:
> so a user cannot brute-force the 8ball until they obtain the answer they want

This is pretty much how I and everyone I know has ever used an actual magic 8ball.

SebastiaanZ:
I think the randomness and the chance of getting a different answer for the same question is part of the 8ball experience. It's just a random answer generator, which is often implemented purely in hardware form using a container of fluid and a floating die. Brute-forcing your answer has been a feature since its introduction in 1950.

As such, I don't think we have a desire to make ours a non-standard implementation that deterministically tries to divine the Truth.

PR #410, opened 2020-05-17
Set ayncio event loop to Selector on windows.

### LINKED ISSUE #409: Specify event loop on Windows
Sister issue from `bot` repo: https://github.com/python-discord/bot/issues/894

With Python 3.8, the default asyncio event loop on Windows is now the [ProactorEventLoop](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.ProactorEventLoop). One of our dependencies, `aiodns`, doesn't like that, as reported today by one of our contributors:

loop

For `bot`, we've already addressed the issue by explicitly selecting a supported event loop: https://github.com/python-discord/bot/pull/903/.

This is an easy fix that should probably be applied somewhat soon so that future contributors do not run into this issue. Of course ideally we'd have this implemented by someone who develops or is capable of developing on Windows.

---

Numerlor:
## Relevant Issues
Fixes #409 

## Description
Sets the event loop to Selector to prevent bot crashes on windows when an aiohttp request is made, as described in #409 
Same implementation as in https://github.com/python-discord/bot/pull/903

### DIFF
--- a/bot/__init__.py
+++ b/bot/__init__.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import logging.handlers
 import os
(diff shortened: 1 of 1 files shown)


kwzrd:
Thanks for the effort everyone!

PR #413, opened 2020-05-20
Encoding bug fix

### LINKED ISSUE #334: Cleanup: Add explicit encoding on open() calls
A bunch of files in this repo are opening JSONs without an explicit encoding. This can cause errors when running SeasonalBot locally, if:
- On Windows (which applies the cp-1252 encoding by default)
- The file contains some character which confuses cp-1252
- The user does not have the PYTHONUTF8 environment variable set to 1.

I don't know how rare an actual error is (I suppose most jsons we use are ascii-friendly), but the trivia_quiz cog wouldn't load for me.

Regardless, why not patch them all up?

### LINKED ISSUE #411: Files without an encoding specified may crash bot
On windows (possibly some other OSs), non utf8 default encodings are used which produces `UnicodeDecodeError`s when files with characters not included in the default encodings are read.

The main issue comes from `bot\resources\evergreen\trivia_quiz.json` which is always ran on bot startup, and crashes with the latin1 encoding (english windows); other files incompatible with the encoding are opened with utf8 specified.

Here's a small list of where the encoding should be specified, but may be incomplete as it was filtered manually https://paste.fuelrats.com/uxuyebomev.txt

The https://github.com/rayjolt/flake8-file-encoding flake8 plugin was suggested in #dev-contrib to mitigate this issue, but it does not check Pathlib's opens so I don't think it's worth adding to the linting process

- [X] Anyone can implement the bug fix

jodth07:
## Relevant Issues
- Cleanup: Add explicit encoding on open() calls #334
- Files without an encoding specified may crash bot #411

Closes #334 
Closes #411 

## Description
Added explicit encoding to `utf8` for cogs that open json or other text files.

## Reasoning
Though this did not cause errors in Unix based OS, it was causing errors when loading some of those files in windows OS.

## Additional Details
We may require explicit encoding going forward for other cogs being added to the bots that needs to read in a file.

## Did you:

### DIFF
--- a/bot/exts/christmas/adventofcode.py
+++ b/bot/exts/christmas/adventofcode.py
@@ -429,7 +429,7 @@ class AdventOfCode(commands.Cog):
 
     def _build_about_embed(self) -> discord.Embed:
         """Build and return the informational "About AoC" embed from the resources file."""
-        with self.about_aoc_filepath.open("r") as f:
+        with self.about_aoc_filepath.open("r", encoding="utf8") as f:
             embed_fields = json.load(f)
 
         about_embed = discord.Embed(title=self._base_url, colour=Colours.soft_green, url=self._base_url)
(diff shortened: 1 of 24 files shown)


kwzrd:
Thanks, I'll try to review this soon.

jodth07:
Sounds good, I will be updating this

Issue #417, opened 2020-06-06
Enable Sentry aiohttp integration


lemonsaurus:
Sentry has an integration for aiohttp, which will give us more relevant data collection for this repo to Sentry. I'll add this today.

I'll bump the sentry-sdk version up to 0.14.4 while I'm at it, too.

MarkKoz:
As per https://github.com/python-discord/seasonalbot/pull/418#issuecomment-640119454, the integration is for an aiohttp server, which we don't use.

PR #418, opened 2020-06-06
Add aiohttp integration to Sentry init and bump lockfile versions

lemonsaurus:
This PR makes two small changes:
- Adds the AioHttpIntegration to the `sentry_sdk.init` call
- Bumps all versions in the lockfile.

We're bumping the versions because Sentry is complaining that we're not using the latest version of its sdk.

I've tested this commit, and the bot still runs fine, but I didn't test it very thoroughly. Feel free to do a quick feature review of this if you're reviewing.

Closes: #417

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -43,11 +43,11 @@
         },
         "arrow": {
             "hashes": [
-                "sha256:5390e464e2c5f76971b60ffa7ee29c598c7501a294bc9f5e6dadcb251a5d027b",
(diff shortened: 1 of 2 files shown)


j03b:
I've looked around and found the following:
- [original issue for aiohttp support](https://github.com/getsentry/sentry-python/issues/89)
- [implementing PR](https://github.com/getsentry/sentry-python/pull/189)

It's definitely built for aiohttp server, mentioning things like REMOTE_ADDR, all the tests of it are from a server side so I don't think this is necessary.

This PR should just bump the versions.

lemonsaurus:
Maybe we should just close it then and bump them when it's more urgent to do so.

PR #421, opened 2020-06-25
Caesar cipher command

### LINKED ISSUE #420: Caesar cipher command
## Description
`.caesarcipher` or its alias `.caesar`, can be used to cipher the given text, given an integer offset.

## Reasoning
I've recently noticed that there was no caesar cipher command specifically in the [Fun](https://github.com/python-discord/seasonalbot/blob/bb30d2904b4a3273d4b10dc9548c542d65f0c379/bot/exts/evergreen/fun.py#L29) cog of the seasonalbot, and I think it would be a fair addition.

## Proposed Implementation
The command will be implemented in the [Fun](https://github.com/python-discord/seasonalbot/blob/bb30d2904b4a3273d4b10dc9548c542d65f0c379/bot/exts/evergreen/fun.py#L29) cog, and as such, will behave similarly to commands such as `.uwu` and `.randomcase`, as well as supporting message IDs and links.

image

PureFunctor:
## Relevant Issues
Closes #420.

## Description
This adds a `.caesarcipher` command group with `decrypt`, `encrypt`, and `info` subcommands in the [Fun](https://github.com/python-discord/seasonalbot/blob/bb30d2904b4a3273d4b10dc9548c542d65f0c379/bot/exts/evergreen/fun.py#L29) cog. The `decrypt` and `encrypt` subcommands also support message IDs and links similarly to other [Fun](https://github.com/python-discord/seasonalbot/blob/bb30d2904b4a3273d4b10dc9548c542d65f0c379/bot/exts/evergreen/fun.py#L29) cog commands such as `.uwu` and `.randomcase`.

## Reasoning
The `decrypt` and `encrypt` subcommands should provide a concise way of using the ciphering a message, with `info` adding good educational value to the feature.

## Screenshots
image

## Additional Details
In order to support message IDs and links for the command, the `_caesar_cipher` method was patterned against the [uwu](https://github.com/python-discord/seasonalbot/blob/bb30d2904b4a3273d4b10dc9548c542d65f0c379/bot/exts/evergreen/fun.py#L49) and [random case](https://github.com/python-discord/seasonalbot/blob/bb30d2904b4a3273d4b10dc9548c542d65f0c379/bot/exts/evergreen/fun.py#L69) commands. Maybe this feature can be abstracted to a generalized function to avoid redundancy/for future text-related commands as well?

## Did you:


bot/exts/evergreen/fun.py
```diff
@@ -87,6 +88,91 @@ def conversion_func(text: str) -> str:
...
         await ctx.send(content=converted_text, embed=embed)
 
+    @commands.group(name="caesarcipher", aliases=("caesar",))
+    async def caesarcipher_group(self, ctx: Context) -> None:
+        """
+        Translates a message using the Caesar Cipher.
+
+        See `decrpyt`, `encrypt`, and `info` subcommands.
+        """
+        if ctx.invoked_subcommand is None:
```
ks129, line 98: I think better is use `invoke_without_command=True` in `command.group` decorator.

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +88,91 @@ def conversion_func(text: str) -> str:
...
+            await self.bot.get_cog("Help").new_help(ctx, "caesarcipher")
```
ks129, line 99:
```suggestion
            await ctx.invoke(self.bot.get_command("help"), "caesarcipher")
```

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +88,91 @@ def conversion_func(text: str) -> str:
...
+
+    @caesarcipher_group.command(name="info")
+    async def caesarcipher_info(self, ctx: Context) -> None:
+        """Information about the Caesar Cipher."""
+        with open("bot\\resources\\evergreen\\caesar_info.json", "r") as f:
+            data = json.load(f)
```
ks129, line 105: It's not a good idea to load this on every use. Better do this in `__init__`.

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +88,91 @@ def conversion_func(text: str) -> str:
...
+        if embed is not None:
+            embed = Fun._convert_embed(conversion_func, embed)
+        converted_text = conversion_func(text)
+        # Don't put >>> if only embed present
+        if converted_text:
+            converted_text = f">>> {converted_text.lstrip('> ')}"
+        await ctx.send(content=converted_text, embed=embed)
+
+    @caesarcipher_group.command(name="encrypt", aliases=("rightshift", "rshift"))
+    async def caesarcipher_encrypt(self, ctx: Context, offset: int, *, text: str) -> None:
```
ks129, line 149: I think better is use `Union[str, Message]` converter instead converting this manually. Same in L163.

bot/exts/evergreen/fun.py
kwzrd: I don't mind the current way, it's a fairly common pattern.

The problem with `invoke_without_command` afaik is that it allows sub-commands to bypass checks on the group cmd, which can be unexpected.

PureFunctor: So far I've only changed how the help embed is invoked when no subcommands are passed; I have to ask though:

> allows subcommands to bypass checks

would there be any merit to using `invoke_without_command` given that I'm only invoking the help embed for the command group?

kwzrd: In this case I don't think it makes much of a difference (please correct me if I'm wrong).

Just to explain what I meant: if we use `invoke_without_command=False` (default), and apply some kind of a check, such as `with_role` from our utils, this check will also apply to *all* subcommands, which is generally what we want. This means that if you lock the group to mods only, all subcommands will also prevent non-mod users from invoking them.

With `invoke_without_command=True`, the group callback (which contains the check) will be circumvented by all subcommands, which means that they bypass the check. This can be a little surprising and unexpected.

This all is not too relevant to this case, since the group has no checks (and isn't likely to have in the future), but it's the reason why I don't like using the kwarg too much, as this behaviour is imo somewhat unexpected (although documented). So I think not having the kwarg and doing checking `ctx.invoked_subcommand` to see if the help should be invoked is perfectly fine.

But if you like @ks129's suggestion there's nothing inherently wrong with it, it will save you a line of code, since the group cmd will never run if a subcommand was invoked, it can just send the help.

bot/exts/evergreen/fun.py
```diff
@@ -32,6 +33,9 @@ class Fun(Cog):
     def __init__(self, bot: Bot) -> None:
         self.bot = bot
 
+        with open("bot\\resources\\evergreen\\caesar_info.json", "r") as f:
+            self._caesar_cipher_embed = json.load(f)
```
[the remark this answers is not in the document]
Numerlor: I think `Path("bot/resources/evergreen/caesar_info.json")` looks better, and pathlib will normalize them in the same way

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +92,90 @@ def conversion_func(text: str) -> str:
...
+    async def _caesar_cipher(ctx: Context, offset: int, msg: Union[Message, str]) -> None:
+        """
+        Given an integer `offset`, translates and sends the given `text`.
+
+        A positive `offset` will cause the letters to shift right, while
+        a negative `offset` will cause the letters to shift left.
+
+        Also accepts a valid Discord Message ID or link.
+        """
+        def caesar_func(text: str) -> str:
```
PureFunctor, line 123: I've recently noticed this has an incorrect annotation for its return type. Should this be corrected to `Generator[str, None, None]`?

bot/exts/evergreen/fun.py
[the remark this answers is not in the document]
kwzrd: Perhaps, matter of taste. 🦆

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +92,90 @@ def conversion_func(text: str) -> str:
             converted_text = f">>> {converted_text.lstrip('> ')}"
         await ctx.send(content=converted_text, embed=embed)
 
+    @commands.group(name="caesarcipher", aliases=("caesar",))
```
ks129, line 95: I suggest `cc` alias.

bot/exts/evergreen/fun.py
Numerlor: I think `Iterator[str]` ought be enough instead of the whole Generator typehint when it only yields

kwzrd: I've seen people annotate generators using just the yield type (such as here), but yeah using `Iterator[str]` would probably be best. The [docs](https://docs.python.org/3/library/typing.html#typing.Generator) mention it as a valid alternative to `Generator[str, None, None]`.

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +92,90 @@ def conversion_func(text: str) -> str:
...
+        """Information about the Caesar Cipher."""
+        embed = Embed.from_dict(self._caesar_cipher_embed)
+        embed.colour = Colours.dark_green
+
+        await ctx.send(embed=embed)
+
+    @staticmethod
+    async def _caesar_cipher(ctx: Context, offset: int, msg: Union[Message, str]) -> None:
+        """
+        Given an integer `offset`, translates and sends the given `text`.
```
Den4200, line 116:
```suggestion
        Given an integer `offset`, translates and sends the given `msg`.
```I'm guessing you were referring to the `msg` parameter here, instead of `text`.

File: bot/exts/evergreen/fun.py [hunk 9]
```diff
@@ -87,6 +92,90 @@ def conversion_func(text: str) -> str:
...
+                yield chr((ord(char) - case_start + offset) % 26 + case_start)
+
+        def conversion_func(text: str) -> str:
+            """Encrypts the given string using the Caesar Cipher."""
+            return "".join(caesar_func(text))
+
+        is_message = isinstance(msg, Message)
+
+        text = msg.content if is_message else msg
+        embed = msg.embeds[0] if is_message and msg.embeds else None
```
Den4200, line 139: You can use `Fun._get_text_and_embed` instead here, similar to how `Fun.uwu_command` does it.

PureFunctor: I figured that it would be a bit redundant to use said method to manually convert the `msg` that's already handled by the `Union[Message, str]` converter. I

Den4200: Hmm.. I see. It seems pretty redundant doing it either way then. Personally, I think I'd stick with using `Fun._get_text_and_embed` and just have the type annotation as `str`, so it's consistent with the other methods.

It's up to you though, you can do what you see fit.

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +92,93 @@ def conversion_func(text: str) -> str:
...
+    async def caesarcipher_encrypt(self, ctx: Context, offset: int, *, msg: Union[Message, str]) -> None:
+        """
+        Given a positive integer `offset`, encrypt the given `msg`.
+
+        Performs a right shift of the letters in the message.
+
+        Also accepts a valid Discord Message ID or link.
+        """
+        if offset < 0:
+            await ctx.send(":no_entry: Cannot use a negative offset.")
```
ks129, line 164: Because this repeat on L177, I think maybe is better to move it to `_caesar_cipher`.

PureFunctor: `caesarcipher_decrypt` passes a negated `offset` to the `_caesar_cipher` method to perform a left shift though, because of how I've written the formula for the `true_offset` in L131.

PureFunctor: I've decided to refactor the translation methods as this was a rather trivial change. Thanks for the suggestion.

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +92,93 @@ def conversion_func(text: str) -> str:
...
+    @caesarcipher_group.command(name="encrypt", aliases=("rightshift", "rshift", "enc",))
+    async def caesarcipher_encrypt(self, ctx: Context, offset: int, *, msg: Union[Message, str]) -> None:
+        """
+        Given a positive integer `offset`, encrypt the given `msg`.
+
+        Performs a right shift of the letters in the message.
+
+        Also accepts a valid Discord Message ID or link.
+        """
+        await self._caesar_cipher(ctx, offset, msg, False)
```
kwzrd, line 169: I'd prefer if the last argument was passed by keyword, otherwise you won't know what the bool does.

PureFunctor: I'll go ahead and modify that 👍

bot/exts/evergreen/fun.py
```diff
@@ -87,6 +92,93 @@ def conversion_func(text: str) -> str:
...
+                    continue
+
+                case_start = 65 if char.isupper() else 97
+                true_offset = (ord(char) - case_start + offset) % 26
+
+                yield chr(case_start + true_offset)
+
+        def conversion_func(text: str) -> str:
+            """Encrypts the given string using the Caesar Cipher."""
+            return "".join(caesar_func(text))
```
ikuyarihS, line 143: Is there a reason why these are defined inside the command? From the look of it, you can move it outside as well to prevent them being defined everytime the command is run, declaring them as `@staticmethod`

PureFunctor: I've initially decided to define these inside of the function to be consistent with the form of the other commands inside of the `Fun` cog. I could change them into `@staticmethod`s but I'm not sure whether making them standalone top-level functions would avoid polluting the cog.

ikuyarihS: I would suggest moving them outside, in case we want to write unittests for these. Naming these with a `_` in front of the name to signify that these are to be used internally might be able to help, otherwise you can always move them outside in their own class and just call it inside the Cog.

PureFunctor: I'll go ahead and do that :+1:

PR #424, opened 2020-07-11
Created the statuscat command.

### LINKED ISSUE #423: Cat status codes
## Description

A command such as `.statuscat` would be implemented to visually show what different status codes are, with cats.

## Reasoning

For people who may be confused by status codes or need a reference when working on their projects, [https.cat](https://http.cat/) is the perfect resource. Humans can quickly either what the status code from the output of their program is trying to tell them, or use it as information when creating their website.

## Proposed Implementation

An embed with an image of a cat within. The image provided by [https.cat](https://http.cat/) has the status code and short description of what it means.
image
First link within the description of the embed is the link to the image, while the second is linked to the main website.
No need for API requests, as an image can be fetched by inserting the status code into an f-string. Here's a proposed chunk of code, which will be in a cog:
```python
@commands.command(aliases=['statuscat'])
async def http_cat(self, ctx, code: int):
    url = f'https://http.cat/{code}.jpg'
    embed = discord.Embed(
        description=(
            f'[`{code}`]({url})'
            '`Cat retrieved from:`[`http.cat`](https://http.cat/)'
        )
    )

    embed.set_image(url=url)

    await ctx.send(embed=embed)
```

## Additional Details

Not everyone likes cats, and maybe dogs should be inserted into the bot at a later time. This would be also a good opportunity to show non-cat people the ability of cats to be fantastic and/or odd.

Pros:
- Cats.

Cons:
- Dogs haven't been merged yet.

Xithrius:
## Relevant Issues

Closes #423 

## Description

I've added a command called `"http_cat"` with an alias `"statuscat"`. It recieves an integer from a user, then processes it through `http.HTTPStatus` to see if the status code actually exists. If it does, the URL will be valid and a picture will be displayed within an embed. If not, the embed won't show an image and instead it will tell you that the status code provided does not exist.

## Reasoning

Instead of using `aiohttp` with a `ClientSession` to get information from the URL, I simply used the URL as the image within the embed.

## Screenshots

image

## Additional notes:
When using a browser you can type in any code or string into the `code` section of the f-string `embed.set_image(url=f'https://http.cat/{code}.jpg')` and it will give you a 404 image. For whatever reason the image will not be put into the embed. Setting `code` to 404 will work, though.

## Did you:

### DIFF
--- a/bot/exts/evergreen/status_cats.py
+++ b/bot/exts/evergreen/status_cats.py
@@ -1,10 +1,7 @@
-import logging
 from http import HTTPStatus
 
 import discord
 from discord.ext import commands
-
-log = logging.getLogger(__name__)
 
 
 class StatusCats(commands.Cog):
@@ -20,12 +17,15 @@ class StatusCats(commands.Cog):
 
         try:
             HTTPStatus(code)
-            embed.set_image(url=f'https://http.cat/{code}.jpg')
 
         except ValueError:
             embed.set_footer(text='Inputted status code does not exist.')
 
-        await ctx.send(embed=embed)
+        else:
+            embed.set_image(url=f'https://http.cat/{code}.jpg')
+
+        finally:
+            await ctx.send(embed=embed)
 
 
 def setup(bot: commands.Bot) -> None:
--- a/bot/exts/evergreen/status_cats.py
+++ b/bot/exts/evergreen/status_cats.py
@@ -18,7 +18,6 @@ class StatusCats(commands.Cog):
         """Sends an embed with an image of a cat, potraying the status code."""
         embed = discord.Embed(title=f'**Status: {code}**')
 
-
         try:
             HTTPStatus(code)
             embed.set_image(url=f'https://http.cat/{code}.jpg')
--- a/bot/exts/evergreen/status_cats.py
+++ b/bot/exts/evergreen/status_cats.py
@@ -1,0 +1,33 @@
+import logging
+
+from discord.ext import commands
+from http import HTTPStatus
+import discord
+
+log = logging.getLogger(__name__)
+
+
+class StatusCats(commands.Cog):
+    """Commands that give statuses described and visualized by cats."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+
+    @commands.command(aliases=['statuscat'])
+    async def http_cat(self, ctx, code: int) -> None:
+        """Sends an embed with an image of a cat, potraying the status code."""
+        embed = discord.Embed(title=f'**Status: {code}**')
+        embed.set_image(url=f'https://http.cat/{code}.jpg')
+
+        try:
+            HTTPStatus(code)
+
+        except ValueError:
+            embed.set_footer(text='Inputted status code does not exist.')
+
+        await ctx.send(embed=embed)
+
+
+def setup(bot: commands.Bot) -> None:
+    """Load the StatusCats cog."""
+    bot.add_cog(StatusCats(bot))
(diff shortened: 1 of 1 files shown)


bot/exts/evergreen/status_cats.py
```diff
@@ -0,0 +1,33 @@
...
+    async def http_cat(self, ctx: commands.Context, code: int) -> None:
+        """Sends an embed with an image of a cat, potraying the status code."""
+        embed = discord.Embed(title=f'**Status: {code}**')
+        embed.set_image(url=f'https://http.cat/{code}.jpg')
+
+        try:
+            HTTPStatus(code)
+
+        except ValueError:
+            embed.set_footer(text='Inputted status code does not exist.')
```
manusaurio, line 26: An image is set even if the status code doesn't pass the check in the following lines. This causes a funny side effect if you try `.statuscat 0`: [https://http.cat/0.jpg](https://http.cat/0.jpg) responds with a cat that's not even the 404 one, and it's displayed with the error message from the footer. You could move the `.set_image()` to the `try` or put it in an `else` block.
```suggestion


        try:
            HTTPStatus(code)
            embed.set_image(url=f'https://http.cat/{code}.jpg')

        except ValueError:
            embed.set_footer(text='Inputted status code does not exist.')
```

Xithrius: Looks good to me!

bot/exts/evergreen/status_cats.py
```diff
@@ -0,0 +1,33 @@
...
+        embed = discord.Embed(title=f'**Status: {code}**')
+
+        try:
+            HTTPStatus(code)
+            embed.set_image(url=f'https://http.cat/{code}.jpg')
+
+        except ValueError:
+            embed.set_footer(text='Inputted status code does not exist.')
+
+        await ctx.send(embed=embed)
```
Den4200, line 28:
```suggestion
        try:
            HTTPStatus(code)

        except ValueError:
            embed.set_footer(text='Inputted status code does not exist.')

        else:
            embed.set_image(url=f'https://http.cat/{code}.jpg')

        await ctx.send(embed=embed)
```
It's better to separate out this error catching and leave `embed.set_image` in the `else` clause. You could even add a `finally` clause for `ctx.send`, but that's not strictly necessary here.

bot/exts/evergreen/status_cats.py
```diff
@@ -0,0 +1,33 @@
...
+from http import HTTPStatus
+
+import discord
+from discord.ext import commands
+
+log = logging.getLogger(__name__)
+
+
+class StatusCats(commands.Cog):
+    """Commands that give statuses described and visualized by cats."""
```
Den4200, line 11:
```suggestion
    """Commands that give HTTP statuses described and visualized by cats."""
```
Maybe be a bit more specific here saying they're HTTP statuses?

bot/exts/evergreen/status_cats.py
```diff
@@ -0,0 +1,33 @@
+import logging
+from http import HTTPStatus
+
+import discord
+from discord.ext import commands
+
+log = logging.getLogger(__name__)
```
Den4200, line 7: Is this needed? There doesn't seem to be any logging within this cog.

Xithrius: Did that out of habit. Thanks for noticing!

bot/exts/evergreen/status_cats.py
Xithrius: I shall do this, and add the `finally` as well

PR #430, opened 2020-09-04
!wiki command

### LINKED ISSUE #331: Addition of a new command !wiki

Anubhav1603:
"Closes  #331"

when bot able to search
f1

when times up
fail

when error happen
error

It sends value error if user try to send str or float value
it gives list from which user can choose which one they wnt to check 


bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+            else:
+                page.append(a["title"])
+        log.info("appening titles")
+        return page
+
+    @commands.command(name="wikipedia", aliases=["wiki"])
+    async def w_pedia(self, ctx: commands.Context, *, search: str) -> None:
+        """Gives list of item."""
+        final = []
+        s = ''
```
RohanJnr, line 45: Could use better variable names, example: `search_description` instead of `s`?

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+        if title is None:
+            await ctx.send("Sorry, we could not find a wikipedia article using that search term")
+            return
+
+        for i in title:
+            t = i.replace(" ", "_")  # wikipedia uses "_" as spaces
+            final.append(t)
+
+        async with ctx.typing():
+            for index, (a, j) in enumerate(zip(title, final), start=1):
```
RohanJnr, line 61: here too, instead of `a` and `j`, we can use better names. maybe rename `title` -> `titles` and `a`-> `title`.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+                s = s + f'`{index}` [{a}]({WIKIPEDIA_URL.format(title=j)})\n'
```
RohanJnr, line 62: this is the second time we are looping over the `title` list and the final list is not required. we can simply replace `j` with `a.replace(" ", "_")`. (`a` should be renamed as mentioned in a previous comment).
This way, we don't have to loop over the `title` list 2 times and populate the `final` list as done in line 52.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+        async with self.http_session.get(SEARCH_API.format(search_term=search_term)) as response:
+            data = await response.json()
+        page = []
+
+        search_result = data["query"]["search"]
+        if len(search_result) == 0:
+            return None
+
+        # we dont like "may refere to" pages.
+        for a in search_result:
```
RohanJnr, line 33: we can use a better name instead of `a` here.
Maybe rename `search_result` -> `search_results` and `a` -> `search_result`.

RohanJnr (changes requested):
The feature works well. Just a few coding conventions to change here and there.

Also, this is my first review on a pydis PR, would love to hear some feedback on this.

Anubhav1603:
Ok I will make changes according to what u said :) btw this is my first pr also in pydis too

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+            embed = Embed(colour=Color.blue(), title=f"Wikipedia results for `{search}`", description=s_desc)
+            embed.timestamp = datetime.datetime.utcnow()
+            await ctx.send(embed=embed)
+
+        embed = Embed(colour=Color.green(), description="Enter number to choose")
+        msg = await ctx.send(embed=embed)
+
+        try:
+            user = await ctx.bot.wait_for('message', timeout=60.0, check=check)
+            await ctx.send(WIKIPEDIA_URL.format(title=final[int(user.content) - 1]))
```
ikuyarihS, line 72: There will be an error here when entering a negative number or a number that is greater than the amount of results:

```py
  File "/usr/local/lib/python3.8/site-packages/discord/ext/commands/core.py", line 85, in wrapped
  File "/bot/bot/exts/evergreen/wikipedia.py", line 72, in w_pedia
    await ctx.send(WIKIPEDIA_URL.format(title=final[int(user.content) - 1]))
IndexError: list index out of range
```

Also entering 0 will get the 9th result, you will want to add a validation here.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+            if "may refer to" in search_result["snippet"]:
+                pass
+            else:
+                page.append(search_result["title"])
+        log.info("appening titles")
+        return page
+
+    @commands.command(name="wikipedia", aliases=["wiki"])
+    async def w_pedia(self, ctx: commands.Context, *, search: str) -> None:
+        """Gives list of item."""
```
ikuyarihS, line 43: This docstring does not really tell what this function is about, how it is used, I strongly suggest adding more information, otherwise invoking `.help wiki` will yield this

image

A good docstring can yield good help as well, as seen here:
image

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+        # we dont like "may refere to" pages.
+        for search_result in search_results:
+            if "may refer to" in search_result["snippet"]:
+                pass
+            else:
+                page.append(search_result["title"])
+        log.info("appening titles")
+        return page
+
+    @commands.command(name="wikipedia", aliases=["wiki"])
```
ikuyarihS, line 41: You will want to add a check here, a sort of a cooldown if the previous search has not yet completed to prevent spamming `.wiki` command, as well as multiple command listening to the same thing - an index.

An example is when I do `.wiki movie` and `.wiki alice`
image

image

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class WikipediaCog(commands.Cog):
+    """Get info from wikipedia."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+        self.http_session = bot.http_session
+
+    async def search_wikipedia(self, search_term: str) -> Optional[str]:
```
ikuyarihS, line 22:
```suggestion
    async def search_wikipedia(self, search_term: str) -> Optional[List[str]]:
```
You will want to import `List` from `typing` as well.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+        titles = await self.search_wikipedia(search)
+
+        def check(user: Member) -> bool:
+            return user.author.id == ctx.author.id
+
+        if titles is None:
+            await ctx.send("Sorry, we could not find a wikipedia article using that search term")
+            return
+
+        for i in titles:
```
ikuyarihS, line 56: The naming can be much better than `i` - name it `title` for example

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+                pass
+            else:
+                page.append(search_result["title"])
+        log.info("appening titles")
+        return page
+
+    @commands.command(name="wikipedia", aliases=["wiki"])
+    async def w_pedia(self, ctx: commands.Context, *, search: str) -> None:
+        """Gives list of item."""
+        final = []
```
ikuyarihS, line 44: The naming can be improved here, it is really hard to tell what `final` is, what it will hold, and what it will contain.
```suggestion
        titles_no_underscore: List[str] = []
```
is an example you can use, and it will help to differentiate which variable is for what

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,84 @@
...
+        try:
+            user = await ctx.bot.wait_for('message', timeout=60.0, check=check)
+            await ctx.send(WIKIPEDIA_URL.format(title=final[int(user.content) - 1]))
+
+        except asyncio.TimeoutError:
+            embed = Embed(colour=Color.red(), description=f"Time's up {ctx.author.mention}")
+            await msg.edit(embed=embed)
+
+        except ValueError:
+            await ctx.send("I am sorry but that isn't int value")
```
ikuyarihS, line 79: You may want to only show this / end the parsing here if it is not a command, otherwise this can be a problem:

image

The case is when someone is searching via the command for 5 times, they gonna get 4 errors

ikuyarihS (changes requested):
Overall it looks pretty good, but there are some designs that can be improved.

ikuyarihS (changes requested):
The quality and the design has definitely improved, great job!

Anubhav1603:
sorry my commit msg is looking weired, actually my commit failed and then i corrected changes and commited again with new msg and previous commit msg got disappeared 😓

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,108 @@
...
+            data = await response.json()
+        page = []
+
+        search_results = data["query"]["search"]
+        if len(search_results) == 0:
+            return None
+
+        # we dont like "may refere to" pages.
+        for search_result in search_results:
+            log.info("trying to appening titles")
```
ikuyarihS, line 34:
```suggestion
            log.info("trying to append titles")
```
There are some typos in the file similar to this one, let's fix them so they show up properly in the log as well.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,108 @@
...
+            if "may refer to" in search_result["snippet"]:
+                pass
+            else:
+                page.append(search_result["title"])
+        log.info("Finished appening titles")
+        return page
+
+    @commands.cooldown(1, 10, commands.BucketType.user)
+    @commands.command(name="wikipedia", aliases=["wiki"])
+    async def w_pedia(self, ctx: commands.Context, *, search: str) -> None:
```
ikuyarihS, line 44: The name of the command can be improved, something like `wikipedia_search_command` is better than `w_pedia`

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,108 @@
...
+                s_desc += f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})\n'
+            embed = Embed(colour=Color.blue(), title=f"Wikipedia results for `{search}`", description=s_desc)
+            embed.timestamp = datetime.datetime.utcnow()
+            await ctx.send(embed=embed)
+        embed = Embed(colour=Color.green(), description="Enter number to choose")
+        msg = await ctx.send(embed=embed)
+        chances = 0
+        l_of_list = len(titles_no_underscore)  # getting length of list
+
+        while chances <= 3:
```
ikuyarihS, line 74: The `3` number should not be hard-coded and should be moved up to be a variable for the cog.
Even better, you can make it configurable in the config as well, by declaring them in the [bot/constants.py](https://github.com/python-discord/seasonalbot/blob/master/bot/constants.py) for example

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,108 @@
...
+            chances += 1
+            if chances < 3:
+                error_msg = f'You have `{3 - chances}/3` chances left'
+            else:
+                error_msg = 'Please try again by using `.wiki` command'
+            try:
+                user = await ctx.bot.wait_for('message', timeout=60.0, check=check)
```
ikuyarihS, line 81: This naming is not correct, the returned type for a `wait_for('message', ...)` will be a `discord.Message` and **not** a `discord.User` - so you should definitely change the name here from `user` to `message`, and you can further typehint it as well like so
```suggestion
                message: Message = await ctx.bot.wait_for('message', timeout=60.0, check=check)
```
You will need to import `Message` from `discord`

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,108 @@
...
+    @commands.cooldown(1, 10, commands.BucketType.user)
+    @commands.command(name="wikipedia", aliases=["wiki"])
+    async def w_pedia(self, ctx: commands.Context, *, search: str) -> None:
+        """Returns list of your search query from wikipedia."""
+        titles_no_underscore: List[str] = []
+        s_desc = ''
+
+        titles = await self.search_wikipedia(search)
+
+        def check(user: Member) -> bool:
```
ikuyarihS, line 51: The type of the check here is incorrect, the check function for a `wait_for('message', ...)` will have to be a `discord.Message` and not a `discord.Member`

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,108 @@
...
+            try:
+                user = await ctx.bot.wait_for('message', timeout=60.0, check=check)
+                response = await self.bot.get_context(user)
+                if response.command:
+                    return
+                response = int(user.content)
+                if response < 0:
+                    await ctx.send(f"Sorry, but you can't give negative index, {error_msg}")
+                elif response == 0:
+                    await ctx.send(f"Sorry, please give the range between `1` to `{len(l_of_list)}`, {error_msg}")
```
ikuyarihS, line 89: `l_of_list` in this case is an *int* and will raise an exception, as seen here:
```py
  File "/usr/local/lib/python3.8/site-packages/discord/ext/commands/core.py", line 85, in wrapped
  File "/bot/bot/exts/evergreen/wikipedia.py", line 89, in w_pedia
    await ctx.send(f"Sorry, please give the range between `1` to `{len(l_of_list)}`, {error_msg}")
TypeError: object of type 'int' has no len()
```
This happened when I tried to type `0` as the index.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,108 @@
...
+            if chances < 3:
+                error_msg = f'You have `{3 - chances}/3` chances left'
+            else:
+                error_msg = 'Please try again by using `.wiki` command'
+            try:
+                user = await ctx.bot.wait_for('message', timeout=60.0, check=check)
+                response = await self.bot.get_context(user)
+                if response.command:
+                    return
+                response = int(user.content)
```
ikuyarihS, line 85: This can be very deceiving when `response` is first a `discord.ext.commands.Context` and then it got mutated into an int. I strongly suggest using 2 different variables.
Also this is unnecessary as `wait_for('message', ...)` *already returns a discord.Message* and you can use `.content` of it directly.

ikuyarihS (changes requested):
I like the idea of using chances, that's a great design! We're almost there!

bot/exts/evergreen/wikipedia.py
Anubhav1603: Screenshot_2020-09-07-23-06-58-857
I don't know why that happen to you but when u tried that worked fine can u show me what u searched

ikuyarihS: Sure think, I did what you tried as well, and this is what I got:
image
image
It makes senses as well, since `l_of_list` is indeed an int.

bot/exts/evergreen/wikipedia.py
Anubhav1603: Actually I did this to detect spam of command if user use try to give .wiki in spam bot can detect with this

ikuyarihS: That's fine, but you mutate it into two totally different types, and that can cause confusion to anyone reading the code, so I still strongly suggest using two variables instead.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,108 @@
...
+            return user.author.id == ctx.author.id
+
+        if titles is None:
+            await ctx.send("Sorry, we could not find a wikipedia article using that search term")
+            return
+
+        for title in titles:
+            title_for_creating_link = title.replace(" ", "_")  # wikipedia uses "_" as spaces
+            titles_no_underscore.append(title_for_creating_link)
+        log.info("Finished appening titles with no underscore")
```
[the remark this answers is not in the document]
ikuyarihS: There's a typo here, it should be `appending` instead of `appening`

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+            try:
+                message: Message = await ctx.bot.wait_for('message', timeout=60.0, check=check)
+                response_from_user = await self.bot.get_context(message)
+                if response_from_user.command:
+                    return
+                response = int(message.content)
+                if response < 0:
+                    await ctx.send(f"Sorry, but you can't give negative index, {error_msg}")
+                elif response == 0:
+                    await ctx.send(f"Sorry, please give the range between `1` to `{l_of_list}`, {error_msg}")
```
ikuyarihS, line 92:
```suggestion
                    await ctx.send(f"Sorry, please give an integer between `1` and `{l_of_list}`, {error_msg}")
```
This can be applied to the prompts below as well for `IndexError`

ikuyarihS (changes requested):
Awesome! Only some minor typo and grammar errors and I believe it'll be ready!

ikuyarihS (approved):

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+        self.http_session = bot.http_session
+
+    async def search_wikipedia(self, search_term: str) -> Optional[List[str]]:
+        """Search wikipedia and return the first page found."""
+        async with self.http_session.get(SEARCH_API.format(search_term=search_term)) as response:
+            data = await response.json()
+        page = []
+
+        search_results = data["query"]["search"]
+        if len(search_results) == 0:
```
thomaspet, line 31:
```suggestion
        if not search_results:
```
Pythonic, and might catch a None / False instead of an empty list

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+            return None
+
+        # we dont like "may refere to" pages.
```
thomaspet, line 34:
```suggestion
        # Ignore pages with "may refer to"
```
Sounds better stay neutral, in addition the comment should probably state why we ignore them. Personally I have no idea just reading the comment. Is there no information on those pages?

thomaspet: Additionally, how do you inform the user that their search got them one of these pages? Say it's the only result?

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+        for search_result in search_results:
+            log.info("trying to append titles")
+            if "may refer to" in search_result["snippet"]:
+                pass
+            else:
+                page.append(search_result["title"])
```
thomaspet, line 40:
```suggestion
            log.info("trying to append titles")
            if "may refer to" not in search_result["snippet"]:
                page.append(search_result["title"])
```
More often than not no reason to use "pass".

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+        log.info("Finished appending titles")
+        return page
+
+    @commands.cooldown(1, 10, commands.BucketType.user)
+    @commands.command(name="wikipedia", aliases=["wiki"])
+    async def wikipedia_search_command(self, ctx: commands.Context, *, search: str) -> None:
+        """Returns list of your search query from wikipedia."""
```
thomaspet, line 47: Does it return a list of what i searched for? Or a list of results containing my search quest?

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+            return
+
+        for title in titles:
+            title_for_creating_link = title.replace(" ", "_")  # wikipedia uses "_" as spaces
+            titles_no_underscore.append(title_for_creating_link)
+        log.info("Finished appending titles to titles_no_underscore list")
+
+        async with ctx.typing():
+            for index, title in enumerate(titles, start=1):
+                s_desc += f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})\n'
```
thomaspet, line 67: String concatination is generally bad, I'd use a list of results (or list/generator comprehension) and use str.join for this. (remove definition of `s_desc = ''` as well then)

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+        def check(message: Message) -> bool:
+            return message.author.id == ctx.author.id
+
+        if titles is None:
+            await ctx.send("Sorry, we could not find a wikipedia article using that search term")
+            return
+
+        for title in titles:
+            title_for_creating_link = title.replace(" ", "_")  # wikipedia uses "_" as spaces
+            titles_no_underscore.append(title_for_creating_link)
```
thomaspet, line 62: You don't really need to define the new list first, just list comprehension this?
```suggestion
titles_no_underscore = [title.replace(" ", "_") for title in titles]
```
also `titles_no_underscore` sounds wrong? You're replacing whitespace with `_` so these titles all have underscores?

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+        log.info("Finished appending titles to titles_no_underscore list")
+
+        async with ctx.typing():
+            for index, title in enumerate(titles, start=1):
```
thomaspet, line 66: Why not do this in the previous loop where you already create the underscore links? There are 2 loops now going over the same list. (The previous list comprehension can be dropped if you combine the loops)

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+            embed = Embed(colour=Color.blue(), title=f"Wikipedia results for `{search}`", description=s_desc)
+            embed.timestamp = datetime.datetime.utcnow()
+            await ctx.send(embed=embed)
+        embed = Embed(colour=Color.green(), description="Enter number to choose")
+        msg = await ctx.send(embed=embed)
+        chances = 0
+        total_chances = Wikipedia.total_chance
+        l_of_list = len(titles_no_underscore)  # getting length of list
+
+        while chances <= total_chances:
```
thomaspet, line 77: This could just be a `for _ in range(total_chances )` to remove any chance of infinite loops

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+    @commands.command(name="wikipedia", aliases=["wiki"])
+    async def wikipedia_search_command(self, ctx: commands.Context, *, search: str) -> None:
+        """Returns list of your search query from wikipedia."""
+        titles_no_underscore: List[str] = []
+        s_desc = ''
+
+        titles = await self.search_wikipedia(search)
+
+        def check(message: Message) -> bool:
+            return message.author.id == ctx.author.id
```
thomaspet, line 54: Would it not make sense to check if it's in the same channel as well

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+
+        if titles is None:
```
thomaspet, line 56:
```suggestion
        if not titles:
```
Checks for empty strings, empty lists etc as well.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+        l_of_list = len(titles_no_underscore)  # getting length of list
+
+        while chances <= total_chances:
+            chances += 1
+            if chances < total_chances:
+                error_msg = f'You have `{total_chances - chances}/{total_chances}` chances left'
+            else:
+                error_msg = 'Please try again by using `.wiki` command'
+            try:
+                message: Message = await ctx.bot.wait_for('message', timeout=60.0, check=check)
```
thomaspet, line 84: Why are some variables typed, and others not?

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+
+            except ValueError:
+                await ctx.send(f"Sorry, but you cannot do that, I will only accept an integer, {error_msg}")
+
+            except IndexError:
+                await ctx.send(f"Sorry, please give an integer between `1` to `{l_of_list}`, {error_msg}")
+
+
+def setup(bot: commands.Bot) -> None:
+    """Uptime Cog load."""
```
thomaspet, line 110: Uptime cog?

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+            for index, title in enumerate(titles, start=1):
+                s_desc += f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})\n'
+            embed = Embed(colour=Color.blue(), title=f"Wikipedia results for `{search}`", description=s_desc)
+            embed.timestamp = datetime.datetime.utcnow()
+            await ctx.send(embed=embed)
+        embed = Embed(colour=Color.green(), description="Enter number to choose")
+        msg = await ctx.send(embed=embed)
+        chances = 0
+        total_chances = Wikipedia.total_chance
+        l_of_list = len(titles_no_underscore)  # getting length of list
```
thomaspet, line 75: l_of_list doesn't describe much, i'd at least include `len` spelled out, or `length`. ie: `titles_len`

thomaspet (changes requested):
I might be harsh, but it's my honest review at least.

-Chibli

bot/exts/evergreen/wikipedia.py
Anubhav1603: It returns a list
Screenshot_2020-09-17-16-31-13-895

bot/exts/evergreen/wikipedia.py
Anubhav1603: i am doing str concating because list concating is taking 160 character `` s_desc = "\n".join([f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})``for (index, title) in enumerate(titles, start=1)])` and flake8 is failing that as we are allowed 120 char only

bot/exts/evergreen/wikipedia.py
Anubhav1603: others are mostly string or int thats y i have not typed them

bot/exts/evergreen/wikipedia.py
thomaspet:
```py
s_desc = "\n".join([
    f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})'
    for index, title
    in enumerate(titles, start=1)
])
```
It would look cleaner if you combined the 2 for loops that both iterate `titles`, since you could remove the replace then and just use `title_for_creating_link`, but I digress.

thomaspet:
I still feel like these 2 could be combined
```py
for title in titles:
    title_for_creating_link = title.replace(" ", "_")  # wikipedia uses "_" as spaces
    titles_no_underscore.append(title_for_creating_link)
log.info("Finished appending titles to titles_no_underscore list")

async with ctx.typing():
    for index, title in enumerate(titles, start=1):
        s_desc += f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})\n'
    embed = Embed(colour=Color.blue(), title=f"Wikipedia results for `{search}`", description=s_desc)
    embed.timestamp = datetime.datetime.utcnow()
    await ctx.send(embed=embed)
```
to:
```py
async with ctx.typing():
    for index, title in enumerate(titles, start=1):
        title_for_creating_link = title.replace(" ", "_")  # wikipedia uses "_" as spaces
        
        titles_no_underscore.append(title_for_creating_link)
        s_desc  += f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title_for_creating_link)})\n'
    
    log.info("Finished appending titles to titles_no_underscore list")
    embed = Embed(
        colour=Color.blue(),
        title=f"Wikipedia results for `{search}`",
        description=s_desc
    )
    embed.timestamp = datetime.datetime.utcnow()
    await ctx.send(embed=embed)
```
So that you don't have to iterate twice, and replace twice.

bot/exts/evergreen/wikipedia.py
Anubhav1603: in constants.py i have stored total chances and its incrementing at every input from user.
User has total of 3 chance if he/she gives 3 wrong input looks breaks, so i dont think is there any chance of infinite loop 
or i am missing anything can u tell me how infinite loop can happen.
Loop breaks if these two thing happens also:
 - If user gives right input 
 - If times up

PureFunctor: In my opinion `for _ in range(total_chances)` or more precisely `for _ in range(Wikipedia.total_retries)` is a better idiom for this purpose. It's not that probable that somehow, somewhere, something is going to error out and cause an infinite loop but then again, using common idioms helps better with readability.

PureFunctor:
Looks good! Just a few more changes and it should be fairly ready.

bot/exts/evergreen/wikipedia.py
Anubhav1603: ok

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,109 @@
...
+        chances = 0
+        total_chances = Wikipedia.total_chance
+        titles_len = len(titles_no_underscore)  # getting length of list
+
+        while chances <= total_chances:
+            chances += 1
+            if chances < total_chances:
+                error_msg = f'You have `{total_chances - chances}/{total_chances}` chances left'
+            else:
+                error_msg = 'Please try again by using `.wiki` command'
```
PureFunctor, line 80: In addendum to using `for _ in range(Wikipedia.total_retries)`, this can probably be refactored to the following:
```python
for retry_count in range(1, Wikipedia.total_retries + 1):
    if retry_count < Wikipedia.total_retries:
        retries_left = Wikipedia.total_retries - retry_count
        error_msg = f"You have `{retries_left}/{Wikipedia.total_retries}` chances left"
    else:
        error_msg = 'Please try again by using `.wiki` command'
```
Refactor as you see fit.

bot/exts/evergreen/wikipedia.py
```diff
@@ -68,14 +68,12 @@ def check(message: Message) -> bool:
...
         titles_len = len(titles_no_underscore)  # getting length of list
 
-        while chances <= total_chances:
-            chances += 1
-            if chances < total_chances:
-                error_msg = f'You have `{total_chances - chances}/{total_chances}` chances left'
+        for retry_count in range(1, Wikipedia.total_chance + 1):
+            retries_left = Wikipedia.total_chance - retry_count
+            if retry_count < Wikipedia.total_chance:
+                error_msg = f"You have `{retries_left}/{Wikipedia.total_chance}` chances left"
```
PureFunctor, line 76: Looks good!

bot/exts/evergreen/wikipedia.py
```diff
@@ -103,6 +101,10 @@ def check(message: Message) -> bool:
             except IndexError:
                 await ctx.send(f"Sorry, please give an integer between `1` to `{titles_len}`, {error_msg}")
 
+            except Exception as e:
+                log.info(e)
+                break
+
```
PureFunctor, line 107: Maybe you could add a message on `log.info`? Something like `log.info(f"Caught exception {e}, breaking out of retry loop")`

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+        if not search_results:
+            return None
+
+        # Ignore pages with "may refer to"
+        for search_result in search_results:
+            log.info("trying to append titles")
+            if "may refer to" not in search_result["snippet"]:
+                page.append(search_result["title"])
+        log.info("Finished appending titles")
+        return page
```
PureFunctor, line 46: Returning an `Optional[List[str]]` in `search_wikipedia` seems redundant as you're able to return a list and check for its truthiness. Might be better to have this instead as iterating over an empty list does nothing:
```suggestion
    async def search_wikipedia(self, search_term: str) -> Optional[List[str]]:
        """Search wikipedia and return the first page found."""
        async with self.http_session.get(SEARCH_API.format(search_term=search_term)) as response:
            data = await response.json()
        page = []

        search_results = data["query"]["search"]

        # Ignore pages with "may refer to"
        for search_result in search_results:
            log.info("trying to append titles")
            if "may refer to" not in search_result["snippet"]:
                page.append(search_result["title"])
        log.info("Finished appending titles")
        return page
```

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+        self.http_session = bot.http_session
+
+    @staticmethod
+    def formatted_wiki_urls(index: int, titles: str) -> List[str]:
+        """Making formatted wikipedia links list."""
+        titles = f'`{index}` [{titles}]({WIKIPEDIA_URL.format(title=titles.replace(" ", "_"))})'
+        return titles
```
thomaspet, line 28: I'm for the staticmethod change, but the naming doesn't quite add up. You're typing titles as a string, but titles implies its a list of titles. formatted_wiki_urls & List[str] - also implies it returns a list, but you're just returning a string.

I'd rewrite this function to only handle creating one line, and using loops/comprehensions with it.
```suggestion
    @staticmethod
    def formatted_wiki_url(index: int, titles: str) -> str:
        """Making formatted wikipedia link.."""
        return  f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})'
```

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+
+        def check(message: Message) -> bool:
+            return message.author.id == ctx.author.id and message.channel == ctx.channel
+
+        if not titles:
+            await ctx.send("Sorry, we could not find a wikipedia article using that search term")
+            return
+
+        async with ctx.typing():
+            titles_no_underscore = [title.replace(" ", "_") for title in titles]  # wikipedia uses "_" as spaces
```
thomaspet, line 60: Don't think you need this list at all, just do the replace in place on line 90 where you do
```py
await ctx.send(WIKIPEDIA_URL.format(title=titles_no_underscore[response - 1]))
```
see suggestion

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+                response_from_user = await self.bot.get_context(message)
+                if response_from_user.command:
+                    return
+                response = int(message.content)
+                if response < 0:
+                    await ctx.send(f"Sorry, but you can't give negative index, {error_msg}")
+                elif response == 0:
+                    await ctx.send(f"Sorry, please give an integer between `1` to `{titles_len}`, {error_msg}")
+                else:
+                    await ctx.send(WIKIPEDIA_URL.format(title=titles_no_underscore[response - 1]))
```
thomaspet, line 88:
```suggestion
                    await ctx.send(WIKIPEDIA_URL.format(title=titles[response - 1].replace(" ", "_")))
```
This removes the need for the underscores list, which is still named `titles_no_underscore_, which implies the links don't have underscore in it, but it's the opposite.

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+            titles_no_underscore = [title.replace(" ", "_") for title in titles]  # wikipedia uses "_" as spaces
+            log.info("Finished appending titles to titles_no_underscore list")
+
+            s_desc = "\n".join(self.formatted_wiki_urls(index, title)for index, title in enumerate(titles, start=1))
+            embed = Embed(colour=Color.blue(), title=f"Wikipedia results for `{search}`", description=s_desc)
+            embed.timestamp = datetime.datetime.utcnow()
+            await ctx.send(embed=embed)
+        embed = Embed(colour=Color.green(), description="Enter number to choose")
+        msg = await ctx.send(embed=embed)
+        titles_len = len(titles_no_underscore)  # getting length of list
```
thomaspet, line 69:
```suggestion
        titles_len = len(titles)  # getting length of list
```
Remove the title_no_underscore list, and just use the length of titles.. since that's what its based on anyways

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+            if retry_count < Wikipedia.total_chance:
+                error_msg = f"You have `{retries_left}/{Wikipedia.total_chance}` chances left"
+            else:
+                error_msg = 'Please try again by using `.wiki` command'
+            try:
+                message = await ctx.bot.wait_for('message', timeout=60.0, check=check)
+                response_from_user = await self.bot.get_context(message)
+                if response_from_user.command:
+                    return
+                response = int(message.content)
```
thomaspet, line 82:
```suggestion
                response_from_user = await self.bot.get_context(message)
                
                if response_from_user.command:
                    return
                    
                response = int(message.content)
```
Just personal preference for readability

thomaspet (changes requested):

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,109 @@
...
+
+        if not titles:
+            await ctx.send("Sorry, we could not find a wikipedia article using that search term")
+            return
+
+        async with ctx.typing():
+            titles_no_underscore = [title.replace(" ", "_") for title in titles]  # wikipedia uses "_" as spaces
+            log.info("Finished appending titles to titles_no_underscore list")
+
+            s_desc = "\n".join(self.formatted_wiki_urls(index, title)for index, title in enumerate(titles, start=1))
```
thomaspet, line 63:
```suggestion
            s_desc = "\n".join(self.formatted_wiki_url(index, title) for index, title in enumerate(titles, start=1))
```
and don't forget to rename this.

thomaspet (changes requested):

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,109 @@
+import asyncio
+import datetime
+import logging
+from typing import List, Optional
```
PureFunctor, line 4: `Optional` is not being used anymore and `flake8` errors out when `pre-commit` hooks are run.

PureFunctor (changes requested):

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,109 @@
...
+    """Get info from wikipedia."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+        self.http_session = bot.http_session
+
+    @staticmethod
+    def formatted_wiki_url(index: int, titles: str) -> str:
+        """Making formatted wikipedia link.."""
+        return f'`{index}` [{titles}]({WIKIPEDIA_URL.format(title=titles.replace(" ", "_"))})'
```
thomaspet, line 27:
```suggestion
    def formatted_wiki_url(index: int, title: str) -> str:
        """Making formatted wikipedia link.."""
        return f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})'
```
Titles is still plural

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,109 @@
...
+
+        search_results = data["query"]["search"]
+
+        # Ignore pages with "may refer to"
+        for search_result in search_results:
+            log.info("trying to append titles")
+            if "may refer to" not in search_result["snippet"]:
+                page.append(search_result["title"])
+        log.info("Finished appending titles")
+        return page
```
thomaspet, line 43:
```suggestion
    async def search_wikipedia(self, search_term: str) -> List[str]:
        """Search wikipedia and return the first 10 pages found."""
        async with self.http_session.get(SEARCH_API.format(search_term=search_term)) as response:
            data = await response.json()

        pages = []

        search_results = data["query"]["search"]

        # Ignore pages with "may refer to"
        for search_result in search_results:
            log.info("trying to append titles")
            if "may refer to" not in search_result["snippet"]:
                pages.append(search_result["title"])

        log.info("Finished appending titles")
        return pages
```
`page` is a list, so it should have a plural name.
Dont know why you added back Optional after Pure saying it's not right. This function will always return a list, the only difference is that it might be an empty list. Which i dont think you need to worry about typing.
The docstring says "first page", but that url path, afaik, returns the first 10 pages.

thomaspet (changes requested):

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,109 @@
...
+class WikipediaCog(commands.Cog):
+    """Get info from wikipedia."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+        self.http_session = bot.http_session
+
+    @staticmethod
+    def formatted_wiki_url(index: int, titles: str) -> str:
+        """Making formatted wikipedia link.."""
```
thomaspet, line 26:
```suggestion
        """Making formatted wikipedia link."""
```
Nit-picking

bot/exts/evergreen/wikipedia.py
```diff
@@ -0,0 +1,111 @@
...
+    """Get info from wikipedia."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+        self.http_session = bot.http_session
+
+    @staticmethod
+    def formatted_wiki_url(index: int, titles: str) -> str:
+        """Making formatted wikipedia link."""
+        return f'`{index}` [{titles}]({WIKIPEDIA_URL.format(title=titles.replace(" ", "_"))})'
```
thomaspet, line 27: was this changed back to `titles`? It's a function that takes 1 title at the time, hence why I meant it shouldn't have a plural name for title. (titles). 
```suggestion
    @staticmethod
    def formatted_wiki_url(index: int, title: str) -> str:
        """Formating wikipedia link with index and title."""
        return f'`{index}` [{title}]({WIKIPEDIA_URL.format(title=title.replace(" ", "_"))})'
```

Anubhav1603: Sorry I understood opposite, I have now committed the changes which u propose🙂

thomaspet (approved):
Make sure you test it locally after all these changes.

Anubhav1603:
> Make sure you test it locally after all these changes.

Sure🙂

PureFunctor (approved):
Looks good!

PR #441, opened 2020-09-16
Update .roll to use new dice emojis

### LINKED ISSUE #440: .dice doesn’t display emojis
## Description
After an update to the server's dice emojis' names (together with updated designs), the output of `.dice` will be the string of the old emoji names, instead of the new ones. This causes it to be displayed as text.

## Steps to Reproduce

Send `.roll` in any channel that accepts the command.

## Expected Behaviour

Assuming that the output is 3, the returned string should be:
`:dice_3:`

This would be converted to the correct emoji by Discord.

## Actual Behaviour

The output of `.dice` will be the string of the old emoji names, like this:
`:terning3:`

This is not displayed to an emoji.

## Known Impacted Platforms

- [x] Web
- [x] Desktop
- [x] Android App
- [x] iOS App

## Possible Solutions

Change the emoji names in the command to the new format:

- `:dice_1:`
- `:dice_2:`
- ...
- `:dice_6:`

gustavwilliam:
## Relevant Issues

Closes #440.

## Description

- Emojis in `constants.py` were replaced with the new emojis, following the dice redesign
- `.roll` command was updated to use the new emoji names in `constants.py`

## Did you:

### DIFF
--- a/bot/exts/evergreen/fun.py
+++ b/bot/exts/evergreen/fun.py
@@ -41,7 +41,7 @@ class Fun(Cog):
         elif num_rolls < 1:
             output = ":no_entry: You must roll at least once."
         for _ in range(num_rolls):
-            terning = f"terning{random.randint(1, 6)}"
+            terning = f"dice_{random.randint(1, 6)}"
             output += getattr(Emojis, terning, '')
         await ctx.send(output)
 
--- a/bot/exts/evergreen/fun.py
+++ b/bot/exts/evergreen/fun.py
@@ -41,8 +41,8 @@ class Fun(Cog):
         elif num_rolls < 1:
             output = ":no_entry: You must roll at least once."
         for _ in range(num_rolls):
-            terning = f"dice_{random.randint(1, 6)}"
-            output += getattr(Emojis, terning, '')
+            dice = f"dice_{random.randint(1, 6)}"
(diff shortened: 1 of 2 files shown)


bot/exts/evergreen/fun.py
```diff
@@ -41,7 +41,7 @@ async def roll(self, ctx: Context, num_rolls: int = 1) -> None:
         elif num_rolls < 1:
             output = ":no_entry: You must roll at least once."
         for _ in range(num_rolls):
-            terning = f"terning{random.randint(1, 6)}"
+            terning = f"dice_{random.randint(1, 6)}"
```
Den4200, line 44: Maybe also rename the variable to `dice` or `die`?

Den4200 (changes requested):
Nice, only one small request.

bot/exts/evergreen/fun.py
gustavwilliam: Sure; we could do that. The emojis on the server are called `dice_X`, though.

If we rename this, I think we should rename the server icons as well.

Den4200 (approved):
Looks good to me, thanks!

kwzrd (approved):
Thanks for the fix!

PR #443, opened 2020-09-18
Stop users from viewing messages they shouldn't.

kosayoda:
Using a user token, a user could fetch the message ID of a message in any channel, which may leak information when potential Message objects are automatically converted and parsed.

Now, the bot will only retrive text from a valid Message object if the user has read permissions for the message the channel is in.

image


kwzrd (approved):
Looks LGTM to me, tested with an alt in the dev server.

scragly (approved):

PR #448, opened 2020-09-21
Prevent uwu output from pinging globally allowed roles.

scragly:
Description

This PR introduces sanitisation of the uwu text before uwufication in order to prevent moderator roles being pinged unnecessarily.

I've just changed the `str` typehint to the `clean_content` converter to do this as simply as possible.

## Reasoning

While we do have a global limit to allowed roles being pinged in the bot, all mod+ roles were still capable of being pinged through this method.

## Screenshots

**Before** (role resolves)
image

**After** (role is now plaintext)
image

## Did you:

### DIFF
--- a/bot/exts/evergreen/fun.py
+++ b/bot/exts/evergreen/fun.py
@@ -5,7 +5,7 @@ from typing import Callable, Tuple, Union
 
 from discord import Embed, Message
 from discord.ext import commands
-from discord.ext.commands import Bot, Cog, Context, MessageConverter
+from discord.ext.commands import Bot, Cog, Context, MessageConverter, clean_content
 
 from bot import utils
 from bot.constants import Emojis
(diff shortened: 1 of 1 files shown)


bot/exts/evergreen/fun.py
```diff
@@ -71,7 +71,7 @@ async def roll(self, ctx: Context, num_rolls: int = 1) -> None:
         await ctx.send(output)
 
     @commands.command(name="uwu", aliases=("uwuwize", "uwuify",))
-    async def uwu_command(self, ctx: Context, *, text: str) -> None:
+    async def uwu_command(self, ctx: Context, *, text: clean_content(fix_channel_mentions=True)) -> None:
```
SebastiaanZ, line 74: If the command argument is a message ID or message jumplink, the command will internally fetch the message text on this line in the command:

```py
        text, embed = await Fun._get_text_and_embed(ctx, text)
```

This bypasses the `clean_content` converter (which is only applied to the ID/jumplink, not the message text that is fetched from it later).

SebastiaanZ (changes requested):

bot/exts/evergreen/fun.py
scragly: Good point, thanks. I'll get that fixed up.

scragly: OK, I should have resolved this now, along with also applying the cleaning to the mixed case command so mentions don't scramble the raw mention format, instead always using the resolved names.

SebastiaanZ (approved):

Akarys42 (approved):
Works like a charm

PR #455, opened 2020-09-22
Partnerzodiac

### LINKED ISSUE #414: partnerzodiac ext update with dates | or new zodiac ext
## Description

Add additional parameters that can return 
 - the date based on zodiac sign 
 - zodiac sign based on dates.

## Reasoning
I would like to for when `.partnerzodiac` is called with a zodiac sign,
In addition the zodiac info, we can also have the date range for the signs, such that those unfamiliar with the zodiac signs can learn about the dates as well.

## Proposed Implementation

.partnerzodiac date <zodiac>
.partnerzodiac date virgo
The amazing people who are born between August 23 - September 22, with amazing qualities are Virgos

.partnerzodiac zodiac <date>
.partnerzodiac zodiac september 15
The amazing people who are born between August 23 - September 22 , with amazing qualities are Virgos

## Additional Details
- partnerzodiac can be a subpart of zodiac

Anubhav1603:
"Closes #414 "

I have added two subcommand for partnerzodiac date and zodiac

zodiac
zodia
error
er
date
date


bot/exts/valentines/valentine_zodiac.py
```diff
@@ -29,13 +37,127 @@ def load_json() -> dict:
...
+            embed.description = "Umm you gave wrong zodiac name so i aren't able to find any :sweat_smile:"
+            log.info("Wrong Zodiac name provided")
+        return embed
+        log.info("Zodiac embed ready")
+
+    def zodiac_date_verifer(self, month: str, date: int) -> str:
+        """Returns zodiac sign by checking month and date."""
+        month = month.capitalize()
+        log.info("started searching zodaic sign based on month and date")
+        if month == "January" or month == "Jan":
```
RohanJnr, line 64: I suggest the following method for getting the zodiac sign as per month and date.
Have a DICT to store the starting and ending datetime objects of a zodiac sign
example:
`zodiacs = {
"Aquarius":(datetime(2020, 1, 20), datetime(2020, 2, 18))
}`
You can create a new datetime() object with the month and date passed by the user.
Then, you can check if that datetime() object is in between the starting and ending dates as specified in the dict.
lets call the datetime() object created with the user input *query_datetime*

`for zodiac_name, date_range in zodiacs.items(): 
    if (date_range[0]<=query_datetime<=date_range[1]):  
        # you found the zodiac`

RohanJnr: the starting and ending dates can also be presented in the JSON file, and the datetime() objects can be created during the Cog load.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -29,13 +37,127 @@ def load_json() -> dict:
...
+        log.info("Zodiac name sent")
+
+    @commands.group(name="partnerzodiac", invoke_without_command=True)
+    async def partner_zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
         """Provides a counter compatible zodiac sign to the given user's zodiac sign."""
         try:
             compatible_zodiac = random.choice(self.zodiacs[zodiac_sign.lower()])
         except KeyError:
-            return await ctx.send(zodiac_sign.capitalize() + " zodiac sign does not exist.")
+            return await ctx.send(f"`{zodiac_sign.capitalize()}` zodiac sign does not exist.")
```
RohanJnr, line 160: the function annotation says the return Type is None, so we have to refactor our code to match that
`await ctx.send(f"`{zodiac_sign.capitalize()}` zodiac sign does not exist.")`
`return`

RohanJnr (changes requested):
I tested the commands and all seem to be working.
There are a few changes to the design that I suggest.

bot/exts/valentines/valentine_zodiac.py
Anubhav1603: I think i should skip the year because year can change but date and month will remain same always

RohanJnr:
Currently if I want to get information about the zodiac sign Libra, I will have to do `.partnerzodiac zodiac libra`. I feel that just `.zodiac libra` will be a better choice instead and have `partnerzodiac` as a subcommand.

### The structure I am proposing is:

##### Group Zodiac
 - provides information about zodiac sign by taking zodiac sign name as input.
 - provides information about zodiac sign by taking month and date as input.

##### SubCommand partnerzodiac
 - provides a counter compatible zodiac sign to the given user's zodiac sign

bot/exts/valentines/valentine_zodiac.py
RohanJnr: The year can be anything as that wouldn't matter I believe

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -29,13 +37,127 @@ def load_json() -> dict:
...
+            embed.add_field(name='__Full form__', value=zodiac_fact[f"{c_zodiac}"]["full_form"], inline=False)
+            embed.add_field(name='__Motto__', value=zodiac_fact[f"{c_zodiac}"]["Motto"], inline=False)
+            embed.add_field(name='__Strengths__', value=zodiac_fact[f"{c_zodiac}"]["Strengths"], inline=False)
+            embed.add_field(name='__Weaknesses__', value=zodiac_fact[f"{c_zodiac}"]["Weaknesses"], inline=False)
+            embed.set_thumbnail(url=zodiac_fact[f"{c_zodiac}"]["url"])
+        else:
+            embed.description = "Umm you gave wrong zodiac name so i aren't able to find any :sweat_smile:"
+            log.info("Wrong Zodiac name provided")
+        return embed
+        log.info("Zodiac embed ready")
```
RohanJnr, line 58: Why is there a log statement after the return statement? anything after the return statement will not be executed, you should move the log above the return.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -29,13 +37,127 @@ def load_json() -> dict:
             zodiacs = load(json_data)
             return zodiacs
 
-    @commands.command(name="partnerzodiac")
-    async def counter_zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
+    def zodiac_sign_verify(self, zodiac: str) -> discord.Embed:
```
RohanJnr, line 40: this method can be a staticmethod

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,6 +13,14 @@
 LETTER_EMOJI = ':love_letter:'
 HEART_EMOJIS = [":heart:", ":gift_heart:", ":revolving_hearts:", ":sparkling_heart:", ":two_hearts:"]
 
+zodiac_signs = ["Aries", "Taurus", "Gemini", "Cancer", "Leo", "Virgo", "Libra",
+                "Scorpio", "Sagittarius", "Capricorn", "Aquarius", "Pisces"]
+
+
+with open(Path("bot/resources/valentines/zodiac_explanation.json"), "r", encoding="utf8") as file:
+    """Load zodiac zodiac explanation from static JSON resource."""
+    zodiac_fact = load(file)
```
RohanJnr, line 22: there is already a file IO operation within the Class `ValentineZodiac` so it will be better to have both of them either inside the class as instance variables or outside the class as global variables.
I would say have it inside the class, and load the JSON with another method, say `load_zodiac_facts`
This is for consistency.

RohanJnr (changes requested):

bot/exts/valentines/valentine_zodiac.py
Anubhav1603: i need self in this, so i cant change this to static method

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
 LETTER_EMOJI = ':love_letter:'
 HEART_EMOJIS = [":heart:", ":gift_heart:", ":revolving_hearts:", ":sparkling_heart:", ":two_hearts:"]
 
+ZODIAC_SIGNS = ["Aries", "Taurus", "Gemini", "Cancer", "Leo", "Virgo", "Libra",
+                "Scorpio", "Sagittarius", "Capricorn", "Aquarius", "Pisces"]
```
RohanJnr, line 19: this variable can be removed, as mentioned in further reviews.
```suggestion

```

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
...
+
+MONTH_NAME = {"January": 1, "Jan": 1, "Feburary": 2, "Feb": 2, "March": 3, "Mar": 3,
+              "April": 4, "Apr": 4, "May": 5, "June": 6, "Jun": 6, "July": 7, "Jul": 7,
+              "August": 8, "Aug": 8, "September": 9, "Sept": 9, "October": 10, "Oct": 10,
+              "November": 11, "Nove": 11, "December": 12, "December": 12
+              }
```
RohanJnr, line 25:
```suggestion

```
This can be removed as we can use the calendar module instead as mentioned in one of the reviews below.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
...
-            return zodiacs
+            return zodiacs, zodiac_fact
+
+    def zodiac_sign_verify(self, zodiac: str) -> discord.Embed:
+        """Gives informative zodiac embed."""
+        zodiac = zodiac.capitalize()
+        zodiac_fact = self.zodiac_fact
+        embed = discord.Embed()
+        embed.color = Colours.pink
+        if zodiac in ZODIAC_SIGNS:
```
RohanJnr, line 52: we can use the keys from `self.zodiac_fact` to check instead of having another variable for it.
```suggestion
        if zodiac in self.zodiac_fact:
```

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
...
+            zodiac_fact = load(json_data)
         with p.open(encoding="utf8") as json_data:
             zodiacs = load(json_data)
-            return zodiacs
+            return zodiacs, zodiac_fact
+
+    def zodiac_sign_verify(self, zodiac: str) -> discord.Embed:
+        """Gives informative zodiac embed."""
+        zodiac = zodiac.capitalize()
+        zodiac_fact = self.zodiac_fact
```
RohanJnr, line 49: This is redundant, use `self.zodiac_fact` directly.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
...
+        embed.color = Colours.pink
+        if zodiac in ZODIAC_SIGNS:
+            log.info("Making zodiac embed")
+            embed.title = f"__{zodiac}__"
+            embed.description = zodiac_fact[f"{zodiac}"]["About"]
+            embed.add_field(name='__Full form__', value=zodiac_fact[f"{zodiac}"]["full_form"], inline=False)
+            embed.add_field(name='__Motto__', value=zodiac_fact[f"{zodiac}"]["Motto"], inline=False)
+            embed.add_field(name='__Strengths__', value=zodiac_fact[f"{zodiac}"]["Strengths"], inline=False)
+            embed.add_field(name='__Weaknesses__', value=zodiac_fact[f"{zodiac}"]["Weaknesses"], inline=False)
+            embed.set_thumbnail(url=zodiac_fact[f"{zodiac}"]["url"])
```
RohanJnr, line 60: F strings are not needed here, directly use the variable inside the square bracket.

bot/resources/valentines/zodiac_explanation.json
```diff
@@ -0,0 +1,122 @@
...
+            "About": "Amazing people born between **November 22** to **December 21**. Represented by the archer, Sagittarians are always on a quest for knowledge. The last fire sign of the zodiac, Sagittarius launches its many pursuits like blazing arrows, chasing after geographical, intellectual, and spiritual adventures.",
+            "Motto": "***“Towering genius disdains a beaten path.”***",
+            "Strengths": "generous, idealistic, great sense of humor.",
+            "Weaknesses": "promises more than can deliver, very impatient, will say anything no matter how undiplomatic.",
+            "full_form": "__**S**__eductive\n__**A**__dventurous\n__**G**__rateful\n__**I**__ntelligent\n__**T**__railblazing\n__**T**__enacious adept\n__**A**__dept\n__**R**__esponsible\n__**I**__dealistic\n__**U**__nparalled\n__**S**__ophisticated",
+            "url": "https://www.horoscope.com/images-US/signs/profile-sagittarius.png"
+        },
+    "Capricorn": {
+            "start_at": [12, 22],
+            "end_at": [1, 19],
```
RohanJnr, line 94: here also the ISO format has to be used, but for the ending date, the year needs to be 2021 as we are ending in 2020.

bot/resources/valentines/zodiac_explanation.json
```diff
@@ -0,0 +1,122 @@
+{
+    "Aries": {
+            "start_at":[3, 21],
+            "end_at": [4, 19],
```
RohanJnr, line 4: Using dates with ISO format will be much more readable in my opinion. So, you could change it to:
```suggestion
            "start_at": "2020-03-21",
            "end_at": "2020-04-19",
```
This has to be done to all zodiac signs in the file.

RohanJnr (changes requested):
The command works but here are few changes that will make the code look cleaner :D

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
...
     """A Cog that returns a counter compatible zodiac sign to the given user's zodiac sign."""
 
     def __init__(self, bot: commands.Bot):
         self.bot = bot
-        self.zodiacs = self.load_json()
+        self.zodiacs, self.zodiac_fact = self.load_comp_json()
 
     @staticmethod
-    def load_json() -> dict:
+    def load_comp_json() -> dict:
```
RohanJnr, line 36:
```suggestion
    def load_comp_json() -> tuple:
```
the return type is a tuple here

PureFunctor: I also suggest using `Tuple[dict, dict]` can also be used for brevity. Make sure to import `Tuple` from `typing` though.

Anubhav1603: Y 2 dict?

PureFunctor: You're returning a tuple of `zodiacs` and `zodiac_fact`, both of which are dictionaries.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
...
         """Load zodiac compatibility from static JSON resource."""
+        p1 = Path("bot/resources/valentines/zodiac_explanation.json")
         p = Path("bot/resources/valentines/zodiac_compatibility.json")
```
PureFunctor, line 39: Why not name these `explanation_file` and `compatibility_file` respectively? `p1` and `p` seems too abstract and inconsistent.
```suggestion
        explanation_file = Path("bot/resources/valentines/zodiac_explanation.json")
        compatibility_file = Path("bot/resources/valentines/zodiac_compatibility.json")
```

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
...
+        with p1.open(encoding="utf8") as json_data:
+            zodiac_fact = load(json_data)
         with p.open(encoding="utf8") as json_data:
             zodiacs = load(json_data)
```
PureFunctor, line 43: Don't forget to change these as well.
```suggestion
        with explanation_file.open(encoding="utf8") as json_data:
            zodiac_fact = load(json_data)
        with compatibility_file.open(encoding="utf8") as json_data:
            zodiacs = load(json_data)
```

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -13,29 +15,115 @@
...
-            return zodiacs
+            return zodiacs, zodiac_fact
```
PureFunctor, line 44: More of a stylistic choice but you can dedent this by a block.

PureFunctor (changes requested):

bot/exts/valentines/valentine_zodiac.py
Anubhav1603: err
actually this i am getting when i did that

Anubhav1603: i have fixed this issue sorry for bothering u

PureFunctor:
Looks good so far, just needs to address a few things @RohanJnr pointed out.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,116 @@ class ValentineZodiac(commands.Cog):
...
+                zodiac = zodiac_name
+        log.info("Zodiac name sent")
+        return zodiac
+
+    @commands.group(name='zodiac', invoke_without_command=True)
+    async def zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
+        """Provides information about zodiac sign by taking zodiac sign name as input."""
+        if zodiac_sign.startswith("`"):
+            await ctx.send("Please don't include `")
+            return
```
RohanJnr, line 77: What is the need of this? I don't think users will generally use backticks.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,116 @@ class ValentineZodiac(commands.Cog):
...
+        await ctx.send(embed=final_embed)
+
+    @zodiac.command(name="partnerzodiac")
+    async def partner_zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
         """Provides a counter compatible zodiac sign to the given user's zodiac sign."""
+        embed = discord.Embed()
+        embed.color = Colours.pink
+        if zodiac_sign.startswith("`"):
+            await ctx.send("Please don't include `")
+            return
```
RohanJnr, line 118: here also, I don't think we need this.

bot/resources/valentines/zodiac_explanation.json
```diff
@@ -0,0 +1,122 @@
...
+    "About": "Amazing people born between **November 22** to **December 21**. Represented by the archer, Sagittarians are always on a quest for knowledge. The last fire sign of the zodiac, Sagittarius launches its many pursuits like blazing arrows, chasing after geographical, intellectual, and spiritual adventures.",
+    "Motto": "***\u201cTowering genius disdains a beaten path.\u201d***",
+    "Strengths": "generous, idealistic, great sense of humor.",
+    "Weaknesses": "promises more than can deliver, very impatient, will say anything no matter how undiplomatic.",
+    "full_form": "__**S**__eductive\n__**A**__dventurous\n__**G**__rateful\n__**I**__ntelligent\n__**T**__railblazing\n__**T**__enacious adept\n__**A**__dept\n__**R**__esponsible\n__**I**__dealistic\n__**U**__nparalled\n__**S**__ophisticated",
+    "url": "https://www.horoscope.com/images-US/signs/profile-sagittarius.png"
+  },
+  "Capricorn": {
+    "start_at": "2020-12-22",
+    "end_at": "2020-01-19",
```
RohanJnr, line 94: the year should be 2021 here, because it doesn't make sense to go from 2020 December to 2020 Jan.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,116 @@ class ValentineZodiac(commands.Cog):
...
         """Load zodiac compatibility from static JSON resource."""
-        p = Path("bot/resources/valentines/zodiac_compatibility.json")
-        with p.open(encoding="utf8") as json_data:
+        explanation_file = Path("bot/resources/valentines/zodiac_explanation.json")
+        compatibility_file = Path("bot/resources/valentines/zodiac_compatibility.json")
+        with explanation_file.open(encoding="utf8") as json_data:
+            zodiac_fact = load(json_data)
+            for _, zodiac_data in zodiac_fact.items():
+                zodiac_data['start_at'] = datetime.fromisoformat(zodiac_data['start_at'])
+                zodiac_data['end_at'] = datetime.fromisoformat(zodiac_data['end_at'])
```
RohanJnr, line 36: use values() instead of items()
```suggestion
            for zodiac_data in zodiac_fact.values():
                zodiac_data['start_at'] = datetime.fromisoformat(zodiac_data['start_at'])
                zodiac_data['end_at'] = datetime.fromisoformat(zodiac_data['end_at'])
```

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -1,7 +1,10 @@
+import calendar
 import logging
 import random
+from datetime import datetime
```
RohanJnr, line 4: use date instead of DateTime as we have nothing to do with DateTime.

RohanJnr (changes requested):
Just a few more changes and we are good to go xD

bot/exts/valentines/valentine_zodiac.py
Anubhav1603: Actually if user search for ` the embed looks weired that y I have included this

bot/resources/valentines/zodiac_explanation.json
Anubhav1603: Ok

bot/exts/valentines/valentine_zodiac.py
Anubhav1603: Actually I am taking date as param in command so if I use date directly it messes up the code

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,110 @@ class ValentineZodiac(commands.Cog):
...
+            embed.description = f"{error_msg}\n{error}"
+            log.info("Wrong Zodiac name provided")
+        log.info("Zodiac embed ready")
+        return embed
+
+    def zodiac_date_verifier(self, query_datetime: date) -> str:
+        """Returns zodiac sign by checking month and date."""
+        for zodiac_name, zodiac_data in self.zodiac_fact.items():
+            if zodiac_data["start_at"] <= query_datetime <= zodiac_data["end_at"]:
+                zodiac = zodiac_name
```
RohanJnr, line 68: you can add a break statement after this line to stop the loop as we have already found the sign,

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,110 @@ class ValentineZodiac(commands.Cog):
...
-            color=Colours.pink
-        )
-        embed.add_field(
-            name=f'A letter from Dr.Zodiac {LETTER_EMOJI}',
-            value=compatible_zodiac['description']
-        )
+            err_comp = [f"`{i}` {zod_name}" for i, zod_name in enumerate(self.zodiac_fact.keys(), start=1)]
+            error = ("\n").join(err_comp)
+            error_msg = f"**{zodiac_sign}** is not a valid zodiac sign, here is the list of valid zodiac signs."
+            embed.description = f"{error_msg}\n{error}"
```
RohanJnr, line 128: This part of the code is being repeated. It's better to make a function for it.
repeated at line 56

RohanJnr (changes requested):

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,115 @@ class ValentineZodiac(commands.Cog):
...
-    @commands.command(name="partnerzodiac")
-    async def counter_zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
+    def error(self, zodiac: str) -> discord.Embed:
+        """Returns error embed."""
+        embed = discord.Embed()
+        embed.color = Colours.pink
+        err_comp = [f"`{i}` {zod_name}" for i, zod_name in enumerate(self.zodiac_fact.keys(), start=1)]
+        err = ("\n").join(err_comp)
+        error_msg = f"**{zodiac}** is not a valid zodiac sign, here is the list of valid zodiac signs."
+        embed.description = f"{error_msg}\n{err}"
```
RohanJnr, line 48: a lot of un-necessary variables here.
you can replace with this
```suggestion
        embed.description = f"**{zodiac}** is not a valid zodiac sign, here is the list of valid zodiac signs.\n"
        for i, zod_name in enumerate(self.zodiac_fact.keys(), start=1):
            embed.description += f"`{i}` {zod_name}\n"
```

RohanJnr (changes requested):

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -33,48 +25,55 @@ def __init__(self, bot: commands.Bot):
...
+        log.info("Wrong Zodiac name provided")
+        return embed
 
     def zodiac_sign_verify(self, zodiac: str) -> discord.Embed:
```
PureFunctor, line 52: `zodiac_sign_verify` doesn't really express what it actually does, maybe this could be renamed to `zodiac_build_embed`?
```suggestion
    def zodiac_build_embed(self, zodiac: str) -> discord.Embed:
```

PureFunctor (changes requested):

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,120 @@ class ValentineZodiac(commands.Cog):
...
+
+    @zodiac.command(name="date")
+    async def date_and_month(self, ctx: commands.Context, query_date: int, month: Union[int, str]) -> None:
+        """Provides information about zodiac sign by taking month and date as input."""
+        if isinstance(month, str):
+            try:
+                month = month.capitalize()
+                month = list(calendar.month_abbr).index(month[:3])
+            except ValueError:
+                await ctx.send(f"Sorry, but `{month}` is wrong month name.")
```
wookie184, line 98: Something like ```await ctx.send(f"Sorry, but `{month}` is not a valid month name.")``` would sound better grammatically.

bot/exts/valentines/valentine_zodiac.py
wookie184: You could actually just `return zodiac_name` directly from loop here

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,120 @@ class ValentineZodiac(commands.Cog):
...
+        else:
+            try:
+                zodiac_sign_based_on_month_and_date = self.zodiac_date_verifier(date(2020, month, query_date))
+                log.info("zodiac sign based on month and date received")
+            except ValueError as e:
+                log.info("invalid date or month given")
+                final_embed = discord.Embed()
+                final_embed.color = Colours.pink
+                final_embed.description = f"Zodiac sign is not found because, {e}"
+                log.info(e)
```
wookie184: Pretty minor thing, but these two `log.info`s could probably be combined, and capital letters could also be used at the start of the sentences.

wookie184 (commented):
This is my first review here so not sure if I was too picky, most of these things are minor and could probably be ignored, but I thought I would point them out anyway.

bot/resources/valentines/zodiac_explanation.json
```diff
@@ -0,0 +1,122 @@
+{
+  "Aries": {
+    "start_at": "2020-03-21",
```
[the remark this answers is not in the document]
Anubhav1603: actually year dont matter ,we only need month and date and this year is leap year thats y i have hardcoded the year

RohanJnr (approved):

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
 
-    @commands.command(name="partnerzodiac")
-    async def counter_zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
+    def error(self, zodiac: str) -> discord.Embed:
+        """Returns error embed."""
+        embed = discord.Embed()
+        embed.color = Colours.pink
+        error_comp = "\n".join(
+            [f"`{i}` {zod_name}" for i, zod_name in enumerate(self.zodiac_fact.keys(), start=1)]
+        )
```
wookie184, line 47: I just realised I made a typo in my previous review and you had already removed the square brackets so I basically accidentally told you to put them back in. Completely my fault, sorry for the confusion, but it would be best without them if you wouldn't mind changing it back.
```suggestion
        error_comp = "\n".join(
            f"`{i}` {zod_name}" for i, zod_name in enumerate(self.zodiac_fact.keys(), start=1)
        )
```

wookie184 (changes requested):
Previous changes look great, thanks. Final few small things I've just noticed after trying it out, after this it should be good to go as far as I'm concerned :+1:

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+        explanation_file = Path("bot/resources/valentines/zodiac_explanation.json")
+        compatibility_file = Path("bot/resources/valentines/zodiac_compatibility.json")
+        with explanation_file.open(encoding="utf8") as json_data:
+            zodiac_fact = json.load(json_data)
+            for zodiac_data in zodiac_fact.values():
+                zodiac_data['start_at'] = date.fromisoformat(zodiac_data['start_at'])
+                zodiac_data['end_at'] = date.fromisoformat(zodiac_data['end_at'])
+        with compatibility_file.open(encoding="utf8") as json_data:
+            zodiacs = json.load(json_data)
+        return zodiacs, zodiac_fact
```
thomaspet: Is there a reason this can't be a single json file, or be returned as a dictionary of { zodiac: { explanation, compatibility }} etc?

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
 
-    @commands.command(name="partnerzodiac")
-    async def counter_zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
+    def error(self, zodiac: str) -> discord.Embed:
```
thomaspet, line 41: This doesn't error, like the name implies.  I'd rename it to something like `generate_invalidname_embed` or `invalidname_embed` etc.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+        """Returns error embed."""
+        embed = discord.Embed()
+        embed.color = Colours.pink
```
thomaspet, line 44: Is a pink error message part of the specs? Else i'd go with something a bit more red-ish

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+    async def date_and_month(self, ctx: commands.Context, query_date: int, month: Union[int, str]) -> None:
+        """Provides information about zodiac sign by taking month and date as input."""
+        if isinstance(month, str):
+            month = month.capitalize()
+            try:
+                month = list(calendar.month_abbr).index(month[:3])
+            except ValueError:
+                await ctx.send(f"Sorry, but `{month}` is not a valid month name.")
+                return
+        if (month == 1 and (1 <= query_date <= 19)) or (month == 12 and (22 <= query_date <= 31)):
```
thomaspet: Since `and` has a higher [operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence) the outer `()` can be removed here.. Possibly all of them, looking at it, since `==`, `<=` and `>=` has higher precedence than or/and.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+            zodiac = "capricorn"
+            final_embed = self.zodiac_build_embed(zodiac)
+        else:
+            try:
+                zodiac_sign_based_on_month_and_date = self.zodiac_date_verifier(date(2020, month, query_date))
```
thomaspet:
```suggestion
                zodiac_sign_based_on_date = self.zodiac_date_verifier(date(2020, month, query_date))
```
Date == month_and_date

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+                log.info("zodiac sign based on month and date received.")
+            except ValueError as e:
+                final_embed = discord.Embed()
+                final_embed.color = Colours.pink
```
thomaspet, line 107: Is pink spec, else I don't think it's much of an error color.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+                final_embed.description = f"Zodiac sign is not found because, {e}"
+                log.info(f"error caused due to: {e}.")
```
thomaspet, line 109:
```suggestion
                log.info(f"Error in "zodiac date" command:\n{e}.")
```
Location is very helpful when debuging

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+            else:
+                final_embed = self.zodiac_build_embed(zodiac_sign_based_on_month_and_date)
+
+        await ctx.send(embed=final_embed)
+        log.info("Zodiac sign embed based on month and date is now sent.")
```
thomaspet, line 114:
```suggestion
        log.info("Zodiac sign embed based on date is now sent.")
```

thomaspet (changes requested):
Few of my remarks

bot/exts/valentines/valentine_zodiac.py
thomaspet: `import datetime` -> `datetime.date` or `from datetime import date as datetime` etc.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+            final_embed = self.zodiac_build_embed(zodiac)
+        else:
+            try:
+                zodiac_sign_based_on_month_and_date = self.zodiac_date_verifier(datetime(2020, month, date))
+                log.info("zodiac sign based on month and date received.")
+            except ValueError as e:
+                final_embed = discord.Embed()
+                final_embed.color = Colours.pink
+                final_embed.description = f"Zodiac sign is not found because, {e}"
+                log.info(f"Error in "zodiac date" command:\n{e}.")
```
AtieP, line 109: This ends in a `SyntaxError`.

thomaspet: Hah... Sorry about providing suggestions with invalid syntax

Anubhav1603: Np, I will fix this ,don't worry🙂

bot/exts/valentines/valentine_zodiac.py
Anubhav1603: actually i opted to make separate json file to keep content clean

bot/exts/valentines/valentine_zodiac.py
thomaspet: Did you miss this one?

Anubhav1603: ahh sorry

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+                zodiac_data['start_at'] = datetime.fromisoformat(zodiac_data['start_at'])
+                zodiac_data['end_at'] = datetime.fromisoformat(zodiac_data['end_at'])
+        with compatibility_file.open(encoding="utf8") as json_data:
+            zodiacs = json.load(json_data)
+        return zodiacs, zodiac_fact
 
-    @commands.command(name="partnerzodiac")
-    async def counter_zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
+    def generate_invalidname_embed(self, zodiac: str) -> discord.Embed:
+        """Returns error embed."""
```
thomaspet, line 42: What error embed? This should be a bit more descripting, like the fact that it's creating a new embed, and that its an embed to display an error message about invalid zodiac name provided.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+        )
+        embed.description = (
+            f"**{zodiac}** is not a valid zodiac sign, here is the list of valid zodiac signs.\n"
+            f"{error_comp}"
+        )
+        log.info("Invalid zodiac name provided.")
+        return embed
+
+    def zodiac_build_embed(self, zodiac: str) -> discord.Embed:
+        """Gives informative zodiac embed."""
```
thomaspet, line 56: It doesn't really "give" an embed as much as building an embed for either the facts OR error embed.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+            embed.title = f"__{zodiac}__"
+            embed.description = self.zodiac_fact[zodiac]["About"]
+            embed.add_field(name='__Full form__', value=self.zodiac_fact[zodiac]["full_form"], inline=False)
+            embed.add_field(name='__Motto__', value=self.zodiac_fact[zodiac]["Motto"], inline=False)
+            embed.add_field(name='__Strengths__', value=self.zodiac_fact[zodiac]["Strengths"], inline=False)
+            embed.add_field(name='__Weaknesses__', value=self.zodiac_fact[zodiac]["Weaknesses"], inline=False)
+            embed.set_thumbnail(url=self.zodiac_fact[zodiac]["url"])
+        else:
+            embed = self.generate_invalidname_embed(zodiac)
+        log.info("Zodiac embed ready.")
```
thomaspet, line 71: This log can probably be moved to inside the if with "Successfully created zodiac information embed" etc. As this log is sort of a false positive if we end up with an invalid name embed.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+        return embed
+
+    def zodiac_date_verifier(self, query_datetime: datetime) -> str:
+        """Returns zodiac sign by checking month and date."""
```
thomaspet, line 75: I wouldn't specify month as it's not specifically checking month, just date. 
A date can be refered to a full date, e.g 24.04.2020, or just the day. In this case date can be used for the full thing.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+
+    @zodiac.command(name="date")
+    async def date_and_month(self, ctx: commands.Context, date: int, month: Union[int, str]) -> None:
+        """Provides information about zodiac sign by taking month and date as input."""
+        if isinstance(month, str):
+            month = month.capitalize()
+            try:
+                month = list(calendar.month_abbr).index(month[:3])
+            except ValueError:
+                await ctx.send(f"Sorry, but `{month}` is not a valid month name.")
```
thomaspet, line 96: A log might be usefull here to see what the user tried to give when debugging

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+            try:
+                zodiac_sign_based_on_date = self.zodiac_date_verifier(datetime(2020, month, date))
+                log.info("zodiac sign based on month and date received.")
+            except ValueError as e:
+                final_embed = discord.Embed()
+                final_embed.color = Colours.pink
+                final_embed.description = f"Zodiac sign is not found because, {e}"
+                log.info(f'Error in "zodiac date" command:\n{e}.')
+            else:
+                final_embed = self.zodiac_build_embed(zodiac_sign_based_on_date)
```
thomaspet, line 111: I believe you can remove the else (dedent it to the commands scope), and reassign `zodiac` instead of creating `zodiac_sign_based_on_date` here. (Also remove `final_embed = self.zodiac_build_embed(zodiac)` on line 100)

The removes duplicate logic, removes an extra variable assignment, and less logic to follow.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+            zodiac = "capricorn"
+            final_embed = self.zodiac_build_embed(zodiac)
+        else:
+            try:
+                zodiac_sign_based_on_date = self.zodiac_date_verifier(datetime(2020, month, date))
+                log.info("zodiac sign based on month and date received.")
+            except ValueError as e:
+                final_embed = discord.Embed()
+                final_embed.color = Colours.pink
+                final_embed.description = f"Zodiac sign is not found because, {e}"
```
thomaspet, line 108:
```suggestion
                final_embed.description = f"Zodiac sign could not be found.\n\`\`\`\n{e}\`\`\`"
```
The wording seemed a bit too "personal" for an error message. Also code block the error to display it better.

thomaspet: (remove \ from \`) just github formatting being weird

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+                log.info(f'Error in "zodiac date" command:\n{e}.')
+            else:
+                final_embed = self.zodiac_build_embed(zodiac_sign_based_on_date)
+
+        await ctx.send(embed=final_embed)
+        log.info("Zodiac sign embed based on date is now sent.")
```
thomaspet, line 114: `log.info("Embed successfully sent.")` Is how you logged previous command, please be consisted with the log messages.
e.g: "Embed from date succesfully sent."

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+
+    @zodiac.command(name="partnerzodiac", aliases=['partner'])
+    async def partner_zodiac(self, ctx: commands.Context, zodiac_sign: str) -> None:
         """Provides a counter compatible zodiac sign to the given user's zodiac sign."""
```
thomaspet, line 118: a "counter compatible"?  sounds like something that counters the provided one, in a negative way. This is for how compatible the provided zodiac is with a ..random(?) one. 

could this also be modifed to have an optional parameter to provide a zodiac, else choose a random one?

(The docstring should specify that the other zodiac is selected at random)

thomaspet: I'm also guessing you didn't write this, but somehow claimed it. Would appreciate it if you looked at this, even if it isn't your code.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -19,36 +22,117 @@ class ValentineZodiac(commands.Cog):
...
+        explanation_file = Path("bot/resources/valentines/zodiac_explanation.json")
+        compatibility_file = Path("bot/resources/valentines/zodiac_compatibility.json")
+        with explanation_file.open(encoding="utf8") as json_data:
+            zodiac_fact = json.load(json_data)
+            for zodiac_data in zodiac_fact.values():
+                zodiac_data['start_at'] = datetime.fromisoformat(zodiac_data['start_at'])
+                zodiac_data['end_at'] = datetime.fromisoformat(zodiac_data['end_at'])
+        with compatibility_file.open(encoding="utf8") as json_data:
+            zodiacs = json.load(json_data)
+        return zodiacs, zodiac_fact
```
thomaspet, line 39:
```suggestion
                zodiac_data['end_at'] = datetime.fromisoformat(zodiac_data['end_at'])

        with compatibility_file.open(encoding="utf8") as json_data:
            zodiacs = json.load(json_data)

        return zodiacs, zodiac_fact
```
Purely cosmetic apperance. I find it easier to read these with a bit of whitespace, as whitespace is free.

thomaspet (changes requested):
Getting there, a few final touches suggestions. Mostly around docstring and logging, but remark or 2 about logic.
Keep it up :)

thomaspet:
This is just a suggestion, however probably a cool feature. You can create a date or zodiac converter that throws on invalid date & zodiac, and just converts the input (date or zodiac string) to the zodiac name. That way you don't need the date sub-command and you can implement it for the partnerzodiac command as well.

Anubhav1603:
> This is just a suggestion, however probably a cool feature. You can create a date or zodiac converter that throws on invalid date & zodiac, and just converts the input (date or zodiac string) to the zodiac name. That way you don't need the date sub-command and you can implement it for the partnerzodiac command as well.

actually i need date and month both ,current date subcommand can take str and int for month but if i merge this with this subcommand into `zodiac` then user will be forced to give month and date in int only

bot/exts/valentines/valentine_zodiac.py
Anubhav1603: > I believe you can remove the else (dedent it to the commands scope), and reassign `zodiac` instead of creating `zodiac_sign_based_on_date` here. (Also remove `final_embed = self.zodiac_build_embed(zodiac)` on line 100)
> 
> The removes duplicate logic, removes an extra variable assignment, and less logic to follow.

in case of value error final_embed will get error out

thomaspet:
> 
> 
> > This is just a suggestion, however probably a cool feature. You can create a date or zodiac converter that throws on invalid date & zodiac, and just converts the input (date or zodiac string) to the zodiac name. That way you don't need the date sub-command and you can implement it for the partnerzodiac command as well.
> 
> actually i need date and month both ,current date subcommand can take str and int for month but if i merge this with this subcommand into `zodiac` then user will be forced to give month and date in int only

That's why I suggested a converter. The date subcommand just uses the date to fetch a zodiac in the end, right? So your converter would first check if it's a string -> check if it's a zodiac, if it is, ok done, return the zodiac.
If it's not a zodiac it will do the month logic you have in the month sub-command. (see get the date from string or int, get the zodiac from the date and so on..)
This logic returns a zodiac, which in the end you can return from the converter as well. 
If all these fail you throw a convertexception which you can catch with an errorhandler and return whichever error fits from.

bot/exts/valentines/valentine_zodiac.py
thomaspet: You could just return in the excpetion and send the embed then, but i guess the benefits aren't that great anymore then.

bot/exts/valentines/valentine_zodiac.py
kwzrd: The logs here should also definitely be below `INFO` level, perhaps [`TRACE`](https://github.com/python-discord/seasonalbot/blob/master/CONTRIBUTING.md#logging-levels). There's no need to see in production logs that an embed was built.

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -105,35 +107,33 @@ async def date_and_month(self, ctx: commands.Context, date: int, month: Union[in
...
-                f'{zodiac_sign.capitalize()}{emoji1}{compatible_zodiac["Zodiac"]}\n'
-                f'{emoji2}Compatibility meter : {compatible_zodiac["compatibility_score"]}{emoji2}'
-            )
-            embed.add_field(
-                name=f'A letter from Dr.Zodiac {LETTER_EMOJI}',
-                value=compatible_zodiac['description']
-            )
-        except KeyError:
-            embed = self.generate_invalidname_embed(zodiac_sign)
+        compatible_zodiac = random.choice(self.zodiacs[zodiac_sign.capitalize()])
```
thomaspet, line 125: Won't this just throw now if you give an invalid zodiac sign?

Maybe assign the zodiacs to a variable using `self.zodiacs.get(zodiac_sign.capitalize())` and check if it's None before continueing instead of try/excepting.

thomaspet (changes requested):
I'm for making all logs a trace, except those in except block. As in we should be informed if something goes wrong, but keep it more or less quiet on info level if everything goes smoothly. 
Else i believe you've introduced a new bug in the partnerzodiac command

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -92,7 +92,9 @@ async def date_and_month(self, ctx: commands.Context, date: int, month: Union[in
             month = month.capitalize()
             try:
                 month = list(calendar.month_abbr).index(month[:3])
+                log.info('Valid month name entered by user')
             except ValueError:
+                log.info('Invalid month name entered by user')
```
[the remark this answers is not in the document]
thomaspet: Upper one trace, lower one warning or info

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -44,14 +44,16 @@ def generate_invalidname_embed(self, zodiac: str) -> discord.Embed:
...
-        )
+        error_msg = f"**{zodiac}** is not a valid zodiac sign, here is the list of valid zodiac signs.\n"
+        valid_zodiac_name = ''
+        for name in sorted(self.zodiac_fact.keys()):
+            if name == "Leo":
+                valid_zodiac_name += f'{name}\n '
+            elif name == 'Virgo':
+                valid_zodiac_name += name
+            else:
+                valid_zodiac_name += f'{name}, '
```
thomaspet, line 55: This seems a bit over engineered imo.
```suggestion
        valid_zodiac_name = ", ".join(self.zodiac_fact).replace("Leo, ", "Leo,\n")
```
Quick and dirty, given Leo is always written this way, and contains a space. So if it somehow ends up at the end it breaks.

So a better solution might just be to go by index
```py
", ".join(names[ :len(self.zodiac_fact) // 2]) + ",\n" + ", ".join(names[len(self.zodiac_fact) // 2: ])
```

thomaspet: This can be written a bit cleaner with 2 variables for first and last part and add them afterwards.

thomaspet (changes requested):

thomaspet:
Did you read the message, or just click merge? It's a dirty solution that will break if the order of the keys ever change, or the way they are written. It's pretty error prone.

Anubhav1603:
> Did you read the message, or just click merge? It's a dirty solution that will break if the order of the keys ever change, or the way they are written. It's pretty error prone.

I just read the code and clicked commit 😅, just noticed u have written something at comment also

bot/exts/valentines/valentine_zodiac.py
```diff
@@ -45,8 +45,9 @@ def generate_invalidname_embed(self, zodiac: str) -> discord.Embed:
         embed = discord.Embed()
         embed.color = Colours.soft_red
         error_msg = f"**{zodiac}** is not a valid zodiac sign, here is the list of valid zodiac signs.\n"
-        valid_zodiac_name = ", ".join(self.zodiac_fact).replace("Leo, ", "Leo,\n")
-        embed.description = error_msg + valid_zodiac_name
+        first_part_of_valid_name_list = ", ".join(name for name in list(self.zodiac_fact.keys())[:7])
+        end_part_of_valid_name_list = ", ".join(name for name in list(self.zodiac_fact.keys())[7:])
```
thomaspet, line 49: Instead of calling keys, and converting all to a list twice, create an extra variable for it. I also would suggest agaisnt hardcoded slicing, not because there could be more or less zodiacs, but for good practice. Since the original goal was to slice in the middle, there should be code that illustrates that intent. 7 is a magic number nobody reading this will know at the top of their head is the middle of all zodiacs (or at least close to it)

```suggestion
        names = list(self.zodiac_fact.keys())
        middle_index = len(names) // 2
        front_half_names = ", ".join(names[:middle_index])
        back_half_names = ", ".join(names[middle_index:])
```
For naming the variables, since i'm slicing a variable names now i dont need to say "valid names" as this variable is just related to the previous, so we can trace it that way. You also don't need to say list, I personally distinguish between a single value and a collection of values (like a list) by using singular and plurar names. Like "customer" and "customers", latter being multiple.
In our case names is already defined as a collection, and can move the naming convention further.

For the rest of the variable names, we're splitting something in half, so we can use "first half, last half" or "front half, back half" etc. shortening it a bit more. Long names might be descriptive, and sometimes required, but keeping short names that still are able to explain the variables purpose is prefered. In my opinion.

The naming stuff is a suggestion, not a requirement here. However you should break up the logic a bit (see example)

thomaspet (changes requested):

tagptroll1 (approved):
if my memory serves me right this should be good now

PR #456, opened 2020-09-22
updated d.py to 1.4.1 and added voice as extra

### LINKED ISSUE #432: Update discord.py version to 1.4.1
## Description
Update the discord.py version to 1.4.1, and also put the voice extra back in

## Reasoning
It currently gets the package straight from GitHub,
https://github.com/python-discord/seasonalbot/blob/da468c51502c88c00687e2869f05a1a46c715e45/Pipfile#L10
which was necessary to allow for usage of `AllowedMentions` before it was released, but wasn't ideal. Now that the new version has been released I believe this is no longer necessary. Currently it also doesn't include the `voice` extra which is required for the `.spookysound` command, so this change should fix that.
## Proposed Implementation
Change the line in the pipfile to `discord-py = {extras = ["voice"], version = "~=1.4.1"}` and lock it. I think that's all that needs to be done.

(I don't mind)

Anubhav1603:
"Closes #432 "

uninstalled old dpy and installed new d.py with voice as extra

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ENV PIP_NO_CACHE_DIR=false \
 # Install git to be able to dowload git dependencies in the Pipfile
 RUN apt-get -y update \
     && apt-get install -y \
-        git \
+        ffmpeg \
     && rm -rf /var/lib/apt/lists/*
 
 # Install pipenv
--- a/Pipfile
+++ b/Pipfile
@@ -7,12 +7,12 @@ name = "pypi"
 aiodns = "~=2.0"
 arrow = "~=0.14"
 beautifulsoup4 = "~=4.8"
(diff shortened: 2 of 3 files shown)


Den4200 (dismissed):
Is there a reason we need voice features from discord.py? I can't think of any cogs that use it off the top of my head. If it's truly not needed, we shouldn't install it.

Now that we don't need git anymore, make sure to also remove its installation from the Dockerfile.

Anubhav1603:
In [Issue 432](https://github.com/python-discord/seasonalbot/issues/432) reason is mentioned ,we require this for spooky sound command 😁

PureFunctor (approved):
Looks good

Akarys42 (approved):
Works like a charm, thanks!

PR #459, opened 2020-09-25
Make .bm handle embed-suppression syntax for message links

bast0006:
This PR adds support for embed-suppressed link syntax to the .bm command. Without this PR, running the command:
```
.bm <https://discordapp.com/channels/267624335836053506/635950537262759947/759110063553511435> Test
```
fails with an unhelpful error message:
bm-cmd-test

While this particular error doesn't display what exactly went wrong, doing the following makes it clearer:
bm-cmd-test-3

The discord.Message converter is not handling the <> and [] wrapping syntax. 

I've implemented a BetterMessageConverter in bot.utils.converters that unwraps [], <>, and [<>] from the argument before attempting a conversion to a message object. While this does provide stray support for <message-id> and [message-id], I'm not sure it's worth the effort to adjust how the parsing regex works, and this fixes the issue without introducing additional failure modes. 

The .bm command was then adjusted to use the new BetterMessageConverter.


bast0006:
And I need to figure out how to get pre-commit working. I'll get back to this in a few hours to fix the flake8 issue that my local config didn't catch.

thomaspet (approved):
Don't know what the `[]` syntax is used for here, other than you can markdown urls, however that's not done when bookmarking messages.
Else it looks good to me.

bot/utils/converters.py
```diff
@@ -0,0 +1,16 @@
+import discord
+from discord.ext.commands.converter import MessageConverter
+
+
+class BetterMessageConverter(MessageConverter):
```
Den4200, line 5: Perhaps a more descriptive name such as `MessageLinkConverter`?

bot/utils/converters.py
```diff
@@ -0,0 +1,16 @@
...
+    """A converter that handles embed-suppressed links like <http://example.com>."""
+
+    async def convert(self, ctx: discord.ext.commands.Context, argument: str) -> discord.Message:
+        """Wrap the commands.MessageConverter to handle <> delimited message links."""
+        # It's possible to wrap a message in [<>] as well, and it's supported because its easy
+        if argument.startswith("[") and argument.endswith("]"):
```
Den4200, line 11: Is there any reason to have square brackets around links? As far as I know, there is no reason to do that.

Den4200 (changes requested):

bot/utils/converters.py
bast0006: Perhaps, although "MessageLinkConverter" doesn't really match IMO, since this also handles message ID's, and pairs of ChannelID-MessageID. I couldn't think of a better name. WrappedMessageConverter?

bot/utils/converters.py
bast0006: It's similar to <>, in that it suppresses the embed, but unlike <> it remains visible in the message content.

Den4200: It doesn't seem to suppress the embed for me. I might be doing something wrong?

image

Anubhav1603: Use <> instead of []

bot/utils/converters.py
Den4200: Hmm that's true. `WrappedMessageConverter` is probably fine then.

bot/utils/converters.py
bast0006: I'm honestly not sure what is up with that. I was under the impression it worked on _all_ embeds, but apparently it doesn't. The testing I've done suggests that <> works for everything, but [] only seems to fail when applied to message links. 

I can take the check for [] out if desired, it was pretty opportunistic (there's space, so why not support it) although it looks like it's either bugged, or just doesn't work atm.

Den4200: How odd. I think it's fine to leave it in then.

Den4200 (approved):
Looks good to me, and works well. Thanks @bast0006!

PR #462, opened 2020-09-28
Add emojis cog

### LINKED ISSUE #461: Add cog that shows info about emojis
## Description
Add a command that displays relevant info about emojis, like name, ID, created_at, CDN link...

## Reasoning
People could get more information about the server emojis.

## Proposed Implementation
Simple: create a new command called `emoji_info` or something. It will get a second parameter called `emoji` typehinted with `discord.Emoji`. Then, proceed to show relevant info about the emote. discord.py itself will raise an error if the emote doesn't exist.

AtieP:
## Relevant Issues
Closes #461

## Description
I've created a cog (`Emojis`) that has an `emojis` group and that has an `info` command.

## Screenshots

Screenshot_2020-09-28_16-59-51

## Did you:


bot/exts/evergreen/emojis.py
```diff
@@ -0,0 +1,39 @@
...
+from bot.utils.time import time_since
+
+
+class Emojis(commands.Cog):
+    """Has commands related to emojis."""
+
+    @commands.group(name="emoji", invoke_without_command=True)
+    async def emojis_group(self, ctx: commands.Context, emoji: Emoji) -> None:
+        """A group of commands related to emojis."""
+        # No parameters = same as invoking info subcommand
```
thomaspet, line 15:
```suggestion
        # No parameters = same as invoking info subcommand
```
I don't think this comment adds anything, ctx.invoice(info_command) speaks for itself.

bot/exts/evergreen/emojis.py
```diff
@@ -0,0 +1,39 @@
...
+            description=textwrap.dedent(f"""
+                Name: {emoji.name}
+                Created: {time_since(emoji.created_at)}
+                ID: {emoji.id}
+                Emote source: [Link]({emoji.url})
+            """),
+            color=Color.blurple()
+        )
+        emoji_information.set_thumbnail(
+            url=str(emoji.url)
```
thomaspet, line 32: You can actually just give it `emoji.url` as set_thumbnail will internally call str before using it. Which would shorten it enough to one line it

AtieP: No, it won't work. `emoji.url` is `discord.Asset` and needs to be explicitly casted to `str`.

AtieP: ah actually wait

thomaspet: Which, as i said, set_thumbnail does internally. https://github.com/Rapptz/discord.py/blob/master/discord/embeds.py#L337

AtieP: Yeah yeah, I'm sorry. I had problems before when setting `title` to something other than `str` before and I thought here it would happen the same.

thomaspet (approved):
lgtm

bot/exts/evergreen/emojis.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class Emojis(commands.Cog):
+    """Has commands related to emojis."""
+
+    @commands.group(name="emoji", invoke_without_command=True)
+    async def emojis_group(self, ctx: commands.Context, emoji: Emoji) -> None:
+        """A group of commands related to emojis."""
+        # No parameters = same as invoking info subcommand
+        await ctx.invoke(self.info_command, emoji)
```
ks129, line 16: Not required to change it, but what about using `await self.info_command(...)` instead invoking?

gustavwilliam:
Looks good. I only have a few comments.

gustavwilliam:
The `Created` section is too detailed. I got `Created: 15 days, 3 hours, 28 minutes and 44 seconds ago ` during testing, which is way too much detail in almost all scenarios.

I think hours would be a reasonable level of detail to include.

ks129:
I think just displaying date would be okay, no need x days ago etc.

gustavwilliam:
As we discussed in #dev-contrib, I think having `Emote source` as a link would be better than including the link in the text "Link". Other possible wordings of this would be:
- Emoji source image
- Image source

Either way, making the entire text (like "Emoji source image") a link would probably be optimal.

gustavwilliam:
@ks129 days ago is the format used in some of our other commands as well. Keeping it consistent would be optimal. If we want to include the exact date, I think adding it in a parentheses after the "days after" could work.

AtieP:
I don't know how to change the `Created` string though, I am not who created the function.

bot/exts/evergreen/emojis.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @emojis_group.command(name="info")
+    async def info_command(self, ctx: commands.Context, emoji: Emoji) -> None:
+        """Returns relevant information about a Discord Emoji."""
+        emoji_information = Embed(
+            title=f'Information about "{emoji.name}"',
+            description=textwrap.dedent(f"""
+                Name: {emoji.name}
+                Created: {time_since(emoji.created_at)}
```
gustavwilliam, line 25:
```suggestion
                Created: {time_since(emoji.created_at, precision="hours")}
```
Setting this to "days" could also be an option, if it feels to verbose.

bot/exts/evergreen/emojis.py
```diff
@@ -0,0 +1,39 @@
+import textwrap
+
+from discord import Color, Embed, Emoji
+from discord.ext import commands
+
+from bot.utils.time import time_since
+
+
+class Emojis(commands.Cog):
+    """Has commands related to emojis."""
```
gustavwilliam, line 10: I'd love to see this explained a bit further. What does it actually do? All info in the docstring is already given by the class name.

gustavwilliam (changes requested):
Just a few minor comments

bot/exts/evergreen/emojis.py
AtieP: What's the output if the precision are hours?

bot/exts/evergreen/emojis.py
AtieP: I don't know, suggest something lol

bot/exts/evergreen/emojis.py
AtieP: > What's the output if the precision are hours?  
And the emoji was created in less than an hour?

gustavwilliam: The lowest detail that it'll include is hours. Like this:
`Created: 15 days and 3 hours ago `
Instead of:
`Created: 15 days, 3 hours, 36 minutes and 38 seconds ago`

AtieP: Ah, okay. I will accept this.

bot/exts/evergreen/emojis.py
gustavwilliam: Think about what the commands actually do. Feel free to take inspiration from the other docstrings in the repo.

bot/exts/evergreen/emojis.py
Den4200: We had a discussion about this in #dev-core before, and came to the conclusion that we should use the `invoke` method instead. It does some extra things that awaiting the command itself does not do.

bot/exts/evergreen/emojis.py
Den4200: As @thomaspet mentioned, this comment doesn't add anything, so it is not needed.
```suggestion
```

bot/utils/time.py
```diff
@@ -0,0 +1,84 @@
...
+
+    precision specifies the smallest unit of time to include (e.g. "seconds", "minutes").
+    max_units specifies the maximum number of units of time to include (e.g. 1 may include days but not hours).
+    """
+    now = datetime.datetime.utcnow()
+    delta = abs(relativedelta(now, past_datetime))
+
+    humanized = humanize_delta(delta, precision, max_units)
+
+    return f"{humanized} ago"
```
Den4200, line 84: There shouldn't be any need for these utils, since we have the `arrow` installed. Try using that instead.

bot/exts/evergreen/emojis.py
```diff
@@ -0,0 +1,39 @@
...
+                Name: {emoji.name}
+                Created: {time_since(emoji.created_at, precision="hours")}
+                ID: {emoji.id}
+                [Emoji source image]({emoji.url})
+            """),
+            color=Color.blurple()
+        )
+        emoji_information.set_thumbnail(
+            url=emoji.url
+        )
```
Den4200, line 33: I don't see a reason to have this span over multiple lines.
```suggestion
        emoji_information.set_thumbnail(url=emoji.url)
```

Den4200 (changes requested):
This was purely a code review, I'll test out this new command soon.

Xithrius:
@AtieP Poking. Will you be doing the requested changes in the future?

Xithrius:
User no longer seems to be active. If anyone wants to take over, let me know.

bot/utils/time.py
kosayoda: Our version works a lot better, the `arrow` version is clunky and not customizable.

kosayoda:
Emoji related commands are now under the same cog+command group, `emoji`.

`.emojicount` is now `.emoji count`
image

Emoji information is now `.emoji info`, and the embed is changed slightly.
image

`.emoji` brings up the help, and `.emoji <emoji>` is a shortcut to `.emoji info <emoji>`

Xithrius (approved):

bot/utils/time.py
```diff
@@ -0,0 +1,84 @@
+import datetime
+
+from dateutil.relativedelta import relativedelta
+
+
+# All these functions are from https://github.com/python-discord/bot/blob/master/bot/utils/time.py
+def _stringify_time_unit(value: int, unit: str) -> str:
```
Shivansh-007, line 7: Arrow library has `humanize()` function which can do this.

bot/utils/time.py
```diff
@@ -0,0 +1,84 @@
+import datetime
+
+from dateutil.relativedelta import relativedelta
+
+
+# All these functions are from https://github.com/python-discord/bot/blob/master/bot/utils/time.py
```
Shivansh-007, line 6: I don't think this comment is needed.

bot/utils/time.py
```diff
@@ -0,0 +1,84 @@
...
+        return "0 seconds"
+    elif value == 1:
+        return f"{value} {unit[:-1]}"
+    elif value == 0:
+        return f"less than a {unit[:-1]}"
+    else:
+        return f"{value} {unit}"
+
+
+def humanize_delta(delta: relativedelta, precision: str = "seconds", max_units: int = 6) -> str:
```
Shivansh-007, line 28: Arrow library has `humanize()` function which can do this.

bot/exts/evergreen/emoji.py
```diff
@@ -1,31 +1,33 @@
...
 from bot.constants import Colours, ERROR_REPLIES
 from bot.utils.pagination import LinePaginator
+from bot.utils.time import time_since
 
 log = logging.getLogger(__name__)
 
 
-class EmojiCount(commands.Cog):
-    """Command that give random emoji based on category."""
+class Emoji(commands.Cog):
```
Shivansh-007, line 18: Is overwriting the import `discord.Emoji`.

bot/exts/evergreen/emoji.py
```diff
@@ -1,31 +1,33 @@
...
         self.bot = bot
 
     @staticmethod
-    def embed_builder(emoji: dict) -> Tuple[discord.Embed, List[str]]:
+    def embed_builder(emoji: dict) -> Tuple[Embed, List[str]]:
         """Generates an embed with the emoji names and count."""
-        embed = discord.Embed(
+        embed = Embed(
             color=Colours.orange,
             title="Emoji Count",
```
Shivansh-007, line 29: Please end sentences in full stops.

Do this for the other cases too. Not commenting everywhere.

bot/exts/evergreen/emoji.py
```diff
@@ -67,8 +69,16 @@ def generate_invalid_embed(emojis: list) -> Tuple[discord.Embed, List[str]]:
         msg.append(f"These are the valid categories\n```{error_comp}```")
         return embed, msg
 
-    @commands.command(name="emojicount", aliases=["ec", "emojis"])
-    async def emoji_count(self, ctx: commands.Context, *, category_query: str = None) -> None:
+    @commands.group(name="emoji", invoke_without_command=True)
+    async def emoji_group(self, ctx: commands.Context, emoji: Optional[Emoji]) -> None:
+        """A group of commands related to emojis."""
+        if emoji is not None:
```
Shivansh-007, line 75:
```suggestion
        if emoji:
```

Shivansh-007 (changes requested):

bot/utils/time.py
scragly: I haven't looked into the differences so this isn't my opinion on it, but I noticed this touches a previously reviewed aspect, which you can find kosa's response to here:

> Our version works a lot better, the `arrow` version is clunky and not customizable.

https://github.com/python-discord/sir-lancebot/pull/462#discussion_r588906493

Shivansh-007: Alright! Didn't see the earlier reviews.

bot/exts/evergreen/emoji.py
Xithrius: If the user doesn't input a name for this argument, `emoji` will be explicitly `None`. I think it's better to say that's what we're checking for.

bot/exts/evergreen/emoji.py
scragly: Titles shouldn't end in periods, it's unconventional unless it's a fully structured sentence.

Shivansh-007: Alright! Then it is meant for the other parts of the code. There are 2-3 cases.

bot/exts/evergreen/emoji.py
Shivansh-007: Well `if emoji:` is the same as `if emoji is not None:`. And according to me `if emoji:` seems the better way

scragly: Let's avoid arguing over code style opinions unless the style is clearly defined in our projects. 

Since both works, and there's not really an explicit bias in our projects other than a slight lean towards "explicit is better than implicit", there's no need to request a change for this.

bot/exts/evergreen/emoji.py
scragly: I had a look over the code but I'm unable to see more than 1 string that may require a period to be added.

If you meant for each of the list items in the `.emoji info` embed, they're not really at the level of a fully constructed sentence either. The closest may be the Created line due to humanisation, but I think it's too ambiguous to enforce and would clash with the rest of the output as it would be the only line to do so. Better to just leave it in my opinion. 

If there was any other places and I've just missed them, it may be best to specify, apologies.

scragly: Actually on a detailed look, it appears that even that one I noticed isn't part of the diff here, instead looks to be from previous commits, so there may be more from code outside of the range I'm reading over.

Shivansh-007: Not part of this PR (5th one is part of this PR). But will be good if it would be done as they part of the `emoji` command:

1. https://github.com/python-discord/sir-lancebot/blob/af22fe556a3b1ab8a98245cc8f96e1d81d0f4289/bot/exts/evergreen/emoji.py#L37

2. https://github.com/python-discord/sir-lancebot/blob/af22fe556a3b1ab8a98245cc8f96e1d81d0f4289/bot/exts/evergreen/emoji.py#L39

3. https://github.com/python-discord/sir-lancebot/blob/af22fe556a3b1ab8a98245cc8f96e1d81d0f4289/bot/exts/evergreen/emoji.py#L46

4. https://github.com/python-discord/sir-lancebot/blob/af22fe556a3b1ab8a98245cc8f96e1d81d0f4289/bot/exts/evergreen/emoji.py#L48

5. https://github.com/python-discord/sir-lancebot/blob/af22fe556a3b1ab8a98245cc8f96e1d81d0f4289/bot/utils/time.py#L84

bot/utils/time.py
kosayoda: See: https://github.com/python-discord/sir-lancebot/pull/462#discussion_r496356447

bot/utils/time.py
kosayoda: I don't see the harm of having the original source linked, so I'm inclined to keep it.

bot/exts/evergreen/emoji.py
kosayoda: Good catch!

bot/exts/evergreen/emoji.py
kosayoda: The only thing suggested here that's part of my PR is not a sentence, so I don't see a reason for the period at the end. I'll tweak the other ones.

Shivansh-007 (approved):
Thanks! Looks Good To Me!

Tested locally.

scragly (approved):

PR #468, opened 2020-10-01
Hacktoberfest Stats: Check GitHub user exists before searching for PRs

### LINKED ISSUE #467: GitHub API request for 'some_user_that_doesn't_exist' failed with message: The listed users canno...
Sentry Issue: [SEASONALBOT-1E](https://sentry.io/organizations/python-discord/issues/1925901672/?referrer=github_integration)

```
GitHub API request for 'some_user_that_doesn't_exist' failed with message: The listed users cannot be searched either because the users do not exist or you do not have permission to view the users.
```

Den4200:
## Description
The `.hackstats` command logs an error when a given username does not exist on GitHub. This is not an actual issue, so we want to avoid logging this as an error. To prevent this, we should check the user exists before searching for their PRs.

I've also added authentication to the GitHub API requests in this cog and slightly changed the user agent.

Closes #467.

### DIFF
--- a/bot/exts/halloween/hacktoberstats.py
+++ b/bot/exts/halloween/hacktoberstats.py
@@ -24,6 +24,11 @@ REQUEST_HEADERS = {"User-Agent": "Python Discord Hacktoberbot"}
 if GITHUB_TOKEN := Tokens.github:
     REQUEST_HEADERS["Authorization"] = f"token {GITHUB_TOKEN}"
 
+GITHUB_NONEXISTENT_USER_MESSAGE = (
+    "The listed users cannot be searched either because the users do not exist "
+    "or you do not have permission to view the users."
+)
+
 
 class HacktoberStats(commands.Cog):
     """Hacktoberfest statistics Cog."""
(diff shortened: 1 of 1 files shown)


bot/exts/halloween/hacktoberstats.py
```diff
@@ -242,9 +246,13 @@ async def get_october_prs(github_username: str) -> List[dict]:
             f"&per_page={per_page}"
         )
 
-        headers = {"user-agent": "Discord Python Hacktoberbot"}
         async with aiohttp.ClientSession() as session:
-            async with session.get(query_url, headers=headers) as resp:
+            async with session.get(f"https://api.github.com/users/{github_username}", headers=REQUEST_HEADERS) as resp:
+                if resp.status == 404:
+                    logging.debug(f"No GitHub user found named '{github_username}'")
+                    return
```
scragly, line 253: This seems to be an unnecessary extra call for every command usage; we already know if it's a user we can't read based on the response we get with the original request, we just need to avoid logging it as an error as it's an intended behaviour.

scragly (changes requested):

bot/exts/halloween/hacktoberstats.py
Den4200: Yeah, I actually thought about that right after I submitted the PR. I've unnecessarily added more complexity and an extra request per command invocation. Addressed in deee8cf!

scragly (approved):
Looks good to me.

Senjan21 (approved):

PR #478, opened 2020-10-04
Bump 'discord.py' to 1.5.1 and use explicit intents

kwzrd:
After https://github.com/python-discord/bot/pull/1196, we should make sure SeasonalBot is also ready for the upcoming gateway intents changes. This PR actions the dependency upgrade.

~~Amongst @python-discord/core-developers we need to decide whether we will require the priviledged members intent as we do for the Python bot. If we do not use it, the (default) member converter will not work, and it will therefore not be possible to convert user IDs into Member objects.~~

With discord.py [`1.5.1`](https://discordpy.readthedocs.io/en/latest/whats_new.html#v1-5-1), the Member converter now does lazily fetch Member instances if necessary, so that should allow us to circumvent this issue entirely. I have adjusted the PR to bump directly to `1.5.1`.

Feel free to push commits to this PR.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,7 +12,7 @@ pillow = "~=7.2"
 pytz = "~=2019.2"
 sentry-sdk = "~=0.14.2"
 PyYAML = "~=5.3.1"
-"discord.py" = {extras = ["voice"], version = "~=1.5.0"}
+"discord.py" = {extras = ["voice"], version = "~=1.5.1"}
 
 [dev-packages]
 flake8 = "~=3.8"
--- a/Pipfile
+++ b/Pipfile
@@ -12,7 +12,7 @@ pillow = "~=7.2"
 pytz = "~=2019.2"
 sentry-sdk = "~=0.14.2"
 PyYAML = "~=5.3.1"
-"discord.py" = {extras = ["voice"], version = "~=1.4.1"}
+"discord.py" = {extras = ["voice"], version = "~=1.5.0"}
 
 [dev-packages]
 flake8 = "~=3.8"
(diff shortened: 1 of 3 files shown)


kwzrd:
We don't have too much time until the 7th so I propose that for now we keep use the privileged Guild Members intent. It doesn't make things worse in any way (we're disabling other intents, so we'll already see less traffic). In the future we may want to disable it but there should probably be more conversation w.r.t. that (CC: @scragly).

I mostly copied the setup from bot (see description for PR link), but I've disabled `typing` events everywhere (this only disables the event, bot can still send typing), and also disabled `ban` intents.

From a quick test: bot starts, reacts to cmds and events in both guild and DMs.

Senjan21 (approved):
Looks good, if we want to disable members intents we can do it at later date.

kwzrd:
This should relock to [`1.5.1`](https://discordpy.readthedocs.io/en/latest/whats_new.html#v1-5-1) which may also make the member intent redundant for d.py will now lazily fetch members for the member convertor.

kwzrd:
I have relocked straight to [`1.5.1`](https://discordpy.readthedocs.io/en/latest/whats_new.html#v1-5-1) (2fa29e3) which seems to circumvent the Member converter issue (d.py now fetches what's needed automatically). I think this essentially means that we don't need the Member intent, so I have disabled it in 59a9811. My testing indicates that the converter works fine:

evidence_of_my_selfless_work

First invocation is on `1.5.0`, second on `1.5.1`, both with the Member intent disabled. I adjusted the PR descr & name to reflect this.

Senjan21 (approved):
Looks good, thanks!

jb3 (approved):
Looks good

PR #479, opened 2020-10-04
[Bug] emoji command

Anubhav1603:
# Description
I have added Linepaginator and added check for animated emoji, so now if embed description limit exceeds bot automatically creates new embed and add pagination emojies to it

### DIFF
--- a/bot/exts/evergreen/emoji_count.py
+++ b/bot/exts/evergreen/emoji_count.py
@@ -31,19 +31,19 @@ class EmojiCount(commands.Cog):
 
         if len(emoji) == 1:
             for category_name, category_emojis in emoji.items():
-                msg.append(
-                    f"There {'is' if len(category_emojis)==1 else 'are'} **{len(category_emojis)}** "
-                    f"{'emoji' if len(category_emojis)==1 else 'emojis'} in the **{category_name}** category"
-                )
+                if len(category_emojis) == 1:
+                    msg.append(f"There is **{len(category_emojis)}** emoji in **{category_name}** category")
+                else:
+                    msg.append(f"There are **{len(category_emojis)}** emojis in **{category_name}** category")
                 embed.set_thumbnail(url=random.choice(category_emojis).url)
 
         else:
             for category_name, category_emojis in emoji.items():
                 emoji_choice = random.choice(category_emojis)
-                emoji_info = (
-                    f"There {'is' if len(category_emojis) == 1 else 'are'} **{len(category_emojis)}** "
-                    f"{'emoji' if len(category_emojis) == 1 else 'emojis'} in the **{category_name}** category"
-                )
+                if len(category_emojis) > 1:
+                    emoji_info = f"There are **{len(category_emojis)}** emojis in **{category_name}** category"
+                else:
+                    emoji_info = f"There is **{len(category_emojis)}** emoji in **{category_name}** category"
                 if emoji_choice.animated:
                     msg.append(f'<a:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}')
                 else:
--- a/bot/exts/evergreen/emoji_count.py
+++ b/bot/exts/evergreen/emoji_count.py
@@ -61,8 +61,8 @@ class EmojiCount(commands.Cog):
         msg.append(f"These are the valid categories\n```{error_comp}```")
         return embed, msg
 
-    @commands.command(name="emoji_count", aliases=["ec"])
-    async def emoji_count(self, ctx: commands.Context, *, category_query: str = None) -> None:
+    @commands.command(name="emojis_count", aliases=["ec"])
+    async def emojis_count(self, ctx: commands.Context, *, category_query: str = None) -> None:
         """Returns embed with emoji category and info given by the user."""
         emoji_dict = defaultdict(list)
 
--- a/bot/exts/evergreen/emoji_count.py
+++ b/bot/exts/evergreen/emoji_count.py
@@ -31,7 +31,10 @@ class EmojiCount(commands.Cog):
 
         if len(emoji) == 1:
             for category_name, category_emojis in emoji.items():
-                msg.append(f"There is **{len(category_emojis)}** emoji in the **{category_name}** category")
+                msg.append(
+                    f"There {'is' if len(category_emojis)==1 else 'are'} **{len(category_emojis)}** "
+                    f"{'emoji' if len(category_emojis)==1 else 'emojis'} in the **{category_name}** category"
+                )
                 embed.set_thumbnail(url=random.choice(category_emojis).url)
 
         else:
--- a/bot/exts/evergreen/emoji_count.py
+++ b/bot/exts/evergreen/emoji_count.py
@@ -29,7 +29,7 @@ class EmojiCount(commands.Cog):
 
         if len(emoji) == 1:
             for key, value in emoji.items():
-                msg.append(f"There are **{len(value)}** emojis in the **{key}** category")
+                msg.append(f"There is **{len(value)}** emoji in the **{key}** category")
                 embed.set_thumbnail(url=random.choice(value).url)
 
         else:
--- a/bot/exts/evergreen/emoji_count.py
+++ b/bot/exts/evergreen/emoji_count.py
@@ -7,6 +7,7 @@ import discord
 from discord.ext import commands
 
 from bot.constants import Colours, ERROR_REPLIES
+from bot.utils.pagination import LinePaginator
 
 log = logging.getLogger(__name__)
 
@@ -27,8 +28,9 @@ class EmojiCount(commands.Cog):
 
         if len(emoji) == 1:
             for key, value in emoji.items():
-                embed.description = f"There are **{len(value)}** emojis in the **{key}** category"
+                msg = f"There are **{len(value)}** emojis in the **{key}** category"
                 embed.set_thumbnail(url=random.choice(value).url)
+                embed.description = msg
         else:
             msg = ''
             for key, value in emoji.items():
@@ -39,7 +41,7 @@ class EmojiCount(commands.Cog):
                 else:
                     msg += f'<:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}'
             embed.description = msg
-        return embed
+        return embed, msg
 
     @staticmethod
     def generate_invalid_embed(ctx: commands.Context) -> discord.Embed:
@@ -54,8 +56,8 @@ class EmojiCount(commands.Cog):
             emoji_dict[emoji.name.split("_")[0]] = []
 
         error_comp = ', '.join(key for key in emoji_dict.keys())
-        embed.description = f"These are the valid categories\n```{error_comp}```"
-        return embed
+        msg = f"These are the valid categories\n```{error_comp}```"
+        return embed, msg
 
     def emoji_list(self, ctx: commands.Context, categories: dict) -> Dict:
         """Generates an embed with the emoji names and count."""
@@ -83,10 +85,11 @@ class EmojiCount(commands.Cog):
         emoji_dict = self.emoji_list(ctx, emoji_dict)
 
         if len(emoji_dict) == 0:
-            embed = self.generate_invalid_embed(ctx)
+            embed, msg = self.generate_invalid_embed(ctx)
         else:
-            embed = self.embed_builder(emoji_dict)
-        await ctx.send(embed=embed)
+            embed, msg = self.embed_builder(emoji_dict)
+        msg = msg.split("\n")
+        await LinePaginator.paginate([line for line in msg], ctx=ctx, embed=embed)
 
 
 def setup(bot: commands.Bot) -> None:
--- a/bot/exts/evergreen/emoji_count.py
+++ b/bot/exts/evergreen/emoji_count.py
@@ -2,7 +2,7 @@ import datetime
 import logging
 import random
 from collections import defaultdict
-from typing import Tuple
+from typing import List, Tuple
 
 import discord
 from discord.ext import commands
@@ -20,7 +20,7 @@ class EmojiCount(commands.Cog):
         self.bot = bot
 
     @staticmethod
-    def embed_builder(emoji: dict) -> Tuple[discord.Embed, str]:
+    def embed_builder(emoji: dict) -> Tuple[discord.Embed, List[str]]:
         """Generates an embed with the emoji names and count."""
         embed = discord.Embed(
             color=Colours.orange,
@@ -37,7 +37,7 @@ class EmojiCount(commands.Cog):
         else:
             for category_name, category_emojis in emoji.items():
                 emoji_choice = random.choice(category_emojis)
-                emoji_info = f'There are **{len(category_emojis)}** emojis in the **{category_name}** category\n'
+                emoji_info = f'There are **{len(category_emojis)}** emojis in the **{category_name}** category'
                 if emoji_choice.animated:
                     msg.append(f'<a:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}')
                 else:
@@ -45,7 +45,7 @@ class EmojiCount(commands.Cog):
         return embed, msg
 
     @staticmethod
-    def generate_invalid_embed(ctx: commands.Context) -> Tuple[discord.Embed, str]:
+    def generate_invalid_embed(emojis: list) -> Tuple[discord.Embed, List[str]]:
         """Generates error embed."""
         embed = discord.Embed(
             color=Colours.soft_red,
@@ -54,15 +54,15 @@ class EmojiCount(commands.Cog):
         msg = []
 
         emoji_dict = defaultdict(list)
-        for emoji in ctx.guild.emojis:
+        for emoji in emojis:
             emoji_dict[emoji.name.split("_")[0]].append(emoji)
 
         error_comp = ', '.join(emoji_dict)
         msg.append(f"These are the valid categories\n```{error_comp}```")
         return embed, msg
 
-    @commands.command(name="emojis_count", aliases=["ec"])
-    async def emojis_count(self, ctx: commands.Context, *, category_query: str = None) -> None:
+    @commands.command(name="emoji_count", aliases=["ec", "emojis"])
+    async def emoji_count(self, ctx: commands.Context, *, category_query: str = None) -> None:
         """Returns embed with emoji category and info given by the user."""
         emoji_dict = defaultdict(list)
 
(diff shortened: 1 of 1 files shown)


bot/exts/evergreen/emoji_count.py
```diff
@@ -27,16 +28,20 @@ def embed_builder(self, emoji: dict) -> discord.Embed:
...
                 emoji_choice = random.choice(value)
                 emoji_info = f'There are **{len(value)}** emojis in the **{key}** category\n'
-                msg += f'<:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}'
+                if emoji_choice.animated:
+                    msg += f'<a:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}'
+                else:
+                    msg += f'<:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}'
             embed.description = msg
-        return embed
+        return embed, msg
```
scragly, line 44: Your return typehint for this method no longer matches what you're returning.

```py
    def embed_builder(self, emoji: dict) -> discord.Embed:
```

scragly (dismissed):

bot/exts/evergreen/emoji_count.py
Anubhav1603: Anything else need fixing @scragly except this?

bot/exts/evergreen/emoji_count.py
```diff
@@ -80,10 +85,11 @@ async def ec(self, ctx: commands.Context, *, emoji: str = None) -> Optional[str]
...
 
         if len(emoji_dict) == 0:
-            embed = self.generate_invalid_embed(ctx)
+            embed, msg = self.generate_invalid_embed(ctx)
         else:
-            embed = self.embed_builder(emoji_dict)
-        await ctx.send(embed=embed)
+            embed, msg = self.embed_builder(emoji_dict)
+        msg = msg.split("\n")
+        await LinePaginator.paginate([line for line in msg], ctx=ctx, embed=embed)
```
thomaspet, line 92: Why not make the embed builders return a Linepaginator, instead of embed + message. Just do the split and ect before returning.

bot/exts/evergreen/emoji_count.py
```diff
@@ -80,10 +85,11 @@ async def ec(self, ctx: commands.Context, *, emoji: str = None) -> Optional[str]
         emoji_dict = self.emoji_list(ctx, emoji_dict)
```
thomaspet, line 85: Can't this be done just above this? If you use a defaultdict you dont need to add a list to the dict at all, it will create one if it doesnt exist
```py
emoji_dict = defaultdict(list)

for emoji in emojis:
    emoji_dict[emoji_prefix].append(emoji)
```
etc

thomaspet (changes requested):
Few other remarks I don't like, which aren't part of the PR so i can't comment on
Change the a variable name `a`, that's not descriptive at all
```py
for a in ctx.guild.emojis: 
```

I'm pretty sure all other commands return None, and should have a descriptive function name.
"ec" is not a good functon name, we use name, and aliases for what the command it self should be called
```py
async def ec(self, ctx: commands.Context, *, emoji: str = None) -> Optional[str]:
```

why are you using emoji_dict.update over just assigning to key? 
```py
emoji_dict.update({a.name.split("_")[0]: []})
```

This does nothing more than create a new dict, which is the exact same as categories
```py
out = {category: [] for category in categories}
```

Use a defaultdict, and you don't need the emoji_list function.
And why not create the paginator in the embed functions (with proper renaming, and typehinting)

bot/exts/evergreen/emoji_count.py
Anubhav1603: linepagintor only takes maxlines and maxsize

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,65 @@ class EmojiCount(commands.Cog):
...
                 log.trace("Emoji Category not provided by the user")
-                emoji_dict.update({a.name.split("_")[0]: []})
-            elif a.name.split("_")[0] in emoji:
+                emoji_dict[emoji.name.split("_")[0]].append(emoji)
+            elif emoji.name.split("_")[0] in emoji_category:
                 log.trace("Emoji Category provided by the user")
-                emoji_dict.update({a.name.split("_")[0]: []})
-
-        emoji_dict = self.emoji_list(ctx, emoji_dict)
+                emoji_dict[emoji.name.split("_")[0]].append(emoji)
```
thomaspet, line 73: This will spam the logs based on how many emojis exist..
Just do the log first, then logic as to what you want to ignore instead of not ignore.

```py
log.trace(f"Emoji Category {'' if category_query else 'not '}provided by the user")
for emoji in ctx.guild.emojis:
    emoji_category = emoji.name.split("_")[0]

    if category_query is not None and emoji_category not in category_query:
        continue

    emoji_dict[emoji_category].append(emoji)
```
renamed emoji_category to category_query in this example, also included a suggestion to how to do the logic.

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,67 @@ class EmojiCount(commands.Cog):
...
         )
+        msg = []
 
-        emoji_dict = {}
+        emoji_dict = defaultdict(list)
         for emoji in ctx.guild.emojis:
-            emoji_dict[emoji.name.split("_")[0]] = []
+            emoji_dict[emoji.name.split("_")[0]].append(emoji)
 
         error_comp = ', '.join(key for key in emoji_dict.keys())
```
thomaspet, line 58:
```suggestion
        error_comp = ', '.join(key for key in emoji_dict)
```
It will index keys implictly

thomaspet:
With that last suggestion I believe it looks as long as you've done extensive testing locally.

thomaspet (approved):

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,67 @@ class EmojiCount(commands.Cog):
     def __init__(self, bot: commands.Bot):
         self.bot = bot
 
-    def embed_builder(self, emoji: dict) -> discord.Embed:
+    def embed_builder(self, emoji: dict) -> [discord.Embed, str]:
```
RohanJnr, line 21: This method can be a static method.

RohanJnr (changes requested):

RohanJnr:
The response when no emojis are found can be improved.
This is how the current response looks like for no emojis
image

You can add a check in `emoji_count()` method
```python
if not ctx.guild.emojis:
    await ctx.send("No emojis found.")
    return
```

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,67 @@ class EmojiCount(commands.Cog):
...
             color=Colours.orange,
             title="Emoji Count",
             timestamp=datetime.datetime.utcnow()
         )
+        msg = []
 
         if len(emoji) == 1:
             for key, value in emoji.items():
-                embed.description = f"There are **{len(value)}** emojis in the **{key}** category"
+                msg.append(f"There are **{len(value)}** emojis in the **{key}** category")
```
RohanJnr, line 32: Use `is` instead of `are` because there is only 1 emoji
```suggestion
                msg.append(f"There is **{len(value)}** emoji in the **{key}** category")
```

RohanJnr (changes requested):

bot/exts/evergreen/emoji_count.py
RohanJnr: actually, you have to check if len(value) is equal to 1, only then use `is`.

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,67 @@ class EmojiCount(commands.Cog):
...
+                if emoji_choice.animated:
+                    msg.append(f'<a:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}')
+                else:
+                    msg.append(f'<:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}')
+        return embed, msg
 
     @staticmethod
-    def generate_invalid_embed(ctx: commands.Context) -> discord.Embed:
+    def generate_invalid_embed(ctx: commands.Context) -> [discord.Embed, str]:
         """Genrates error embed."""
```
RohanJnr, line 47: Should be `generates`
```suggestion
        """Generates error embed."""
```

RohanJnr (changes requested):

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,71 @@ class EmojiCount(commands.Cog):
     def __init__(self, bot: commands.Bot):
         self.bot = bot
 
-    def embed_builder(self, emoji: dict) -> discord.Embed:
+    @staticmethod
+    def embed_builder(emoji: dict) -> [discord.Embed, str]:
```
thomaspet, line 22: This should technically be `-> typing.Tuple[Discord.Embed, str]`

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,71 @@ class EmojiCount(commands.Cog):
...
-        """Generates an embed with the emoji names and count."""
-        out = {category: [] for category in categories}
+            emoji_dict[emoji.name.split("_")[0]].append(emoji)
 
-        for emoji in ctx.guild.emojis:
-            category = emoji.name.split('_')[0]
-            if category in out:
-                out[category].append(emoji)
-        return out
+        error_comp = ', '.join(emoji_category for emoji_category in emoji_dict)
```
thomaspet, line 59:
```suggestion
        error_comp = ', '.join(emoji_dict)
```
This should be enough

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,71 @@ class EmojiCount(commands.Cog):
...
+                if emoji_choice.animated:
+                    msg.append(f'<a:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}')
+                else:
+                    msg.append(f'<:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}')
+        return embed, msg
 
     @staticmethod
-    def generate_invalid_embed(ctx: commands.Context) -> discord.Embed:
-        """Genrates error embed."""
+    def generate_invalid_embed(ctx: commands.Context) -> [discord.Embed, str]:
```
thomaspet, line 47:
```suggestion
    def generate_invalid_embed(ctx: commands.Context) -> Tuple[discord.Embed, str]:
```
Imported from typing

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +18,71 @@ class EmojiCount(commands.Cog):
...
-        for emoji in ctx.guild.emojis:
-            category = emoji.name.split('_')[0]
-            if category in out:
-                out[category].append(emoji)
-        return out
+        error_comp = ', '.join(emoji_category for emoji_category in emoji_dict)
+        msg.append(f"These are the valid categories\n```{error_comp}```")
+        return embed, msg
 
     @commands.command(name="emoji_count", aliases=["ec"])
```
thomaspet, line 63: Maybe an alias for `.emojis` ?  I can't think of any other command using that, and possible `.emoji` & `.emojis_count`

thomaspet (changes requested):

bot/exts/evergreen/emoji_count.py
Anubhav1603: Atie is working on emoji command #462 on different pr

bot/exts/evergreen/emoji_count.py
```diff
@@ -61,8 +61,8 @@ def generate_invalid_embed(ctx: commands.Context) -> Tuple[discord.Embed, str]:
         msg.append(f"These are the valid categories\n```{error_comp}```")
         return embed, msg
 
-    @commands.command(name="emoji_count", aliases=["ec"])
-    async def emoji_count(self, ctx: commands.Context, *, category_query: str = None) -> None:
+    @commands.command(name="emojis_count", aliases=["ec"])
```
thomaspet, line 64: My bad if i explained it badly, but what I meant was to add more aliases, not change the name. The name was fine.
Aliases like `emojis` or `emojis_count` is what i wanted to add.  Since I don't like having to use `_` when calling commands & `ec` is like a hotkey for seasoned keyboard ninjas. Having a middleground like emojis would help.

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +19,71 @@ class EmojiCount(commands.Cog):
     def __init__(self, bot: commands.Bot):
         self.bot = bot
 
-    def embed_builder(self, emoji: dict) -> discord.Embed:
+    @staticmethod
+    def embed_builder(emoji: dict) -> Tuple[discord.Embed, str]:
```
RohanJnr, line 23: this should be
```suggestion
    def embed_builder(emoji: dict) -> Tuple[discord.Embed, List[str]]:
```
make sure to import `List` from `Typing`

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +19,71 @@ class EmojiCount(commands.Cog):
...
+                if emoji_choice.animated:
+                    msg.append(f'<a:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}')
+                else:
+                    msg.append(f'<:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}')
+        return embed, msg
 
     @staticmethod
-    def generate_invalid_embed(ctx: commands.Context) -> discord.Embed:
-        """Genrates error embed."""
+    def generate_invalid_embed(ctx: commands.Context) -> Tuple[discord.Embed, str]:
```
RohanJnr, line 48: same thing here
```suggestion
    def generate_invalid_embed(ctx: commands.Context) -> Tuple[discord.Embed, List[str]]:
```

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +19,71 @@ class EmojiCount(commands.Cog):
...
+            if category_query is not None and emoji_category not in category_query:
+                continue
 
-        emoji_dict = self.emoji_list(ctx, emoji_dict)
+            emoji_dict[emoji_category].append(emoji)
 
         if len(emoji_dict) == 0:
-            embed = self.generate_invalid_embed(ctx)
+            log.trace("Invalid name provided by the user")
+            embed, msg = self.generate_invalid_embed(ctx)
```
RohanJnr, line 83: Instead of sending the context, you could only send the emojis (ctx.guild.emojis) as that is what is actually required.

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +19,71 @@ class EmojiCount(commands.Cog):
...
-            msg = ''
-            for key, value in emoji.items():
-                emoji_choice = random.choice(value)
-                emoji_info = f'There are **{len(value)}** emojis in the **{key}** category\n'
-                msg += f'<:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}'
-            embed.description = msg
-        return embed
+            for category_name, category_emojis in emoji.items():
+                emoji_choice = random.choice(category_emojis)
+                emoji_info = f'There are **{len(category_emojis)}** emojis in the **{category_name}** category\n'
```
RohanJnr, line 40: You could actually remove the `\n` here.
Here are a few screenshots:
**With \n**
image

**Without \n**
image

Experiment with it and see which one looks better. I'm leaning towards removing the `\n`, but do check with more emoji categories.

Anubhav1603: Actually I am using linepaginator that's y it's automatically add spaces in main pydis server we have many category that's y i am using linepaginator so it don't get errored out

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +19,71 @@ class EmojiCount(commands.Cog):
...
             timestamp=datetime.datetime.utcnow()
         )
+        msg = []
 
         if len(emoji) == 1:
-            for key, value in emoji.items():
-                embed.description = f"There are **{len(value)}** emojis in the **{key}** category"
-                embed.set_thumbnail(url=random.choice(value).url)
+            for category_name, category_emojis in emoji.items():
+                msg.append(f"There is **{len(category_emojis)}** emoji in the **{category_name}** category")
```
RohanJnr, line 34: this string need to dynamic based on singular or plural nature of `len(category_emojis)`
```suggestion
                msg.append(
                    f"There {'is' if len(category_emojis)==1 else 'are'} **{len(category_emojis)}** "
                    f"{'emoji' if len(category_emojis)==1 else 'emojis'} in the **{category_name}** category"
                )
```

bot/exts/evergreen/emoji_count.py
```diff
@@ -17,73 +19,71 @@ class EmojiCount(commands.Cog):
...
-            msg = ''
-            for key, value in emoji.items():
-                emoji_choice = random.choice(value)
-                emoji_info = f'There are **{len(value)}** emojis in the **{key}** category\n'
-                msg += f'<:{emoji_choice.name}:{emoji_choice.id}> {emoji_info}'
-            embed.description = msg
-        return embed
+            for category_name, category_emojis in emoji.items():
+                emoji_choice = random.choice(category_emojis)
+                emoji_info = f'There are **{len(category_emojis)}** emojis in the **{category_name}** category'
```
RohanJnr, line 40: here also,
```suggestion
                emoji_info = (
                    f"There {'is' if len(category_emojis) == 1 else 'are'} **{len(category_emojis)}** "
                    f"{'emoji' if len(category_emojis) == 1 else 'emojis'} in the **{category_name}** category"
                )
```

RohanJnr (changes requested):
Almost there.

bot/exts/evergreen/emoji_count.py
```diff
@@ -37,7 +37,10 @@ def embed_builder(emoji: dict) -> Tuple[discord.Embed, List[str]]:
         else:
             for category_name, category_emojis in emoji.items():
                 emoji_choice = random.choice(category_emojis)
-                emoji_info = f'There are **{len(category_emojis)}** emojis in the **{category_name}** category'
+                emoji_info = (
+                    f"There {'is' if len(category_emojis) == 1 else 'are'} **{len(category_emojis)}** "
+                    f"{'emoji' if len(category_emojis) == 1 else 'emojis'} in the **{category_name}** category"
+                )
```
thomaspet, line 46: Personally not a fan of this oneline. Much easier to read if you split it up into an if,else

```suggestion
if len(category_emojis) > 1:
    emoji_info = f"There are **{len(category_emojis)}** emojis in the **{category_name}** category"
else: # assuming there can't be 0 at this stage.
    emoji_info = "There is **1** emoji in the **{category_name}** category"
```

Don't commit the suggestion, definitely not indentation compatible.

RohanJnr (approved):
The code looks good and tested!

tagptroll1 (approved):
Sure, looks good to me

PR #481, opened 2020-10-04
Add space between dice in roll command

gustavwilliam:
## Description

The `.roll` command now has a single space between the outputted dice, when more than 1 roll is specified. The code behind the command is also a lot more refined.

## Reasoning

Previously, when given a number greater than 1, the `.roll` command sent the dice without space between them. This resulted in it looking quite strange.

## Screenshots

Current:
Screenshot 2020-10-04 at 22 45 35

Proposed:
Screenshot 2020-10-04 at 22 45 16

## Did you:

### DIFF
--- a/bot/exts/evergreen/fun.py
+++ b/bot/exts/evergreen/fun.py
@@ -7,7 +7,7 @@ from typing import Callable, Iterable, Tuple, Union
 
 from discord import Embed, Message
 from discord.ext import commands
-from discord.ext.commands import Bot, Cog, Context, MessageConverter, clean_content
+from discord.ext.commands import BadArgument, Bot, Cog, Context, MessageConverter, clean_content
 
 from bot import utils
 from bot.constants import Colours, Emojis
@@ -60,15 +60,17 @@ class Fun(Cog):
     @commands.command()
     async def roll(self, ctx: Context, num_rolls: int = 1) -> None:
         """Outputs a number of random dice emotes (up to 6)."""
-        output = ""
-        if num_rolls > 6:
-            num_rolls = 6
-        elif num_rolls < 1:
-            output = ":no_entry: You must roll at least once."
-        for _ in range(num_rolls):
-            dice = f"dice_{random.randint(1, 6)}"
-            output += getattr(Emojis, dice, '') + " "
-        await ctx.send(output.rstrip())
+        def _get_random_die() -> str:
+            """Generate a random die emoji, ready to be sent on Discord."""
+            die_name = f"dice_{random.randint(1, 6)}"
+            return getattr(Emojis, die_name)
+
+        # Only support between 1 and 6 rolls
+        if 1 <= num_rolls <= 6:
+            dice = " ".join(_get_random_die() for _ in range(num_rolls))
+            await ctx.send(dice)
+        else:
+            raise BadArgument("`!roll` only supports between 1 and 6 rolls.")
 
     @commands.command(name="uwu", aliases=("uwuwize", "uwuify",))
     async def uwu_command(self, ctx: Context, *, text: clean_content(fix_channel_mentions=True)) -> None:
--- a/bot/exts/evergreen/fun.py
+++ b/bot/exts/evergreen/fun.py
@@ -57,17 +57,17 @@ class Fun(Cog):
         with Path("bot/resources/evergreen/caesar_info.json").open("r", encoding="UTF-8") as f:
             self._caesar_cipher_embed = json.load(f)
 
+    @staticmethod
+    def _get_random_die() -> str:
+        """Generate a random die emoji, ready to be sent on Discord."""
+        die_name = f"dice_{random.randint(1, 6)}"
+        return getattr(Emojis, die_name)
+
     @commands.command()
     async def roll(self, ctx: Context, num_rolls: int = 1) -> None:
         """Outputs a number of random dice emotes (up to 6)."""
-        def _get_random_die() -> str:
-            """Generate a random die emoji, ready to be sent on Discord."""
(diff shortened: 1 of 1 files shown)


bot/exts/evergreen/fun.py
```diff
@@ -60,15 +60,13 @@ def __init__(self, bot: Bot) -> None:
...
-            output = ":no_entry: You must roll at least once."
-        for _ in range(num_rolls):
-            dice = f"dice_{random.randint(1, 6)}"
-            output += getattr(Emojis, dice, '')
-        await ctx.send(output)
+
+        dice = (f"dice_{random.randint(1, 6)}" for _ in range(num_rolls))
+        output = " ".join(getattr(Emojis, die, '') for die in dice)
+
+        await ctx.send(output or ":no_entry: You must roll at least once.")
```
lemonsaurus, line 69: I don't like this. You've turned 6 lines into 3, and introduced more complexity. I don't mind your code, but I prefered when it was 6 lines - it had better readability.

Our code style is to not have a newline after docstrings for functions and coroutines, by the way. I'm not sure why the linter let this through, I guess it's not set up with the same flake8-docstrings rules as we have on `bot`.

lemonsaurus (changes requested):

bot/exts/evergreen/fun.py
gustavwilliam: While I'm not fond of the general logic in the code — looping and creating new string objects every time — I still get your point about readability. I'll revert to the original solution, with the slight alterations needed to add the spaces.

Also, there is no newline after the docstring. I still had this part on top: 
```py
if num_rolls > 6:
             num_rolls = 6
```

lemonsaurus: oh okay, just misread that then. my bad.

gustavwilliam: I see what you're saying. Either way, I pushed a new commit that returned to the original logic and simply added the spaces. Sadly didn't see that you added that "for the record" part before finishing it.

mbaruh (changes requested):
If you add spaces the preferable approach for me would be to use `random.choices` + the `join` function on a genex and ditch the loop and concatenations. 
This will also allow you to avoid the strip at the end.

lemonsaurus (changes requested):
Yeah you're right, this isn't much better. Let's use a helper function. Maybe something like this?

```py
def _get_random_die() -> str:  # _ prefix since it's an internal helper.
    """Generate a random die emoji."""
    die_name = f"dice_{random.randint(1, 7)}"  # 1, 7, not 6. it's not inclusive.
    return getattr(Emojis, die_name)  # No point in the third arg, this isn't gonna fail.

# We only support num_roll between 1 and 6.
if 1 <= num_rolls <= 6:
    dice = " ".join(_get_random_die() for _ in num_rolls)
    ctx.send(dice.rstrip())
else:
    raise BadArgument("!roll only supports between 1 and 6 rolls.")
```

gustavwilliam:
Alright, I used your code @lemonsaurus with some minor alterations. This looks good to me; especially compared to the last version we tried out.

mbaruh (approved):
If you ran it and it worked, then it seems good to go.

bot/exts/evergreen/fun.py
```diff
@@ -60,15 +60,17 @@ def __init__(self, bot: Bot) -> None:
...
-        elif num_rolls < 1:
-            output = ":no_entry: You must roll at least once."
-        for _ in range(num_rolls):
-            dice = f"dice_{random.randint(1, 6)}"
-            output += getattr(Emojis, dice, '')
-        await ctx.send(output)
+        def _get_random_die() -> str:
+            """Generate a random die emoji, ready to be sent on Discord."""
+            die_name = f"dice_{random.randint(1, 6)}"
+            return getattr(Emojis, die_name)
```
lemonsaurus, line 66: We don't want to create this function every time the roll command is used, so move it up into the cog.

lemonsaurus (changes requested):

bot/exts/evergreen/fun.py
```diff
@@ -57,17 +57,17 @@ def __init__(self, bot: Bot) -> None:
...
         """Outputs a number of random dice emotes (up to 6)."""
-        def _get_random_die() -> str:
-            """Generate a random die emoji, ready to be sent on Discord."""
-            die_name = f"dice_{random.randint(1, 6)}"
-            return getattr(Emojis, die_name)
-
-        # Only support between 1 and 6 rolls
         if 1 <= num_rolls <= 6:
-            dice = " ".join(_get_random_die() for _ in range(num_rolls))
+            dice = " ".join(Fun._get_random_die() for _ in range(num_rolls))
```
lemonsaurus, line 70:
```suggestion
            dice = " ".join(self._get_random_die() for _ in range(num_rolls))
```
It's fine that `_get_random_die` is a `@staticmethod`, but we should still call it with `self`. Using the class name here violates the locality ideal, and also has a few other problems to consider:
- What happens if we rename this cog? Well, the method would break.
- What happens if you try to inherit from this cog, and then implement your own `_get_random_die` method? Then this would create surprising behavior in the inherited `roll` command, which would not use the new helper.

As a rule of thumb, always use `cls` or `self` when calling a method in your class, and don't stick your class name into `super()` either.

lemonsaurus (changes requested):

bot/exts/evergreen/fun.py
gustavwilliam: Thanks for the explanation. I've been wondering what's considered best practice in such a case.

lemonsaurus (approved):
Looks good.

PR #482, opened 2020-10-05
Startup channel check

### LINKED ISSUE #393: Verify channel constants on startup
## Description
The bot should check that there are no channels missing from the guild which are present in the bot constants.

## Reasoning
As discussed in #342 this will help prevent issues like #335 and has already been implemented for bot.

## Proposed Implementation
With the current implementation of the constants as simple classes (not sure why they're NamedTuples) without any annotations or being an object that holds more info about itself, we can check their contents through the `vars` function and handling any non private names

Another potentially cleaner solution would be changing the classes to something like enums which are iterable and gives us only the values we want without any filtering

Paging @neonsea  as they expressed interest in #342

WillDaSilva:
## Relevant Issues
Closes #393 

## Description
At startup, before the bot sends the "Connected!" message in the `dev_log` channel, a list of the available channels is obtained, and the channels defined in `constants.py` is checked against it. For every constant which refers to a channel which does not exist, an error message is logged which includes the channel name, and its ID.

## Reasoning
As mentioned on #393, changing the kind of class that `Channels` is would probably be more trouble than its worth. This approach adds little complexity, but should protect against the constants referring to channels which do not exist.

## Did you:


Xithrius:
@WillDaSilva if you do plan on continuing to work on this, please fix the linting errors.

Remember to [lint before you push](<https://soundcloud.com/lemonsaurusrex/lint-before-you-push>) via `pipenv run lint`!

Akarys42 (approved):
Lgtm, cheers!

ks129 (approved):

Issue #136, opened 2019-03-07
Change omdb api usage to use tmdb.


scragly:
The snakes extension was moved to this bot now, but it uses the omdb api. 
We should adjust whatever uses omdb to use tmdb, which is already used in a halloween feature.

empathic-geek:
@scragly -  happy to attempt this, this will be my first issue. So please forgive if its slow...

RohanJnr:
ill take this

RohanJnr:
im sorry i will not be able to do this as my broadband service provider blocks tmdb in India for some reason.
here are a few links which say that (and yes obvi i cannot open tmdb website so i cannot work on this :/)
https://twitter.com/themoviedb/status/1085566644573749248
https://www.quora.com/Why-has-ACT-Fibernet-blocked-tmdb-themoviedb-in-India

adityaprakash-bobby:
I would like to take this issue if no one is working on this.

WillDaSilva:
If @adityaprakash-bobby isn't opposed, I'd like to implement this.

PR #484, opened 2020-10-05
Replace OMDB with TMDB

WillDaSilva:
## Relevant Issues
Closes: #136

## Description
OMDB was only used for the `snake movie` command. I replaced its usage in that command, and while doing so streamlined the construction of the embed.

## Did you:


bot/exts/evergreen/snakes/_snakes_cog.py
```diff
@@ -739,71 +740,62 @@ async def hatch_command(self, ctx: Context) -> None:
...
+            data = await response.json()
 
-            if key == "Ratings":  # [{'Source': 'Internet Movie Database', 'Value': '7.6/10'}]
-                rating = random.choice(value)
+        embed = Embed(title=data["title"], color=SNAKE_COLOR)
 
-                if rating["Source"] != "Internet Movie Database":
-                    embed.add_field(name=f"Rating: {rating['Source']}", value=rating["Value"])
+        if data["poster_path"] is not None:
+            embed.set_image(url=f"https://images.tmdb.org/t/p/original{data['poster_path']}?api_key={Tokens.tmdb}")
```
MarkKoz, line 776: Wouldn't this leak the token?

bot/exts/evergreen/snakes/_snakes_cog.py
```diff
@@ -739,71 +740,62 @@ async def hatch_command(self, ctx: Context) -> None:
...
-        Gets a random snake-related movie from OMDB.
+        Gets a random snake-related movie from TMDB.
 
         Written by Samuel.
         Modified by gdude.
+        Modified by Will Da Silva.
         """
-        url = "http://www.omdbapi.com/"
-        page = random.randint(1, 27)
+        page = random.randint(1, 16)
```
MarkKoz, line 749: What happens if the chosen page is greater than the last page of the search results?

MarkKoz (changes requested):
You could narrow this down to a single request if you remove the budget and revenue fields. the search response has all of the other data, with the exception of genre names. It only gives genre ID, but what you could do is retrieve all genres once and cache them.

It's just a thought. I suppose 2 requests are fine for something low-frequency like this.

bot/exts/evergreen/snakes/_snakes_cog.py
WillDaSilva: Good catch. It would seem the token isn't necessary for this request, so I'll just remove it.

bot/exts/evergreen/snakes/_snakes_cog.py
WillDaSilva: Previously the max page was hard-coded, so I figured I'd just go that route for now, and improve it if it's deemed necessary. There are 16 pages in the search results for "snake". In the future, as more movies are added to TMDB, there will eventually be more pages, but that won't break anything here, and I do not expect there will ever be fewer pages than the amount there are now.

This could be fixed by setting the upper range of the random number generation to the `total_pages` field returned by the search, but this would require another request. Alternatively the first request could be limited to 16 pages (or some other low hard-coded number of pages), and then `total_pages` would be used afterwards.

WillDaSilva: Or we could go for some more elaborate pre-fetching/caching.

MarkKoz: > Alternatively the first request could be limited to 16 pages (or some other low hard-coded number of pages), and then total_pages would be used afterwards.

This sounds good. The initial number should be slightly lower than the threshold to give some leeway for films being removed.

MarkKoz (changes requested):
This mostly works, but sometimes there is a 400 error because the embed has a field with an empty value:

```
discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body
In embed.fields.0.value: This field is required
```

I'm also concerned about the terms of service for the TMDB API, which states

> You shall use the TMDb logo to identify your use of the TMDb APIs.
> 
> You shall place the following notice prominently on your application: "This product uses the TMDb API but is not endorsed or certified by TMDb."
> 
> Any use of the TMDb logo in your application shall be less prominent than the logo or mark that primarily describes the application and your use of the TMDb logo shall not imply any endorsement by TMDb.

It's not yet clear what is adequate to comply with this, but it's something we'll need to figure out before merging this. May need to use the exact phrasing in the footer and include the logo as the embed author or something.

WillDaSilva:
It's worth noting that we do not comply with the terms of service in the other places where TMDB is used. We may want to open issues to correct that.

MarkKoz (approved):
I imagined the logo being put as the avatar icon URL, but it's actually to hard to make out the logo when it's that small

bild

I agree that other uses of the API should be adjusted to comply with the ToS.

WillDaSilva:
I opened https://github.com/python-discord/seasonalbot/issues/505

ks129 (approved):
Code OK, tested locally.

ks129:
@WillDaSilva This PR have some conflicts, can you please solve them?

PR #492, opened 2020-10-07
Fix snake video command

### LINKED ISSUE #450: .snake video command not responding
## Description

When using the `.snake video` command, no response is given by the bot. This is the error message received from a local test bot:
```
09/21/20 23:40:12 - bot.exts.evergreen.snakes._snakes_cog ERROR: Unhandled tag command error: Command raised an exception: TypeError: Invalid variable type: value should be str or int, got None of type <class 'NoneType'> (Invalid variable type: value should be str or int, got None of type <class 'NoneType'>)
```
This is most likely the result of an API change or similar, resulting in a NoneType return value.

## Steps to Reproduce

Send `.emoji video` in the #seasonalbot-commands channel.

## Expected Behaviour

A return message, containing a link to a YouTube video about snakes (as describes in the help of the command).

## Actual Behaviour

No message was returned from the bot. The earlier described error could be found in the log of the bot server.

## Known Impacted Platforms

- [x] Web
- [x] Desktop
- [x] Android App
- [x] iOS App

### LINKED ISSUE #490: Unhandled tag command error: Command raised an exception: KeyError: 'items' ('items')
Sentry Issue: [SEASONALBOT-11](https://sentry.io/organizations/python-discord/issues/1830840781/?referrer=github_integration)

```
Unhandled tag command error: Command raised an exception: KeyError: 'items' ('items')
```

ks129:
## Relevant Issues

Closes #490

## Description

After testing, I found that `urllib.parse.quote` is parsing spaces to `%2520%` that YouTube API don't support. Changing this to `urllib.parse.quote_plus` make this parsing spaces to `+` what YouTube API can read and then return again proper response with `items` key defined.

## Did you:

### DIFF
--- a/bot/exts/evergreen/snakes/_snakes_cog.py
+++ b/bot/exts/evergreen/snakes/_snakes_cog.py
@@ -1089,7 +1089,7 @@ class Snakes(Cog):
             }
         )
         response = await response.json()
-        data = response['items']
+        data = response.get("items", [])
 
         # Send the user a video
         if len(data) > 0:
--- a/bot/exts/evergreen/snakes/_snakes_cog.py
+++ b/bot/exts/evergreen/snakes/_snakes_cog.py
@@ -1083,7 +1083,7 @@ class Snakes(Cog):
             url,
             params={
                 "part": "snippet",
-                "q": urllib.parse.quote(query),
+                "q": urllib.parse.quote_plus(query),
                 "type": "video",
                 "key": Tokens.youtube
             }


gustavwilliam (approved):
I got an API key and tried this out. Works like a charm!

kwzrd (commented):
I'm not sure whether this solves the problem entirely. When I try the invoking the command with a valid token and just `quote`, it logs a warning `1104`, rather than failing on a `KeyError`. Can you confirm that you're getting the same behaviour?

ks129:
@kwzrd What? `L1104` is `@snakes_group.command(name='zen')`.

kwzrd:
Ah, sorry I had some extra code above. I meant the warning on line `1102` that starts with "YouTube API error".

kwzrd:
It logs an error, but doesn't fail on the `KeyError`. Which makes me think that the problem is two-fold: there may be a problem with the API key used in production, and there is a problem with the spaces (fixed by this PR).

kwzrd:
The response I get using a valid API token but `quote` instead of `quote_plus` contains an `'items'` key with an empty list as the value.

ks129:
Is this warning like:
```
YouTube API error. Full response looks like {'kind': 'youtube#searchListResponse', 'etag': 'pG-yJlSqu9dpyZEZ1HjJUOB2Us8', 'regionCode': 'EE', 'pageInfo': {'totalResults': 0, 'resultsPerPage': 5}, 'items': []}
```
because when this looks similar to this, then problem is still it, as YouTube API respond such response when this can't find videos and reason why this can't find videos is that this can't parse `%2520%` to space.

kwzrd:
Yes, but the problem we're encountering in production (see the two issues linked in the PR description) is different. It's a `KeyError` on line `1092`. The fix here is good, but won't solve the problem in the linked issues.

kwzrd (approved):
Thanks, looks great. @SebastiaanZ helped us resolve the API token issue, while this PR makes the cog more resilient to future breakage as we'll now log the response on error instead of failing completely. The use of `quote_plus` makes multi-word queries resolve correctly. Great work.

PR #499, opened 2020-10-11
Trivia: only fuzzy match messages sent in game channel

### LINKED ISSUE #351: [Trivia] Use fuzzywuzzy ONLY on the messages from the channel the game was activated in
## Description
The current cog for the trivia game does a fuzzywuzzy check for every message in the entire server while it's being played.

We could check instead if the channel the message was sent from is the right one, then do the fuzzywuzzy.

There are other ways the cog could be enhanced in a larger scale, I'll try to open an issue later or suggest a refactoring.

## Reasoning
The way it currently works looks like a bit of a waste. Considering the main thing our bots do is processing messages in an ever-growing server, I think it'd be better to avoid checking messages that aren't needed (unless it makes the code significantly uglier.)

## Proposed Implementation
I would probably just replace
```python
            def check(m: discord.Message) -> bool:
                ratio = fuzz.ratio(answer.lower(), m.content.lower())
                return ratio > 85 and m.channel == ctx.channel
```
with
```python
            def check(m: discord.Message) -> bool:
                return (
                   m.channel == ctx.channel and
                   fuzz.ratio(answer.lower(), m.content.lower()) > 85
                )
```
Or similar.

Anubhav1603:
"Closes #351 "
this is a simple fix in check func

### DIFF
--- a/bot/exts/evergreen/trivia_quiz.py
+++ b/bot/exts/evergreen/trivia_quiz.py
@@ -122,7 +122,8 @@ class TriviaQuiz(commands.Cog):
             # A function to check whether user input is the correct answer(close to the right answer)
             def check(m: discord.Message) -> bool:
                 return (
-                    m.channel == ctx.channel and fuzz.ratio(answer.lower(), m.content.lower()) > 85
+                    m.channel == ctx.channel
+                    and fuzz.ratio(answer.lower(), m.content.lower()) > 85
                 )
 
             try:
--- a/bot/exts/evergreen/trivia_quiz.py
+++ b/bot/exts/evergreen/trivia_quiz.py
@@ -121,8 +121,9 @@ class TriviaQuiz(commands.Cog):
 
             # A function to check whether user input is the correct answer(close to the right answer)
             def check(m: discord.Message) -> bool:
-                ratio = fuzz.ratio(answer.lower(), m.content.lower())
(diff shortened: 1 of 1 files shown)


kwzrd (changes requested):
I prefer the line break as proposed in the linked issue. It makes the expression much easier to read.

Also, please put a little bit more effort into the commit message(s) and the PR title and description. The fact that you've "changed return" is the one thing we can see from the commit diff. An explanation of why would be appreciated.

thomaspet:
I agree with kwzrd, either break line it or write it out, like

```py
def check(m: discord.Message) -> bool:
    if m.channel == ctx.channel:
        return fuzz.ratio(answer.lower(), m.content.lower()) > 85
    return False # To avoid implicit "return falsey"
```

thomaspet (approved):
For future reference, branch names should also be a bit descriptive. I personally would have named this `fix/tagp/trivia-fuzzy-check` fix for to indicate its a small fix, not a feature, tagp are my initials, and a small descriptive last part.

Anubhav1603:
> For future reference, branch names should also be a bit descriptive. I personally would have named this `fix/tagp/trivia-fuzzy-check` fix for to indicate its a small fix, not a feature, tagp are my initials, and a small descriptive last part.

ty i will take care from nxt time :)

kwzrd (approved):
Thanks!

PR #508, opened 2020-10-21
fix keyerror for hackstats

hedythedev:
## Description

hot fix for the `hackstats` command

## Reasoning

it was raising key error:
E3FA0F7C-B992-4E5A-8C49-D5B9DC984E8A

and it seems that `item` was wrongly passed to `_is_accepted` and it should be `itemdict`

## Screenshots

now it works for users that need checking for acceptance
140C34DB-3269-466B-A40F-4B997461A08B


kwzrd (approved):
Thanks for the quick response.

PR #514, opened 2020-11-14
Migrate to GitHub Actions and GHCR

SebastiaanZ:
We are currently migrating our backend to a Kubernetes cluster. As this requires changes to our CI anyway, this is a good time to also migrate our CI Pipeline to GitHub Actions and start using the GitHub Container Registry to push our builds to.

Changes compared to the old situation:

- flake8 linting errors will now show up as GH Actions annotations
- containers are tagged with both "latest" and a short commit SHA

Note: This version of the workflow still pushes to DockerHub **in addition to** GHCR. This is to make the transition between the two deployment systems as seamless as possible.


.github/workflows/lint-build.yaml
```diff
@@ -0,0 +1,126 @@
...
+    name: Lint using pre-commit & flake8
+    runs-on: ubuntu-latest
+    env:
+      PIP_NO_CACHE_DIR: false
+      PIP_USER: 1
+      PIPENV_HIDE_EMOJIS: 1
+      PIPENV_IGNORE_VIRTUALENVS: 1
+      PIPENV_NOSPIN: 1
+      PYTHONUSERBASE: ${{ github.workspace }}/.cache/py-user-base
+      PRE_COMMIT_HOME: ${{ github.workspace }}/.cache/pre-commit-cache
```
lemonsaurus, line 21: Can we document what these actually do? I have no idea what half of them are for.

.github/workflows/lint-build.yaml
```diff
@@ -0,0 +1,126 @@
...
+          python-version: '3.8'
+
+      - name: Python Dependency Caching
+        uses: actions/cache@v2
+        id: python_cache
+        with:
+          path: ${{ env.PYTHONUSERBASE }}
+          key: "python-0-${{ runner.os }}-${{ env.PYTHONUSERBASE }}-\
+          ${{ steps.python.outputs.python-version }}-\
+          ${{ hashFiles('./Pipfile', './Pipfile.lock') }}"
```
lemonsaurus, line 47: How this works is also really opaque to me.

.github/workflows/lint-build.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+      - name: Install dependencies using pipenv
+        if: steps.python_cache.outputs.cache-hit != 'true'
```
lemonsaurus, line 50: This means that we shouldn't install if we can use cached ones? Might also be worth doing a block comment for.

.github/workflows/lint-build.yaml
```diff
@@ -0,0 +1,126 @@
...
+          pip install pipenv
+          pipenv install --dev --deploy --system
+
+      - name: Pre-commit Environment Caching
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.PRE_COMMIT_HOME }}
+          key: "precommit-0-${{ runner.os }}-${{ env.PRE_COMMIT_HOME }}-\
+          ${{ steps.python.outputs.python-version }}-\
+          ${{ hashFiles('./.pre-commit-config.yaml') }}"
```
lemonsaurus, line 61: Absolutely no idea what this does.

.github/workflows/lint-build.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+      # We will not run `flake8` here, as we will use a separate flake8
+      # action. As pre-commit does not support user installs, and we don't
+      # really need it, we set PIP_USER=0.
+      - name: Run pre-commit hooks
+        run: export PIP_USER=0; SKIP=flake8 pre-commit run --all-files
```
lemonsaurus, line 67: Can we go into some more detail about how PIP_USER=0 works? I don't understand it.

.github/workflows/lint-build.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+      # This step requires `pull_request_target` due to the use of annotations
```
lemonsaurus, line 69: I don't get it. Why does it require that? Why do annotations require `pull_request_target`?

.github/workflows/lint-build.yaml
```diff
@@ -0,0 +1,126 @@
...
+    if: github.event_name != 'pull_request_target' && github.ref == 'refs/heads/master'
+    runs-on: ubuntu-latest
+
+    steps:
+      # Create a commit SHA-based tag for the container repositories
+      - name: Create SHA Container Tag
+        id: sha_tag
+        run: |
+          tag=$(cut -c 1-7 <<< $GITHUB_SHA)
+          echo "::set-output name=tag::$tag"
```
lemonsaurus, line 89: If I understand this correctly, it will create a sort of local variable that we can later refer to as `${{ steps.sha_tag.outputs.tag }}`? It might be nice to make it a little clearer what the effect of this is, because it's a bit arcane.

.github/workflows/lint-build.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          persist-credentials: false
```
lemonsaurus, line 94: Above, we're not persisting credentials because it may run from a fork, but this job will never be run from a fork. So what's the rationale here?

lemonsaurus (changes requested):
Can we just add a bit more documentation so stupid people like myself can comprehend this?

MarkKoz (approved):

.github/workflows/lint-build.yaml
SebastiaanZ: I'll add documentation.

It's the cache key: It takes in properties of the run (like python version, current state of dependencies). If a next run has the exact same key, it will restore cache. However, if something changed (python version, dependencies), we do not restore the cache and install it instead.

.github/workflows/lint-build.yaml
SebastiaanZ: That's exactly what it says.

.github/workflows/lint-build.yaml
SebastiaanZ: It says "don't do a user install", like the comment says. Nothing more.

.github/workflows/lint-build.yaml
SebastiaanZ: Because it needs write access to the repository. I'll add a comment about that.

.github/workflows/lint-build.yaml
SebastiaanZ: I'll remove it. There's not really a need for us to persist the credentials here, but I guess it doesn't matter either.

.github/workflows/lint-build.yaml
SebastiaanZ: sure

.github/workflows/lint-build.yaml
MarkKoz: Maybe "user installs" is unclear? It means that it should install to the global site for Python packages rather than a user-specific site. And "site" is Python's terminology for location/directory; it does not website.

.github/workflows/lint-build.yaml
lemonsaurus: ahh, that makes sense.

.github/workflows/lint-build.yaml
lemonsaurus: yes, I was asking.

lemonsaurus (approved):

PR #518, opened 2020-11-15
Add gurkancount command

gustavwilliam:
## Description

This command loops through the display names of all members and checks which ones contain "gurkan".

### Important note:
This enables member intents, which means that it'll have to be enabled on [Discord Developers Portal](https://discord.com/developers/applications), if it isn't already. It also means that we'll be tracking member intents.

This is needed, since the display names of guild members will be checked.

## Screenshots

Here is the new command:
Screenshot 2020-11-15 at 04 23 41

## Did you:

### DIFF
--- a/bot/exts/evergreen/gurkan.py
+++ b/bot/exts/evergreen/gurkan.py
@@ -13,8 +13,9 @@ class Gurkan(commands.Cog):
     @commands.command(name="gurkancount", aliases=("gc", "gurkcount"))
     async def gurkancount(self, ctx: commands.Context) -> None:
         """Get the amount of members with 'gurkan' in their names."""
-        gurkans = [1 for member in ctx.guild.members if "gurkan" in member.display_name.lower()]
-        gurkan_count = sum(gurkans)
+        gurkan_count = sum(
+            1 for member in ctx.guild.members if "gurkan" in member.display_name.lower()
+        )
 
         embed = discord.Embed(
             title="Gurkancount",
--- a/bot/exts/evergreen/gurkan.py
+++ b/bot/exts/evergreen/gurkan.py
@@ -1,0 +1,29 @@
+import discord
+from discord.ext import commands
+
+from bot.constants import Colours
+
+
+class Gurkan(commands.Cog):
+    """Commands related to Gurkan and the Gurkult."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+
+    @commands.command(name="gurkancount", aliases=("gc", "gurkcount"))
+    async def gurkancount(self, ctx: commands.Context) -> None:
+        """Get the amount of members with 'gurkan' in their names."""
+        gurkans = [1 for member in ctx.guild.members if "gurkan" in member.display_name.lower()]
+        gurkan_count = sum(gurkans)
+
+        embed = discord.Embed(
+            title="Gurkancount",
+            color=Colours.bright_green,
+            description=f"There are a total of `{gurkan_count}` gurkans in the server!")
+
+        await ctx.send(embed=embed)
+
+
+def setup(bot: commands.Bot) -> None:
+    """Load the Bookmark cog."""
+    bot.add_cog(Gurkan(bot))
--- a/bot/exts/evergreen/gurkan.py
+++ b/bot/exts/evergreen/gurkan.py
@@ -11,7 +11,7 @@ class Gurkan(commands.Cog):
         self.bot = bot
 
     @commands.command(name="gurkancount", aliases=("gc", "gurkcount"))
-    async def gurkancount(self, ctx: commands.Context) -> None:
+    async def gurkan_count(self, ctx: commands.Context) -> None:
         """Get the amount of members with 'gurkan' in their names."""
         gurkan_count = sum(
             1 for member in ctx.guild.members if "gurkan" in member.display_name.lower()
@@ -20,11 +20,12 @@ class Gurkan(commands.Cog):
         embed = discord.Embed(
             title="Gurkancount",
             color=Colours.bright_green,
-            description=f"There are a total of `{gurkan_count}` gurkans in the server!")
+            description=f"There are a total of `{gurkan_count}` gurkans in the server!"
+        )
 
         await ctx.send(embed=embed)
 
 
 def setup(bot: commands.Bot) -> None:
-    """Load the Bookmark cog."""
+    """Load the Gurkan cog."""
     bot.add_cog(Gurkan(bot))
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -211,6 +211,7 @@ _intents.integrations = False
 _intents.invites = False
 _intents.typing = False
 _intents.webhooks = False
+_intents.members = True
 
 bot = SeasonalBot(
     command_prefix=Client.prefix,


bot/exts/evergreen/gurkan.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Gurkan(commands.Cog):
+    """Commands related to Gurkan and the Gurkult."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+
+    @commands.command(name="gurkancount", aliases=("gc", "gurkcount"))
+    async def gurkancount(self, ctx: commands.Context) -> None:
```
fisher60, line 14: This function name could be clearer if it was using `snake_case` i.e `gurkan_count`

bot/exts/evergreen/gurkan.py
```diff
@@ -0,0 +1,29 @@
...
+        embed = discord.Embed(
+            title="Gurkancount",
+            color=Colours.bright_green,
+            description=f"There are a total of `{gurkan_count}` gurkans in the server!")
+
+        await ctx.send(embed=embed)
+
+
+def setup(bot: commands.Bot) -> None:
+    """Load the Bookmark cog."""
```
fisher60, line 28: Forgot to change the docstring to the `Gurkan Cog` here.

fisher60 (changes requested):
There is not much to review here. Mostly style preferences. Only really need to change the docstring.

The code ran perfectly fine, the only real concern would be the performance of looping through so many members.

bot/exts/evergreen/gurkan.py
```diff
@@ -0,0 +1,29 @@
...
+    @commands.command(name="gurkancount", aliases=("gc", "gurkcount"))
+    async def gurkancount(self, ctx: commands.Context) -> None:
+        """Get the amount of members with 'gurkan' in their names."""
+        gurkans = [1 for member in ctx.guild.members if "gurkan" in member.display_name.lower()]
+        gurkan_count = sum(gurkans)
+
+        embed = discord.Embed(
+            title="Gurkancount",
+            color=Colours.bright_green,
+            description=f"There are a total of `{gurkan_count}` gurkans in the server!")
```
Den4200, line 22: The closing parenthesis should be on the next line.

bot/exts/evergreen/gurkan.py
```diff
@@ -0,0 +1,29 @@
...
+    """Commands related to Gurkan and the Gurkult."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+
+    @commands.command(name="gurkancount", aliases=("gc", "gurkcount"))
+    async def gurkancount(self, ctx: commands.Context) -> None:
+        """Get the amount of members with 'gurkan' in their names."""
+        gurkans = [1 for member in ctx.guild.members if "gurkan" in member.display_name.lower()]
+        gurkan_count = sum(gurkans)
```
Den4200, line 17: Maybe something like this instead? I'm not sure how I feel about having two variables for this.
```suggestion
        gurkan_count = sum(
            1 for member in ctx.guild.members if "gurkan" in member.display_name.lower()
        )
```

Den4200 (changes requested):
There are all purely stylistic changes!

bot/exts/evergreen/gurkan.py
gustavwilliam: I'm not sure about this one, since it's a command. If it was a usual method, it would be using snake_case, but this is a command.

Den4200:
> There is not much to review here. Mostly style preferences. Only really need to change the docstring.
> 
> The code ran perfectly fine, the only real concern would be the performance of looping through so many members.

@fisher60 It takes very, very little resources loop through the members, so no worries here. It's only about 100,000 iterations for our server size.

bot/exts/evergreen/gurkan.py
Den4200: What's wrong with having the name as `snake_case`?

bot/exts/evergreen/gurkan.py
gustavwilliam: Looks much better. I didn't quite like having all of that on a single line nor creating two variables for it, so this solution looks better in some sense.

The only real reason that having two variables would be worth considering is if we would like to add some feature to do additional filtering in the future. That can be added when it becomes relevant, though, so this change looks good to me.

bot/exts/evergreen/gurkan.py
gustavwilliam: We definitely could. I didn't think that was the convention for commands, though. Let's do it.

Den4200: The command name is `gurkancount`, as you set it in the decorator, but the method name should be `snake_case` instead.

Den4200 (approved):
Looks good to me.

lemonsaurus (changes requested):
Let's make it more generic. I'd like to know how many users fit into _any_ arbitrary nickname feature.  We used to have `.lemoncount`, and I think there are still dozens. Let's have a feature that will count number of users with nicknames that contain any substring?

gustavwilliam:
@lemonsaurus We could do that. The question is if we really want users to be able to search for any terms in display names or only a set of terms. What do you think about this?

lemonsaurus:
what would be really cool is if you could hijack any command that ended with _count_ , so that `.gurkancount` and `.lemoncount` and `.vescount` would all just work.

gustavwilliam:
That would be really interesting. Sadly, it would interfere with the emojicount command. If we were to ever do that (which would be super cool), we'd have to turn emoji count into a command + subcommand or something similar.

Either way, if we keep in within the scope of this PR, I wouldn't mind adding an optional argument with the string to search for. I think keeping the command name `gurkancount` would also be slightly funny and work well.

Then again, do we want people to search for *any* term? I don't see any clear issues with that. We have great moderation, so inappropriate terms would be handlers just like with anything.

Den4200:
> what would be really cool is if you could hijack any command that ended with _count_ , so that `.gurkancount` and `.lemoncount` and `.vescount` would all just work.

This is a cool idea, I like it.


> That would be really interesting. Sadly, it would interfere with the emojicount command. If we were to ever do that (which would be super cool), we'd have to turn emoji count into a command + subcommand or something similar.

This shouldn't be an issue, since the logic would have to be implemented in the error handler.


> Then again, do we want people to search for *any* term? I don't see any clear issues with that. We have great moderation, so inappropriate terms would be handlers just like with anything.

I don't see it as an issue either.

gustavwilliam:
I think let's just start with implementing it where you have a second optional argument for specifying the name. If we want to expand it to work with every count, let's do that later.

SebastiaanZ:
One thing to consider is that SeasonalBot currently does not have the "members intent", which means that it also does not receive those events from Discord. Enabling the intent does increase the resource usage of Seasonal Bot (although the "presences intent" is probably way more heavy). I think I'm just a bit confused as to why we'd want this count command and if it's worth the additional load it will have on our back-end.

gustavwilliam:
@SebastiaanZ yes. That's an important question to ask. I don't know exactly how much that would cost to have, but is it something that we have resources for? While we shouldn't add commands just because we can, I also think it would be a fun command to have — which is an important part for SeasonalBot.

As for the reasoning behind having a command like this, I think there's an interest in counting the amount of members with "gurkan" in their display names. You have probably noticed that the count is quite significant. As @lemonsaurus mentions, it would also be fun to check other names, like "lemon" and "ves."

TL;DR: It's a fun command that I think many would appreciate.

lemonsaurus:
hmm, ves makes a good point about user intents. actually the resource load difference would probably be a bit significant.

this seemed like a better idea while I was drunk.

gustavwilliam:
Is that something we don't have resources for?

kwzrd:
We do have resources for it, the question is more whether we want to use them this way. Before intents rolled out, every bot was subscribed by default. When I was working on #478, core dev consensus seemed to be that we'd rather not have the intent if feasible. I'm impartial, but I don't know if this is a feature that would justify adding the intent, if we don't otherwise want it. If it's well implemented, generic and would be a useful utility to e.g. mods, we can also consider adding it to the main bot, which does subsribe to the intent.

lemonsaurus:
yeah, I'm afraid I think member intents is too high a price to pay for this feature. Sorry, gurk.

PR #520, opened 2020-11-16
Fix flake8 check in CI and separate Lint and Build jobs into separate workflows

SebastiaanZ:
This PR makes two changes to our CI:

#### 1. Instead of using a predefined action, we now use `flake8` directly to generate annotations for a CI check

Unfortunately, the flake8 action we were using from the marketplace required us to use the `pull_request_target` event, which runs in the context of the target repository to protect secrets. However, this also meant that flake8 would run on files already merged into our master branch, not the actual changes made in teh PR! That's obviously pretty useless as a guard against merging linting errors into our repository.

This change sidesteps the issue by removing the marketplace action and replacing it by a direct `flake8` run command. To make sure error output ends up as a GitHub Actions Annotation, we ask `flake8` to format its error messages using the correct GitHub Workflow Command format:

```
::error file={filename},line={line},col={col}::{message}
```

Whenever something is printed to stdout/stderr in that format, GitHub Actions will automatically interpret it as the ["register error message" workflow command](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-error-message). Since this doesn't require a GitHub Token with "write" permissions, we
can now switch back to the safer `pull_request` event that only gets a read-only GitHub Token.

This change will result in these neat annotations when a PR contains linting errors:

**Note:** I had to fix some linting errors that were introduced when we ripped out the persistence feature. See the second commit.

#### 2. The `lint` and `build` jobs now have their own workflow file.

I've separated the lint and build jobs into two separate workflows:

- Lint:  .github/workflows/lint.yaml
- Build: .github/workflows/build.yaml

The main difference is that the Build workflow will be triggered if the Lint workflow completes, but only if the current branch is "master" (i.e., it will not run for pull requests). The build job will check if the Lint run was successful and if it were, it actually builds the container, pushes it the GitHub Container Repository, and triggers the deployment to our kubernetes cluster.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,64 @@
+name: Build
+
+on:
+  workflow_run:
+    workflows: ["Lint"]
+    branches:
+      - master
+    types:
+      - completed
+
+jobs:
+  build:
+    if: github.event.workflow_run.conclusion == 'success'
+    name: Build, Push, & Deploy Container
+    runs-on: ubuntu-latest
+
+    steps:
+      # Create a commit SHA-based tag for the container repositories
+      - name: Create SHA Container Tag
+        id: sha_tag
+        run: |
+          tag=$(cut -c 1-7 <<< $GITHUB_SHA)
+          echo "::set-output name=tag::$tag"
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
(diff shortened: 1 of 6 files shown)


MarkKoz (approved):

lemonsaurus (approved):
image

PR #532, opened 2020-11-27
Add check for core dev approvals on pull requests

SebastiaanZ:
This commit adds a workflow that checks if a pull request has received at least one approval from a core developer. If not, the check will fail and block us from merging the Pull Request.

The reason we're adding this check is because we're moving away from making the entire Core Developers team a "code owner" of every line in every repository. This created a lot of "notification" spam, which drowned out the actually relevant notifications.

As we still want to make sure that there's at least one approving core developer, we now solve this using a workflow that checks for core dev approvals.

**Note:** I've annotated the workflow file itself extensively with comments. I've opted not to repeat that information here in the body of the pull request.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,10 +1,7 @@
-# Request Dennis for any PR
-* @Den4200
-
 # CI & Docker
-.github/workflows/**                    @Akarys42 @SebastiaanZ
-Dockerfile                              @Akarys42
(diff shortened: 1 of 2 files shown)


Akarys42 (approved):
Looks good to me sir!

Den4200 (approved):
Beautifully documented and well written, nice work! Everything looks correct to me, but I have not tested it.

PR #535, opened 2020-12-01
Refactor Advent of Code background tasks to account for deseasonification

SebastiaanZ:
The Advent of Code extension runs two background tasks during the event:

- `countdown_day` to announce the release of the next puzzle
- `countdown_status` to update the bot's status with the time left until the next puzzle is released

However, both tasks were written with Season Manager in mind, meaning that they rely on the extension only being loaded in December. In addition, some changes in Discord's API flow mean that the `countdown_status` task has a high chance of failing during its first cycle. It's unclear to me whether or not the `countdown_day` background task is currently running as expected.

## Main changes

I've made several changes to background tasks to adapt them to the absence of the Season Manager. In concrete terms, instead of exiting immediately when the bot starts up outside of the boundaries of the Advent of Code, the tasks will now **hibernate to just before the currently configured event starts**, waking up 1 hour (`countdown_day`) and 2 hours (`countdown_status`) before the event starts to prepare for the first puzzle release.

After the currently configured event (determined by the current value of the existing environment variable `AOC_YEAR`) has finished, both tasks will terminate. Once we update the configuration of the bot for next year's Advent of Code, by updating the environment variables, the tasks will start hibernating when the bot starts again. This means that the tasks don't hibernate unnecessary for most of the year.

To facilitate the "hibernate until the Advent of Code starts" behavior, I've implemented a helper function to do just that.

## Additional changes

- Our notification task, `countdown_day`, had the tendency to wake up early, which is why we padded its sleep value with one second. The reason for this tendency was because we were using `datetime.timedelta.seconds`, which truncates the fractional part of the duration. By using `datetime.timedelta.total_seconds()`, which does return a float that includes the fractional seconds, we should now wake up really close to the actual target time. I still added a padding of 0.1 seconds, just to be sure, although I don't think it's strictly necessary.

- The notification task, `countdown_day`, currently fails if it starts up really close to the puzzle announcement time. This is because it does not wait for the guild cache to be fully loaded. To avoid that, I've added a `wait_until_guild_available` call.

- The status task, `countdown_status`, immediately tries to change the bot's Playing status after starting up. This errors out as the [websocket instance has been fully initialized by `discord.py` yet](https://sentry.io/organizations/python-discord/issues/2089169634/?referrer=slack) (sentry link, core dev access only). Additionally, I've read mention of the API rejecting such a request (in dramatic fashion) if you make it too early. The `wait_until_guild_available` lets the task wait until the websocket has been fully initialized and the API connection is mature.

  The error the task currently gets:
```python
12/13/20 09:44:57 - bot.exts.christmas.advent_of_code._helpers ERROR: Background task `AoC Status Countdown` failed:
  File "/bot/bot/exts/christmas/advent_of_code/_cog.py", line 52, in countdown_status
    await bot.change_presence(activity=discord.Game(playing))
  File "/usr/local/lib/python3.8/site-packages/discord/client.py", line 1049, in change_presence
    await self.ws.change_presence(activity=activity, status=status, afk=afk)
AttributeError: 'NoneType' object has no attribute 'change_presence'
```

- I've moved the non-cog functions to the `_helpers.py` module within the `advent_of_code` package.

## Did you:


MarkKoz (approved):
I've not tested this but I trust that you have. The changes are more minimal than the diff implies so I feel confident to approve.

SebastiaanZ:
> The changes are more minimal than the diff implies so I feel confident to approve.

Yeah, I decided to move it to the helpers file, which caused the large diff. The main difference now is that there's logic for handling out-of-season start-ups of the tasks that are still able to wake up when the AoC draws near without having to restart the bot after the event has started.

bot/exts/christmas/advent_of_code/_helpers.py
```diff
@@ -393,8 +398,8 @@ def is_in_advent() -> bool:
     return datetime.datetime.now(EST).day in range(1, 25) and datetime.datetime.now(EST).month == 12
 
 
-def time_left_to_aoc_midnight() -> Tuple[datetime.datetime, datetime.timedelta]:
-    """Calculates the amount of time left until midnight in UTC-5 (Advent of Code maintainer timezone)."""
+def time_left_to_est_midnight() -> Tuple[datetime.datetime, datetime.timedelta]:
+    """Calculate the amount of time left until midnight EST/UTC-5."""
```
HassanAbouelela, line 402: This isn't really important, but if you're generalizing this function, you might as well accept a timezone parameter, instead of only returning EST (you'd just have to move the EST constant into the function and pass it the timezone).

HassanAbouelela (approved):
Everything looks good. Whether my comment gets actioned or not does not matter much.

bot/exts/christmas/advent_of_code/_helpers.py
ks129: This function is meant to be used in AoC cog and utility, so having it accept here timezone doesn't have point. When we make this helper function for all cogs, then this may have reason.

SebastiaanZ: I don't think it's necessary at the moment to factor it out at the moment.

PR #540, opened 2020-12-02
Modify error handler check for locally handled errors.

### LINKED ISSUE #538: Normalize check for existing error handler between @Sir-lancebot and @python bots
## Description
Currently, the way we check if an error is handled locally in the cog/command is different in both the bots.
@Sir-lancebot checks if the `command` has the attribute `on_error` whereas @Python checks if the `error` has the attribute `handled`

## Proposed Implementation
Use @Python bot's way of checking if the error is handled locally, i.e, to check if the `error` has the attribute `handled`.

This is being done to keep the bots on the same page.

RohanJnr:
closes #538 

Error handler now checks if the error has the attribute "handled" for locally handled errors instead of checking if the command has the attribute "on_error".

## Did you:

### DIFF
--- a/bot/exts/evergreen/error_handler.py
+++ b/bot/exts/evergreen/error_handler.py
@@ -42,8 +42,8 @@ class CommandErrorHandler(commands.Cog):
     @commands.Cog.listener()
     async def on_command_error(self, ctx: commands.Context, error: commands.CommandError) -> None:
         """Activates when a command opens an error."""
-        if hasattr(ctx.command, 'on_error'):
-            logging.debug("A command error occured but the command had it's own error handler.")
+        if hasattr(error, 'handled'):
+            logging.debug(f"Command {ctx.command} had its error already handled locally; ignoring.")
             return
 
         error = getattr(error, 'original', error)
--- a/bot/exts/evergreen/error_handler.py
+++ b/bot/exts/evergreen/error_handler.py
@@ -42,7 +42,7 @@ class CommandErrorHandler(commands.Cog):
     @commands.Cog.listener()
     async def on_command_error(self, ctx: commands.Context, error: commands.CommandError) -> None:
         """Activates when a command opens an error."""
-        if hasattr(error, 'handled'):
+        if getattr(error, 'handled', False):
             logging.debug(f"Command {ctx.command} had its error already handled locally; ignoring.")
             return
 
--- a/bot/exts/evergreen/snakes/_snakes_cog.py
+++ b/bot/exts/evergreen/snakes/_snakes_cog.py
@@ -15,7 +15,7 @@ import aiohttp
 import async_timeout
 from PIL import Image, ImageDraw, ImageFont
 from discord import Colour, Embed, File, Member, Message, Reaction
-from discord.ext.commands import BadArgument, Bot, Cog, CommandError, Context, bot_has_permissions, group
+from discord.ext.commands import Bot, Cog, CommandError, Context, bot_has_permissions, group
 
 from bot.constants import ERROR_REPLIES, Tokens
 from bot.exts.evergreen.snakes import _utils as utils
@@ -1131,21 +1131,11 @@ class Snakes(Cog):
     @video_command.error
     async def command_error(self, ctx: Context, error: CommandError) -> None:
         """Local error handler for the Snake Cog."""
-        embed = Embed()
-        embed.colour = Colour.red()
-
-        if isinstance(error, BadArgument):
-            embed.description = str(error)
-            embed.title = random.choice(ERROR_REPLIES)
-
-        elif isinstance(error, OSError):
+        if isinstance(error, OSError):
+            error.handled = True
+            embed = Embed()
+            embed.colour = Colour.red()
             log.error(f"snake_card encountered an OSError: {error} ({error.original})")
             embed.description = "Could not generate the snake card! Please try again."
             embed.title = random.choice(ERROR_REPLIES)
-
-        else:
-            log.error(f"Unhandled tag command error: {error} ({error.original})")
-            return
-
-        await ctx.send(embed=embed)
-    # endregion
+            await ctx.send(embed=embed)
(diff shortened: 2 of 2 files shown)


janine9vn (approved):
I tested this change locally on my test server and I haven't found anything breaking. It's a simple enough change and I couldn't find anywhere else in the code that uses the `on_error` attribute for the `ctx.command`.

bot/exts/evergreen/error_handler.py
```diff
@@ -42,8 +42,8 @@ def error_embed(message: str, title: Union[Iterable, str] = ERROR_REPLIES) -> Em
     @commands.Cog.listener()
     async def on_command_error(self, ctx: commands.Context, error: commands.CommandError) -> None:
         """Activates when a command opens an error."""
-        if hasattr(ctx.command, 'on_error'):
-            logging.debug("A command error occured but the command had it's own error handler.")
+        if hasattr(error, 'handled'):
```
jb3, line 45: I'm tempted to say we should look at the value of handled here. Is there a situation where it could be false? I would opt for something like:
```suggestion
        if getattr(error, 'handled', False):
```

jb3 (changes requested):
Just some thoughts, but nothing major. Feel free to correct me if I am wrong :smile:

bot/exts/evergreen/error_handler.py
RohanJnr: I only reason I could think of settings the value to false is when you want the local error handler and the global error handler to handle the error:
```python
if isinstance(error, BadArgument):
    # Maybe user just wants to log some data in the local error handler
    error.handled = False  # default error handling by global error handler.
```
The only way it is kinda helping is readability by convaying the message that both the error handlers should be hit.
These situation are kinda rare and I thinking just not settings the attribute will  be easier.

jb3: I think if there is a situation that makes sense we should try account for it, but I'm open to other opinions.

Those proposed situations don't sound like they are that unlikely, just not something we do now.

RohanJnr: alright 👍

MarkKoz (changes requested):
The snakes cog has a command error handler for several commands. The handled attribute needs to be set on the errors it handles. Maybe the handler can be removed completely, but it does have a special case for `OSError` for some unknown reason.

MarkKoz:
There's also an `on_command_error` in the bot subclass. The error handled in there may also require the attribute, but I'm not sure if it will propagate from there.

bot/exts/evergreen/snakes/_snakes_cog.py
```diff
@@ -1131,21 +1131,11 @@ async def zen_command(self, ctx: Context) -> None:
...
         """Local error handler for the Snake Cog."""
-        embed = Embed()
-        embed.colour = Colour.red()
-
-        if isinstance(error, BadArgument):
-            embed.description = str(error)
-            embed.title = random.choice(ERROR_REPLIES)
-
-        elif isinstance(error, OSError):
+        if isinstance(error, OSError):
```
MarkKoz, line 1134: I realised this actually doesn't even do anything since a `CommandError`'s class hierarchy has no relation to `OSError`. It would have to use  `error.original`, if that attribute even happens to exist. And again, I'm not even sure why this has a special case for `OSError` in the first place. Does anyone else know if this is even needed?

Akarys42: I am guessing this is how the original cog was designed, it doesn't seem very needed now though.

SebastiaanZ: I have a radical solution: If none of us can think of a reason for it being here after reading the code and it's not documented, we should remove it. It would get this PR going (which is important for the Advent of Code PR, which we ideally merge soon, as the event is ongoing) and, if we do that it was needed, we'll notice that at some point and can add it back in.

I think it's preferable to having some vestigial piece of code in our cog that's non-critical (it's not part of the core features), seemingly useless, has no known purpose, but is blocking development progress.

SebastiaanZ: Right, so, after some thinking, I think an `OSError` may be raised if we failed to download any bytes for an image. It would be a bit weird for this to happen, but I think `Image.open` raises an OSError for an empty BytesIO read.

SebastiaanZ:
What needs to happen to make this move forward? I can look into the `isinstance` check, if needed, to get this PR to a mergeable state. Normally, I'd not mind a few extra days (or even weeks), but this PR is currently blocking the PR to extend/change the Advent of Code channels, which is obviously an ongoing event. Would be nice to get them both merged at some point in the near future, so we can benefit during the current edition.

bot/exts/evergreen/snakes/_snakes_cog.py
```diff
@@ -1126,26 +1126,15 @@ async def zen_command(self, ctx: Context) -> None:
...
-        embed = Embed()
-        embed.colour = Colour.red()
-
-        if isinstance(error, BadArgument):
-            embed.description = str(error)
-            embed.title = random.choice(ERROR_REPLIES)
-
-        elif isinstance(error, OSError):
-            log.error(f"snake_card encountered an OSError: {error} ({error.original})")
+        original_error = error.original
```
MarkKoz, line 1132: Not all errors will have this attribute, so use `getattr` or `hasattr`.

MarkKoz (changes requested):
Sorry for a late response; I've been busier this week. I also haven't been as vigilant with GitHub notification since I've disabled e-mail notifications and started relying on the browser add-on.

Anyway, I didn't realise I was holding up another PR over some relatively insignificant feature. However, it seems almost fixed by this point.

bot/exts/evergreen/snakes/_snakes_cog.py
RohanJnr: alright

MarkKoz (approved):

SebastiaanZ:
> Sorry for a late response; I've been busier this week. I also haven't been as vigilant with GitHub notifications since I've disabled e-mail notifications and started relying on the browser add-on.

Just to be sure, no issue, obviously. I was trying to put a feeler out for if Rohan/Iceman would have time to work on it or if we could solve in some other way.

PR #541, opened 2020-12-02
Aoc2020

janine9vn:
## Description & Reasoning
Some of the advent of code commands can be rather "spammy" and clog up the primary channel when there's good discussion going on. This PR aims to limit some of the advent of code commands to `#aoc-commands` and only allow the shorter ones that are more likely to foster discussion in the primary channel. 

## Screenshots
Image of what happens if a user tries to use the `.aoc stats` command in the primary channel, followed by subscribe.
image

## Additional Details
Requires #540 to be merged first.

## Did you:

### DIFF
--- a/bot/exts/christmas/advent_of_code/_cog.py
+++ b/bot/exts/christmas/advent_of_code/_cog.py
@@ -21,7 +21,7 @@ AOC_REQUEST_HEADER = {"user-agent": "PythonDiscord AoC Event Bot"}
 
 COUNTDOWN_STEP = 60 * 5
 
-AOC_WHITELIST = WHITELISTED_CHANNELS + (Channels.advent_of_code,)
+AOC_WHITELIST = WHITELISTED_CHANNELS + (Channels.advent_of_code_commands,)
 
 
 async def countdown_status(bot: commands.Bot) -> None:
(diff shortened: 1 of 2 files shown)


bot/exts/christmas/advent_of_code/_cog.py
```diff
@@ -349,3 +353,11 @@ def _build_about_embed(self) -> discord.Embed:
...
         about_embed.set_footer(text="Last Updated")
         return about_embed
+
+    async def cog_command_error(self, ctx: commands.Context, error: Exception) -> None:
+        """Custom error handler if an advent of code command was posted in the wrong channel."""
+        if isinstance(error, InChannelCheckFailure):
+            await ctx.send(f":x: Please use <#{Channels.advent_of_code_commands}> for aoc commands instead, please.")
+            error.handled = True
+        else:
+            raise error
```
MarkKoz, line 363: I don't believe it's desirable to re-raise the  error because I'm concerned it then won't get handled properly by the global command error handler. That aforementioned handler will receive the error without it needing to be raised here.

MarkKoz (changes requested):

bot/exts/christmas/advent_of_code/_cog.py
janine9vn: Noted! I was unsure if having the custom cog error handler would stop the error from propagating through. New commit has removed the re-raising of the error.

bot/exts/christmas/advent_of_code/_cog.py
```diff
@@ -349,3 +353,9 @@ def _build_about_embed(self) -> discord.Embed:
 
         about_embed.set_footer(text="Last Updated")
         return about_embed
+
+    async def cog_command_error(self, ctx: commands.Context, error: Exception) -> None:
+        """Custom error handler if an advent of code command was posted in the wrong channel."""
+        if isinstance(error, InChannelCheckFailure):
+            await ctx.send(f":x: Please use <#{Channels.advent_of_code_commands}> for aoc commands instead, please.")
```
MarkKoz, line 360: I think one "please" is enough 😄

MarkKoz (approved):

mbaruh (approved):

PR #542, opened 2020-12-05
Use custom status embeds for workflow runs

SebastiaanZ:
This pull request introduces custom status embeds for our workflow runs. The standard embeds GitHub sends to our webhook do not contain a lot of information and, because they're being sent for every individual workflow, generate a lot of noise in our `#dev-log` channel. By using a custom action to send embeds, we can fine tune when we want to send out an embed.

**Note:** The required webhook token has been added an organization secret and the default "check run" embeds Discord sends are already disabled, as we were running into rate limit issues.

## The embeds

The embeds come in two distinct flavors: for pull request check runs and for general workflows. The reason is that when we're dealing with a pull request, we can include information about the pull request itself; when we're not dealing with a pull_request (e.g., a push to master), we obviously don't have such information to display.

### Pull Request Embed

This embed includes information on which pull request triggered the run (analogous to other GitHub embeds) and the embed author. It also includes the source branch, which will be prefixed in the usual way when the PR was made from a fork (i.e., `owner:branch` instead of `branch`). 

### General Workflow Embed

This embed contains the relevant information for a non-PR workflow run. The "commit" links to the commit that triggered the workflow.

## Selected Workflows and Conditions

I've only added the action to the workflows in the chain `lint->build` (or just `lint` for PRs), as I think these are the only workflows we'd want to get an embed for.

In case of a `pull_request` event, the status embed at the end of the `lint` workflow is always sent, regardless of exit status. This means that we get informed in case of success, failure, and cancellation. 

In case of a `push` event, the `lint` embed will only be sent if the workflow failed, as it means our deploy sequence grinds to a halt. If `lint` succeeds, it does not send an embed, but rather moves on to the `build` workflow. The job in the `build` workflow (which only runs if `lint` was a success) always sends an embed to inform us on whether or not building/deploying the bot worked.


Akarys42 (approved):
The embed looks awesome, I love it!

Den4200 (approved):
Nice work!

PR #552, opened 2021-01-08
Migrate to IGDB V4 API

### LINKED ISSUE #548: Migrate IGDB v3 API to v4 for games command
## Description

IGDB shut down APIv3 in October 2020. `games` command is currently built to use v3 of API, but now, when URLs is gone, this command is currently not usable. Command must be migrated APIv4, about to what migrating can be read https://api-docs.igdb.com/#upgrading-to-v4-from-v3.

## Steps to Reproduce

When you have API key of IGDB, then genres refreshing task fails, because API not found, also command fails.

## Expected Behaviour

Should successfully refresh genres and run commands.

## Actual Behaviour

Genres refreshing and command is failing.

## Known Impacted Platforms

- [x] Web
- [x] Desktop
- [x] Android App
- [x] iOS App

## Possible Solutions

Only way is migrating from APIv3 to APIv4.

ChrisLovering:
## Relevant Issues

Closes #548

## TODOs
- [x] Use OAuth flow to get an access token
- [x] Try to renew access token before it expires
- [x] Change all interactions with API to use new endpoints
- [x] Make sure the data we use is still available and in the same format
- [x] Deal with the new rate limits imposed by the V4 API

## Description

1. Changed env vars from a single api token, to the client id and client secret 
2. Added a background task to the Games cog to re-fetch an OAuth key before the current one expires, currently set to do it 2 days before expiry, but can be easily changed if we want more notice.
3. If first attempt to get OAuth access token fails, unload the cog.
4. If attempts to renew OAuth access token fail, raise error but continue to use current token.
5. Delay the start of the `refresh_genres_task` until we get an OAuth access token.
6. Change all `get` requests to `post` requests.

## Reasoning

1. This is required by the new OAuth flow of V4
2. Instead of using a set period of time and using a `@task.loop()` a background task like this allows us to vary the amount of time between expiry and when we renew.
3. The env vars aren't valid, so no need to keep the cog loaded
4. This could be caused by the client id/secret being changed. If this happens, the access token will still be valid, so we can keep using it, but raise error so it can be actioned.
5. This will fail without a valid access token, so we need to wait until the first iteration of the renewal loop
6. This is a requirement of the V4 API

## Additional Details

Added imports of `asyncio.sleep` for the renewal loop, and `datetime.timedelta` to make it easy to format the valid length of the current access token

V4 of the API introduces rate limits on the endpoints. These seem to be getting handled automatically, so no conscious effort has been made to deal with them in this PR.

## Did you:

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -174,7 +174,7 @@ class Emojis:
     christmas_tree = "\U0001F384"
     check = "\u2611"
     envelope = "\U0001F4E8"
-    trashcan = "<:trashcan:796854840293589003>"
+    trashcan = "<:trashcan:637136429717389331>"
     ok_hand = ":ok_hand:"
 
     dice_1 = "<:dice_1:755891608859443290>"
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -174,7 +174,7 @@ class Emojis:
     christmas_tree = "\U0001F384"
     check = "\u2611"
     envelope = "\U0001F4E8"
-    trashcan = "<:trashcan:637136429717389331>"
+    trashcan = "<:trashcan:796854840293589003>"
     ok_hand = ":ok_hand:"
 
     dice_1 = "<:dice_1:755891608859443290>"
@@ -257,7 +257,8 @@ class Tokens(NamedTuple):
     youtube = environ.get("YOUTUBE_API_KEY")
     tmdb = environ.get("TMDB_API_KEY")
     nasa = environ.get("NASA_API_KEY")
-    igdb = environ.get("IGDB_API_KEY")
+    igdb_client_id = environ.get("IGDB_CLIENT_ID")
+    igdb_client_secret = environ.get("IGDB_CLIENT_SECRET")
     github = environ.get("GITHUB_TOKEN")
 
 
--- a/bot/exts/evergreen/game.py
+++ b/bot/exts/evergreen/game.py
@@ -171,7 +171,10 @@ class Games(Cog):
                             f"OAuth response message: {result['message']}"
                         )
                     else:
-                        logger.warning("Invalid OAuth credentials. Unloading Games cog.")
+                        logger.warning(
+                            "Invalid OAuth credentials. Unloading Games cog. "
+                            f"OAuth response message: {result['message']}"
+                        )
                         self.bot.remove_cog('Games')
 
                     return
--- a/bot/exts/evergreen/game.py
+++ b/bot/exts/evergreen/game.py
@@ -23,6 +23,9 @@ BASE_URL = "https://api.igdb.com/v4"
 CLIENT_ID = Tokens.igdb_client_id
 CLIENT_SECRET = Tokens.igdb_client_secret
 
+# The number of seconds before expiry that we attempt to re-fetch a new access token
+ACCESS_TOKEN_RENEWAL_WINDOW = 60*60*24*2
+
 # URL to request API access token
 OAUTH_URL = "https://id.twitch.tv/oauth2/token"
 
@@ -168,8 +171,8 @@ class Games(Cog):
 
             self.access_token = result["access_token"]
 
-            # Set next renewal to 2 days before expire time
-            next_renewal = result["expires_in"] - 60*60*24*2
+            # Attempt to renew before the token expires
+            next_renewal = result["expires_in"] - ACCESS_TOKEN_RENEWAL_WINDOW
 
             time_delta = timedelta(seconds=next_renewal)
             logger.info(f"Successfully renewed access token. Refreshing again in {time_delta}")
--- a/bot/exts/evergreen/game.py
+++ b/bot/exts/evergreen/game.py
@@ -2,7 +2,8 @@ import difflib
 import logging
 import random
 import re
-from datetime import datetime as dt
+from asyncio import sleep
+from datetime import datetime as dt, timedelta
 from enum import IntEnum
 from typing import Any, Dict, List, Optional, Tuple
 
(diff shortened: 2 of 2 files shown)


bot/exts/evergreen/game.py
```diff
@@ -136,8 +152,33 @@ def __init__(self, bot: Bot):
...
+
         self.refresh_genres_task.start()
 
+    async def renew_access_token(self) -> None:
+        """Refeshes V4 access token 2 days before expiry."""
+        while True:
+            async with self.http_session.post(OAUTH_URL, params=OAUTH_PARAMS) as resp:
+                result = await resp.json()
+                if resp.status != 200:
+                    logger.error(
```
ChrisLovering, line 165: Let me know if this is better suited to a a logger.warning, rather than error

bot/exts/evergreen/game.py
```diff
@@ -136,8 +152,33 @@ def __init__(self, bot: Bot):
...
+                        "Failed to renew IGDB access token, unloading Games cog."
+                        f"OAuth response message: {result['message']}"
+                    )
+                    self.bot.remove_cog('Games')
```
ChrisLovering, line 169: I wasn't sure whether to unload the cog, or continue to use the current API token until it expires.

bot/exts/evergreen/game.py
```diff
@@ -136,8 +152,33 @@ def __init__(self, bot: Bot):
 
         self.genres: Dict[str, int] = {}
 
+        self.bot.loop.create_task(self.renew_access_token())
+
         self.refresh_genres_task.start()
 
+    async def renew_access_token(self) -> None:
+        """Refeshes V4 access token 2 days before expiry."""
```
ChrisLovering, line 160: MD: Change this comment since I changed the period to a constant

bot/exts/evergreen/game.py
```diff
@@ -17,10 +18,25 @@
...
-BASE_URL = "https://api-v3.igdb.com"
+BASE_URL = "https://api.igdb.com/v4"
 
-HEADERS = {
-    "user-key": Tokens.igdb,
+CLIENT_ID = Tokens.igdb_client_id
+CLIENT_SECRET = Tokens.igdb_client_secret
+
+# The number of seconds before expiry that we attempt to re-fetch a new access token
+ACCESS_TOKEN_RENEWAL_WINDOW = 60*60*24*2
```
ChrisLovering, line 27: Currently set to 2. This means if renewal fails, we will have 2 days grace where the current access token will still work.

Xithrius: Thank you for confirming this.

Xithrius (approved):
LGTM!

ChrisLovering:
https://pythondiscord.com/pages/contributing/sir-lancebot/sir-lancebot-env-var-reference/#tokensapis
Will need to be updated, The env var there has been replaced with 2 env vars. The current link in that table correctly explains how to get them.

bot/constants.py
```diff
@@ -257,7 +257,8 @@ class Tokens(NamedTuple):
     youtube = environ.get("YOUTUBE_API_KEY")
     tmdb = environ.get("TMDB_API_KEY")
     nasa = environ.get("NASA_API_KEY")
-    igdb = environ.get("IGDB_API_KEY")
+    igdb_client_id = environ.get("IGDB_CLIENT_ID")
+    igdb_client_secret = environ.get("IGDB_CLIENT_SECRET")
```
ChrisLovering, line 261: https://pythondiscord.com/pages/contributing/sir-lancebot/sir-lancebot-env-var-reference/#tokensapis
Will need to be updated to reflect these env var changes.
The current link in the reference table correctly explains how to get them.

The old env var is not used anywhere now, so it has been deleted.
These 2 new env vars will need to be available within the bot's environment when this goes live (preferably before).

bot/exts/evergreen/game.py
```diff
@@ -135,8 +151,47 @@ def __init__(self, bot: Bot):
...
+                            f"OAuth response message: {result['message']}"
+                        )
+                    else:
+                        logger.warning(
+                            "Invalid OAuth credentials. Unloading Games cog. "
+                            f"OAuth response message: {result['message']}"
+                        )
+                        self.bot.remove_cog('Games')
+
+                    return
```
ChrisLovering, line 180: Reviewing my own code a few days after writing.
I'm not sure if its 100% obvious what this return is doing, this might warrant a comment to explain this is used to stop the bg task now that we can't renew/unloaded the cog

bot/exts/evergreen/game.py
```diff
@@ -135,8 +151,47 @@ def __init__(self, bot: Bot):
...
-
-        self.refresh_genres_task.start()
+        self.headers = BASE_HEADERS
+
+        self.bot.loop.create_task(self.renew_access_token())
+
+    async def renew_access_token(self) -> None:
+        """Refeshes V4 access token a number of seconds before expiry. See `ACCESS_TOKEN_RENEWAL_WINDOW`."""
+        while True:
+            async with self.http_session.post(OAUTH_URL, params=OAUTH_PARAMS) as resp:
```
ChrisLovering, line 161: I am wondering if its worth pulling this out into another function for ease of parsing the code?

bot/exts/evergreen/game.py
ks129: This seems pretty obvious when looking at logging.

bot/exts/evergreen/game.py
ks129: This does one thing and should be in one function.

ks129 (approved):

Den4200:
Needed variables have been added to the production environment.

PR #554, opened 2021-01-13
Get env vars from .env file

ChrisLovering:
## Description

Removed explicit mentioned of env vars from docker compose, replaced with an env_file

## Reasoning

This means users of the docker-compose workflow won't need to worry about accidentally commiting their docker-compose file which contains all their secrets!

This will pull all of the env vars from the .env file in the project root dir

See [this chat](https://discord.com/channels/267624335836053506/635950537262759947/798892238855471105) for context.

## Additional Details

Before merging we should:

- Update the contributing wiki with these changes
- Announce this change, so users can extract their secrets from the docker-compose file and put them into a .env file before pulling

## Did you:


ChrisLovering:
Some alternatives to this approach would be:
- Explicitly list all env vars in the `enviroment` section of `docker-compose.yml` so that they get pulled into the container from `.env`
- Tell users to put their `.env` file inside `./bot/` (So it get copied to the container) and use dotenv in contants.py to load from there

docker-compose.yml
```diff
@@ -11,14 +11,8 @@ services:
...
       - redis
 
-    environment:
-      - BOT_TOKEN
-      - BOT_DEBUG
-      - BOT_GUILD
-      - BOT_ADMIN_ROLE_ID
-      - CHANNEL_DEVLOG
-      - CHANNEL_COMMUNITY_BOT_COMMANDS
-      - REDIS_HOST=redis
```
MarkKoz, line 21: If this doesn't already have a default value set in Python then it should. Alternatively, the default can remain in the compose file but I prefer defining it in Python.

ChrisLovering: The default is currently set to `redis.default.svc.cluster.local` https://github.com/python-discord/sir-lancebot/blob/c6353a8613d8e30d66a91fc4ed4017b16c4390ab/bot/constants.py#L278

MarkKoz: Okay, that's the production host and it should stay that way. Therefore, it's best to leave it set to `redis` within the compose file since it can spin up a redis container. If running without Docker Compose, I presume users are already told they have to manually enable fakeredis?

ChrisLovering: I've added the default back in.
They are, but it's a little buried in the lancebot wiki atm https://pythondiscord.com/pages/contributing/sir-lancebot/sir-lancebot-env-var-reference/
I'm planning on drafting a guide to be added onto the wiki for the docker setup (As it's not currently documented).

docker-compose.yml
```diff
@@ -12,13 +12,9 @@ services:
...
     environment:
-      - BOT_TOKEN
-      - BOT_DEBUG
-      - BOT_GUILD
-      - BOT_ADMIN_ROLE_ID
-      - CHANNEL_DEVLOG
-      - CHANNEL_COMMUNITY_BOT_COMMANDS
       - REDIS_HOST=redis
+    env_file:
+      - .env
```
MarkKoz, line 17: `docker-compose up` already loads a `.env` automatically if it exists in the same directory as the compose file, but I suppose being explicit isn't bad.

MarkKoz (approved):

docker-compose.yml
ChrisLovering: I haven't been able to produce this behaviour. Having the `.env` file in the root and running `docker-compose up` doens't seem to load the variables automatically

HassanAbouelela (approved):
LGTM, thanks for updating this.

PR #555, opened 2021-01-14
Ports LinePaginator from python-discord bot to fix overflowing line issue

melodicht:
## Relevant Issues

Closes #533 .

## Description

The solution proposed by OP was used. The `LinePaginator` class from python-discord bot was ported over.

## Reasoning

Since OP seems like a credible person, their solution has been implemented. Moreover, I have checked through the code and ensured that it is compatible with sir-lancebot.

## Screenshots

image

## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] If dependencies have been added or updated, run `pipenv lock`?
- [x] **Lint your code** (`pipenv run lint`)?
- [x] Set the PR to **allow edits from contributors**?

PR #572, opened 2021-01-31
Cheat.sh Command - Python cheatsheet

### LINKED ISSUE #570: Python Cheat Sheet Command
## Description
This is just like tldr pages. It is a really good resource and I use them regularly either when trying to help others or wanna look up something. 

## Screenshots
105967971-7e079680-60ac-11eb-8e30-a0e8b7deecf2

## Proposed Implementation

- https://cht.sh/
- It supports a number of languages, so we can default it to python.

Shivansh-007:
## Relevant Issues
Closes #570 

## Description
I had to overwrite the error result of the cht.sh as it was not according to our needs and the response always returned a `200` status code. And i also have defaulted the language to `python` so there is no way to use any other language rn.

### Output too long
Screenshot from 2021-01-31 15-11-32
### Error Embed
Screenshot from 2021-01-31 15-11-43
### Normal Result
Screenshot from 2021-01-31 15-12-51

## Additional Details
I thought i had commit my first commit, but i didn't, and hence the first commits' name is `Fix Import Order`.

## Did you:


bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        Search cheat.sh.
+
+        Gets a post from https://cheat.sh/python/ by default.
+        Usage:
+        --> .cht read json
+        """
+        url = f'https://cheat.sh/python/{quote_plus(" ".join(search_terms))}'
+        headers = {
+            'User-Agent': 'curl/7.68.0'
```
ChrisLovering, line 75: Could do: From https://github.com/chubin/cheat.sh/blob/62a4a586877a8a651b1c52f8398ebbe6dafb6f76/bin/app.py#L76 it looks like the `python-requests` user agent is treated the same as the `curl` user agent. I haven't tested this, but it looks like we don't need to set our own headers here?

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,93 @@
...
+```md
+.cht read json
+.cht hello world
+.cht lambda
+```
+If the problem persists send a message in <#{channel}>
+"""
+
+
+class ChtSh(commands.Cog):
```
ChrisLovering, line 25: Could do: rename the cog to `CheatSheet` or `CheatSh` to be more explicit.

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,93 @@
...
+    def result_fmt(self, url: str, body_text: str) -> t.Tuple[bool, t.Union[str, Embed]]:
+        """Format Result."""
+        if body_text.startswith("#  404 NOT FOUND"):
+            embed = self.fmt_error_embed()
+            return True, embed
+
+        body_space = min(1986 - len(url), 1000)
+
+        if len(body_text) > body_space:
+            description = f"**Result Of cht.sh**\n" \
```
ChrisLovering, line 48: Could do: Make successfully found cheat sheets also output in embeds. This would remove the need for the isEmbed var, and may make the output look better?

ChrisLovering (changes requested):
Only two things I think we should do for this PR. The could do's are optional if you feel like implementing them :)

bot/exts/evergreen/cheatsheet.py
Shivansh-007: > Could do: Make successfully found cheat sheets also output in embeds. This would remove the need for the isEmbed var, and may make the output look better?

Embeds have a limit on number of characters on a single line hence the code block will break.  This is why i didn't add it in a embed.

bot/exts/evergreen/cheatsheet.py
Shivansh-007: We do otherwise  the content we receive is contains all this escape codes and stuff, I had tried it (sometime ago) so just directly used this right now.

ChrisLovering: Huh, that's weird. Looking at their source it should behave the exact same way.

If this works, and without it, it doesn't then leave it here :)

bot/exts/evergreen/cheatsheet.py
ChrisLovering: Ahh yes, we have a line paginator which would be useful here. But it hasn't been ported over from the other bot yet. See https://github.com/python-discord/sir-lancebot/pull/555

bot/exts/evergreen/cheatsheet.py
ChrisLovering: fwiw, this is because the `self.bot.http_session` is a `aiohttp.ClientSession()` so the user agent is `Python/3.8 aiohttp/3.7.3`

ChrisLovering: My PR into cheat.sh was merged (https://github.com/chubin/cheat.sh/pull/280), so you shouldn't have to overwrite the headers now :). See https://discord.com/channels/267624335836053506/635950537262759947/805490234287849513

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,97 @@
...
+from discord.ext.commands import BucketType, Context
+
+from bot.constants import Channels, Colours, ERROR_REPLIES
+
+ERROR_MESSAGE = """
+Unknown cheat sheet. Please try to reformulate your query.
+
+**Examples**:
+```md
+.cht read json
```
ChrisLovering, line 17: Could do: Get the command prefix from `constants.Client.prefix` rather than hard coding it. If you decide to do this, please do it throughout the file, rather than just this line.

This is purely optional, as I don't think there are plans to change lancebot's prefix on the python server any time soon. It only really impacts us as the dev team :)

ChrisLovering (changes requested):
Just 1 optional change. I've tested these changes locally and they work nicely! This should definitely help us out in the help channels, good job! :)

ChrisLovering (approved):
I meant to approve this, as the only change outstanding is optional.

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,97 @@
...
+        Gets a post from https://cheat.sh/python/ by default.
+        Usage:
+        --> .cht read json
+        """
+        url = f'https://cheat.sh/python/{quote_plus(" ".join(search_terms))}'
+
+        escape_tt = str.maketrans({"`": "\\`"})
+        ansi_re = re.compile(r"\x1b\[.*?m")
+
+        async with self.bot.http_session.get(url) as response:
```
Xithrius, line 85: You should probably have the URL as a constant and then format it within the `async with` block, as such:
```suggestion
        async with self.bot.http_session.get(URL.format(quote_plus(" ".join(search_terms))) as response:
```

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,97 @@
...
+        Search cheat.sh.
+
+        Gets a post from https://cheat.sh/python/ by default.
+        Usage:
+        --> .cht read json
+        """
+        url = f'https://cheat.sh/python/{quote_plus(" ".join(search_terms))}'
+
+        escape_tt = str.maketrans({"`": "\\`"})
+        ansi_re = re.compile(r"\x1b\[.*?m")
```
Xithrius, line 83: Make all of these constants.

Xithrius (changes requested):

Shivansh-007:
@Xithrius @ChrisLovering Requested changes mentioned!

bot/constants.py
```diff
@@ -259,6 +254,7 @@ class Roles(NamedTuple):
     rockstars = 458226413825294336
     core_developers = 587606783669829632
     events_lead = 778361735739998228
+    everyone = 267624335836053506
```
HassanAbouelela, line 257: I'd prefer if you call this `everyone_role` to better match the python-bot constants. Could you also remove the developer role from the list?

bot/exts/evergreen/cheatsheet.py
```diff
@@ -67,25 +73,29 @@ def result_fmt(self, url: str, body_text: str) -> t.Tuple[bool, t.Union[str, Emb
...
-
-        escape_tt = str.maketrans({"`": "\\`"})
-        ansi_re = re.compile(r"\x1b\[.*?m")
-
-        async with self.bot.http_session.get(url) as response:
-            result = ansi_re.sub("", await response.text()).translate(escape_tt)
-
-        is_embed, description = self.result_fmt(url, result)
+        category = utils.get(ctx.message.guild.categories, id=Categories.help_in_use)
+        all_allowed_channels = WHITELISTED_CHANNELS + tuple(category.channels)
```
HassanAbouelela, line 86: For the category check, you don't really have to fetch all the channels. You can instead check if `ctx.channel.id == Categories.help_in_use`

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,108 @@
...
+                          f"```python\n{body_text}```\n" \
+                          f"{url}"
+        return False, description
+
+    @commands.command(
+        name="cheat",
+        aliases=("cht.sh", "cheatsheet", "cheat-sheet", "cht"),
+    )
+    @commands.cooldown(1, 10, BucketType.user)
+    @with_role(Roles.everyone_role)
```
Xithrius, line 76: Is this to check if the user is in the server? If not, could you explain why it's here?

Xithrius (commented):
Looks good to me, except this one thing.

bot/exts/evergreen/cheatsheet.py
Shivansh-007: It basically override the channels it can be used in,

1. Allowing everyone to use the command will allow the command to be used in channels.
2. Then i put a check to see whether it is in the white listed channels or the categories, If it is, then it returns None.
3. Else, it continues to give out the output.

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,108 @@
...
+        if not (
+            ctx.channel.category.id == Categories.help_in_use
+            or ctx.channel.id in WHITELISTED_CHANNELS
+        ):
+            return
+
+        async with self.bot.http_session.get(
+                URL.format(search=quote_plus(" ".join(search_terms)))
+        ) as response:
+            result = ANSI_RE.sub("", await response.text()).translate(ESCAPE_TT)
```
Vthechamp22, line 94: The response looks very abrupt to me, could you try wrapping this with the `async with self.bot.typing`?

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,108 @@
...
+
+        is_embed, description = self.result_fmt(
+            URL.format(search=quote_plus(" ".join(search_terms))),
```
Vthechamp22, line 97: You could make a variable holding this and use it in both the places. It won't matter much, but it'll look a little cleaner to me.

bot/exts/evergreen/cheatsheet.py
```diff
@@ -0,0 +1,108 @@
...
+            embed = self.fmt_error_embed()
+            return True, embed
+
+        body_space = min(1986 - len(url), 1000)
+
+        if len(body_text) > body_space:
+            description = f"**Result Of cht.sh**\n" \
+                          f"```python\n{body_text[:body_space]}\n" \
+                          f"... (truncated - too many lines)```\n" \
+                          f"Full results: {url} "
```
Vthechamp22, line 64:
```suggestion
description = ("**Result Of cht.sh**\n"
               f"```python\n{body_text[:body_space]}\n"
               "... (truncated - too many lines)```\n"
               f"Full results: {url} "
            )
```
This would look neater than the `\` escaping of the string I think. Also, if you do change this, please change it for the next `description` variable also. (Also, the formatting isn't very good in this so please fix it when you change it)

Vthechamp22 (changes requested):
It looks nice to me overall, just a few minor changes can be done according to me.

bot/exts/evergreen/cheatsheet.py
Shivansh-007: Alright will do that!

bot/exts/evergreen/cheatsheet.py
Shivansh-007: I didn't get you, sorry.

bot/exts/evergreen/cheatsheet.py
Shivansh-007: The response is pretty past, I don't see the need of that.

bot/exts/evergreen/cheatsheet.py
Vthechamp22: You could make a variable, named `url` or something and assign it to `URL.format(search=quote_plus(" ".join(search_terms)))` and use it wherever you mention `URL.format(search=quote_plus(" ".join(search_terms)))`

bot/exts/evergreen/cheatsheet.py
Vthechamp22: It wasn't very fast for me, especially ones that had a large response like `.cht :learn`

Xithrius (approved):

PR #575, opened 2021-02-04
Fixes Valentine Commands Bug.

### LINKED ISSUE #574: Remove Member Intent Reliant Features From Lovefest
## Description

After discord introduced the intents  system, sir-lancebot no longer has access to the list of members for a role. This has led to a weird situation with some of the commands in the lovefest cog, as they relied on fetching members for a given role.

## Steps to Reproduce

(Make sure your bot does not have the members intent enabled)
1. Sign up for the lovefest role.
2. Try to use one of the commands that rely on fetching members, such as `.bemyvalentine` with no arguments.

## Actual Behaviour

The member fetch now returns an empty list, making the intended purpose fail, or returns a list with very few members, causing them to be bombarded by pings.

## Possible Solutions

[Functions that rely on member intents](https://discordpy.readthedocs.io/en/latest/api.html#discord.Intents.members) being enabled should be removed from the lovefest cog.

(If you find other features that suffer from similar bugs, open a separate issue for them.)

Shivansh-007:
## Relevant Issues
Closes #574 

## Description

- Removes random selection of members who have love fest role which was causing a caching problem since we don't have `intents.members = True`
- Again caused due to intents, the bot was not able to find the user on DMChannel, so i changed it to be passed through the sir-lancebot-commands channel and deletes the their message. Rest it acts similarly, sending the errors/confirmation messages in DMs and sending the valentine message in DMs.
- Instead of taking a random staff if the `whom` is `None` in `.love` command, it would now take the user.

## Reasoning
As the issue was approved.

## Did you:


bot/exts/valentines/be_my_valentine.py
```diff
@@ -118,56 +109,41 @@ async def send_valentine(
...
-            msg = "You are not supposed to use this command in the server, DM the command to the bot."
-            return await ctx.send(msg)
-
-        if user:
-            if Lovefest.role_id not in [role.id for role in user.roles]:
-                message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
-                return await ctx.send(message)
+        if Lovefest.role_id not in [role.id for role in user.roles]:
+            message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
+            await ctx.message.delete()
```
ChrisLovering, line 126: I don't think we should delete the message that invokes the command. Have the command that caused the error s good reference for the user, to see where they went wrong.

bot/exts/valentines/be_my_valentine.py
```diff
@@ -118,56 +109,41 @@ async def send_valentine(
...
+            return await ctx.author.send(message)
 
         if user == ctx.author:
             # Well a user cant valentine himself/herself.
-            return await ctx.send('Come on dude, you cant send a valentine to yourself :expressionless:')
+            await ctx.message.delete()
```
ChrisLovering, line 131: Same as above.

bot/exts/valentines/be_my_valentine.py
```diff
@@ -88,26 +87,18 @@ async def send_valentine(
             msg = "You are supposed to use this command in the server."
             return await ctx.send(msg)
 
-        if user:
-            if Lovefest.role_id not in [role.id for role in user.roles]:
-                message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
-                return await ctx.send(message)
+        if Lovefest.role_id not in [role.id for role in user.roles]:
+            message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
+            return await ctx.send(message)
```
ChrisLovering, line 92: I know this is what it was previously, but I'm not sure why we are returning the message that was sent. I think this should actually be
```
await ctx.send(message)
return
```

bot/exts/valentines/be_my_valentine.py
```diff
@@ -118,56 +109,41 @@ async def send_valentine(
...
             title=f'{emoji_1}{title} {user.display_name}{emoji_2}',
             description=f'{valentine} \n **{emoji_2}From anonymous{emoji_1}**',
             color=Colours.pink
         )
         try:
             await user.send(embed=embed)
+            await ctx.message.delete()
         except discord.Forbidden:
-            await ctx.author.send(f"{user} has DMs disabled, so I couldn't send the message. Sorry!")
+            await ctx.send(f"{user} has DMs disabled, so I couldn't send the message. Sorry!")
```
ChrisLovering, line 146: What was the reason for changing this from a DM? should we do the same change for the rest of the messages sent via DM in this cog?

ChrisLovering (changes requested):

bot/exts/valentines/be_my_valentine.py
Shivansh-007: > Again caused due to intents, the bot was not able to find the user on DMChannel, so i changed it to be passed through the sir-lancebot-commands channel and deletes the their message. Rest it acts similarly, sending the errors/confirmation messages in DMs and sending the valentine message in DMs.

bot/exts/valentines/be_my_valentine.py
Shivansh-007: I deleted it so that it still acts like a "Secret", and confirmation are sent to the user on DMs, so the user always has reference.

bot/exts/valentines/be_my_valentine.py
```diff
@@ -70,44 +70,38 @@ async def remove_role(self, ctx: commands.Context) -> None:
...
-            return await ctx.send(msg)
+            await ctx.send(msg)
+            return
 
-        if user:
-            if Lovefest.role_id not in [role.id for role in user.roles]:
-                message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
-                return await ctx.send(message)
+        if Lovefest.role_id not in [role.id for role in user.roles]:
+            message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
```
Akarys42, line 92:
```suggestion
            message = f"You cannot send a valentine to {user} as they do not have the lovefest role!"
```

bot/exts/valentines/be_my_valentine.py
```diff
@@ -118,56 +112,41 @@ async def send_valentine(
...
-            # This command is only DM specific
-            msg = "You are not supposed to use this command in the server, DM the command to the bot."
-            return await ctx.send(msg)
-
-        if user:
-            if Lovefest.role_id not in [role.id for role in user.roles]:
-                message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
-                return await ctx.send(message)
+        if Lovefest.role_id not in [role.id for role in user.roles]:
+            message = f"You cannot send a valentine to {user} as he/she does not have the lovefest role!"
```
Akarys42, line 128:
```suggestion
            message = f"You cannot send a valentine to {user} as they do not have the lovefest role!"
```

Akarys42 (changes requested):
While we are at it

bot/exts/valentines/be_my_valentine.py
ChrisLovering: That makes sense in this context, agreed.

bot/exts/valentines/be_my_valentine.py
ChrisLovering: Above justification makes sense.

bot/exts/valentines/be_my_valentine.py
```diff
@@ -70,44 +70,38 @@ async def remove_role(self, ctx: commands.Context) -> None:
     @commands.cooldown(1, 1800, BucketType.user)
     @commands.group(name='bemyvalentine', invoke_without_command=True)
     async def send_valentine(
-        self, ctx: commands.Context, user: Optional[discord.Member] = None, *, valentine_type: str = None
+        self, ctx: commands.Context, user: discord.Member, *, valentine_type: str = None
     ) -> None:
         """
         Send a valentine to user, if specified, or to a random user with the lovefest role.
```
ChrisLovering, line 76: Update the doc string:
```suggestion
        Send a valentine to specified user with the lovefest role.
```

bot/exts/valentines/be_my_valentine.py
ChrisLovering: `discord.Forbidden` is raised when the bot tries to DM a user, but they have DMs disabled.
In this case, the person the bot was trying to send the valentine to had DMs disabled.

I  this case though, removing the `author` from this changes it so that the bot does contact the author of the command via DM. If this was your intention, I think we should remove the other `ctx.author.send` too, or wrap those in `discord.Forbidden` try/except blocks too.

bot/exts/valentines/be_my_valentine.py
```diff
@@ -118,56 +112,41 @@ async def send_valentine(
     @commands.cooldown(1, 1800, BucketType.user)
     @send_valentine.command(name='secret')
     async def anonymous(
-        self, ctx: commands.Context, user: Optional[discord.Member] = None, *, valentine_type: str = None
+        self, ctx: commands.Context, user: discord.Member, *, valentine_type: str = None
     ) -> None:
         """
         Send an anonymous Valentine via DM to to a user, if specified, or to a random with the lovefest role.
```
ChrisLovering, line 118: Update doc string
```suggestion
        Send an anonymous Valentine via DM to to a specified user with the lovefest role.
```

ChrisLovering (changes requested):

bot/exts/valentines/be_my_valentine.py
Shivansh-007: I would create a helper function for this with `try...except` blocks.

ChrisLovering (approved):
LGTM

Akarys42 (approved):
The proposed changes seem to fix the command, thank you!

PR #598, opened 2021-02-20
Removed the Question about the Nile

### LINKED ISSUE #592: Remove question about the Nile
https://github.com/python-discord/sir-lancebot/blob/13620a6461a78e43e8b16b9ad35691ff30622332/bot/resources/evergreen/trivia_quiz.json#L194

There's no way to definitively say if the Nile river or Amazon river is larger than the other. While the Amazon indisputably discharges more water, I believe there's some debate about where the Amazon definitively starts, and thus there's no way to say which is longer.

Because this question is ambiguous, it should be removed.

Arez1337:
…rder again

## Relevant Issues
Closes #592

## Description
I deleted the Question about the longest river and then changed all following IDs to match the order again

## Reasoning
The answer to this question is not definitively "nile"

## Did you:

- [x ] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x ] If dependencies have been added or updated, run `pipenv lock`?
- [x ] **Lint your code** (`pipenv run lint`)?
- [x ] Set the PR to **allow edits from contributors**?

### DIFF
--- a/bot/resources/evergreen/trivia_quiz.json
+++ b/bot/resources/evergreen/trivia_quiz.json
@@ -2,36 +2,51 @@
   "retro": [
     {
       "id": 1,
-      "hints": ["It is not a mainline Mario Game, although the plumber is present.", "It is not a mainline Zelda Game, although Link is present."],
+      "hints": [
+        "It is not a mainline Mario Game, although the plumber is present.",
+        "It is not a mainline Zelda Game, although Link is present."
+      ],
       "question": "What is the best selling game on the Nintendo GameCube?",
       "answer": "Super Smash Bros"
     },
     {
       "id": 2,
-      "hints": ["It was released before the 90's.", "It was released after 1980."],
+      "hints": [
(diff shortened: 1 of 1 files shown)


Xithrius:
Can you provide proof that the answer to this question is incorrect? Google shows that the question and answer belong together.

wookie184:
The current question is "What's the name of the largest river in the world?", which I think is quite ambiguous (does it mean length, volume, or discharge?).

[Wikipedia](https://en.wikipedia.org/wiki/List_of_rivers_by_length#Definition_of_length) suggests there is disagreement on what is the longest river, although from what I can tell the Amazon seems to be widely regarded as the river with the largest drainage basin and greatest average discharge.

I think removing the question is fine. We could also change it to something like "What's the name of the river with the largest drainage basin in the world?".

ChrisLovering (approved):
This change itself looks good, and it resolves the linked issue! nice work.

Xithrius (approved):

PR #605, opened 2021-03-02
Add WTF Python Command

### LINKED ISSUE #602: Add Subsection Command for WTFPython
While discussing python-discord/meta#63, a suggestion that came up was to add a command to get a direct link to subsections of the python gotchas list. This can be used to quickly direct people towards relevant sections of this resource.

A possible implementation would fuzzy match the headers on the page, and send the link to the relevant sections.

Shivansh-007:
## Relevant Issues
Closes #602 

## Description
Gets the fuzzy match of a search query, returns `None` if the ratio is below `90` else returns the fuzzy match. If the return is `None`, then we send our custom Error Message else it sends the link to the subsection of the WTF Python README.md.

## Reasoning
Approved by a staff.

## Screenshots
**On Dark Theme**

Screenshot from 2021-03-02 15-22-56

**On Light Theme**

Screenshot from 2021-03-02 15-22-43

(I went blind while teking it, and am helf blind rihgt now, so ignre ayn spellign misteke)

## Did you:

### DIFF
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -60,8 +60,8 @@ class WTFPython(commands.Cog):
                 log.trace("Fetching the latest WTF Python README.md")
 
                 if resp.status == 200:
-                    self.raw = await resp.text()
-                    await self.parse_readme(self.raw)
+                    raw = await resp.text()
+                    self.parse_readme(raw)
                     log.debug("Successfully fetched the latest WTF Python README.md, breaking out of retry loop")
                     break
 
@@ -86,7 +86,7 @@ class WTFPython(commands.Cog):
         else:
             log.debug(f"Extension {verb}ed: `{ext}`.")
 
-    async def parse_readme(self, data: str) -> None:
+    def parse_readme(self, data: str) -> None:
         """
         Parses the README.md into a dict.
 
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -14,7 +14,7 @@ from bot.bot import Bot
 
 log = logging.getLogger(__name__)
 
-WTF_PYTHON_RAW_URL = "http://raw.githubusercontent.com/satwikkansal/wtfpython/master/"
+WTF_PYTHON_RAW_URL = "http://raw.githubusercontent.com/satwikkansal/wtfpython/mster/"
 BASE_URL = "https://github.com/satwikkansal/wtfpython"
 FETCH_TRIES = 3
 
@@ -72,12 +72,13 @@ class WTFPython(commands.Cog):
                     )
 
         if failed_tries == 3:
+            log.error("Couldn't fetch WTF Python README.md after 3 tries, unloading extension.")
             action = Action.UNLOAD
         else:
             action = Action.LOAD
 
         verb = action.name.lower()
-        ext = self.__class__.__name__
+        ext = "bot.exts.evergreen.wtf_python"
 
         try:
             action.value(self.bot, ext)
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -53,7 +53,9 @@ class WTFPython(commands.Cog):
         """
         Parses the README.md into a dict.
 
-        Key is the name of the section and link to it.
+        It parses the readme into the `self.headers` dict,
+        where the key is the heading and the value is the
+        link to the heading.
         """
         table_of_contents = re.findall(
             r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
@@ -71,10 +73,7 @@ class WTFPython(commands.Cog):
 
     def fuzzy_match_header(self, query: str) -> Optional[str]:
         """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
-        match, certainty = process.extractOne(
-            query,
-            list(self.headers.keys())
-        )
+        match, certainty = process.extractOne(query, self.headers.keys())
         return match if certainty > 90 else None
 
     @staticmethod
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -37,12 +37,11 @@ class WTFPython(commands.Cog):
         self.bot = bot
         self.get_wtf_python_readme.start()
 
-        self.data: dict = {}
         self.headers: Dict[str] = dict()
 
     @tasks.loop(hours=1)
     async def get_wtf_python_readme(self) -> None:
-        """Gets content of README.md of WTF Python Repository."""
+        """"Gets the content of README.md from the WTF Python Repository."""
         async with self.bot.http_session.get(WTF_PYTHON_RAW_URL) as resp:
             if resp.status == 200:
                 self.raw = await resp.text()
@@ -71,13 +70,9 @@ class WTFPython(commands.Cog):
                 )
 
     def fuzzy_match_header(self, query: str) -> Optional[str]:
-        """Returns the fizzy match of a query if its ratio is above 90 else returns None."""
-        keys = list(self.headers.keys())
-        match = process.extractOne(query, keys)
-        print(match)
-        if match[1] < 90:
-            return
-        return match[0]
+         """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
+        match, certainty = process.extractOne(query, list(self.headers.keys()))
+        return match if certainty > 90 else None
 
     @staticmethod
     async def make_embed(query: str, link: str) -> Embed:
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -1,0 +1,117 @@
+import logging
+import random
+import re
+from typing import Dict, Optional
+
+from discord import Embed
+from discord.ext import commands, tasks
+from fuzzywuzzy import process
+
+from bot import constants
+from bot.bot import Bot
+
+log = logging.getLogger(__name__)
+
+WTF_PYTHON_RAW_URL = ("http://raw.githubusercontent.com/satwikkansal/"
+                      "wtfpython/master/README.md")
+BASE_URL = "https://github.com/satwikkansal/wtfpython"
+THUMBNAIL = "https://raw.githubusercontent.com/satwikkansal/wtfpython/master/images/logo.png"
+
+ERROR_MESSAGE = f"""
+Unknown WTF Python Query. Please try to reformulate your query.
+
+**Examples**:
+```md
+{constants.Client.prefix}wtf wild imports
+{constants.Client.prefix}wtf subclass
+{constants.Client.prefix}wtf del
+```
+If the problem persists send a message in <#{constants.Channels.dev_contrib}>
+"""
+
+
+class WTFPython(commands.Cog):
+    """Cog that allows users to retrieve issues from GitHub."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.get_wtf_python_readme.start()
+
+        self.data: dict = {}
+        self.headers: Dict[str] = dict()
+
+    @tasks.loop(hours=1)
+    async def get_wtf_python_readme(self) -> None:
+        """Gets content of README.md of WTF Python Repository."""
+        async with self.bot.http_session.get(WTF_PYTHON_RAW_URL) as resp:
+            if resp.status == 200:
+                self.raw = await resp.text()
+                await self.parse_readme(self.raw)
+            else:
+                log.debug(f"Failed to get latest WTF Python README.md. Status code {resp.status}")
+
+    async def parse_readme(self, data: str) -> None:
+        """
+        Parses the README.md into a dict.
+
+        Key is the name of the section and link to it.
+        """
+        table_of_contents = re.findall(
+            r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
+        )[0].split("\n")
+        table_of_contents = list(map(str.strip, table_of_contents))
+
+        for header in table_of_contents:
+            match = re.findall(r"\[▶ (.*)\]\((.*)\)", header)
+            if match:
+                self.headers.update(
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fizzy match of a query if its ratio is above 90 else returns None."""
+        keys = list(self.headers.keys())
+        match = process.extractOne(query, keys)
+        print(match)
+        if match[1] < 90:
+            return
+        return match[0]
+
+    @staticmethod
+    async def make_embed(query: str, link: str) -> Embed:
+        """Generates a embed for a search."""
+        embed = Embed(
+            title=f"WTF Python Search Result For {query}",
+            colour=constants.Colours.dark_green,
+            description=f"[Go to Repository Section]({link})"
+        )
+        embed.set_thumbnail(url=THUMBNAIL)
+        return embed
+
+    @commands.command(aliases=("wtf",))
+    async def wtf_python(self, ctx: commands.Context, *query: str) -> None:
+        """
+        Search wtf python.
+
+        Gets the link of the fuzzy matched query from https://github.com/satwikkansal/wtfpython.
+        Usage:
+            --> .wtf wild imports
+        """
+        query = " ".join(query)
+        match = self.fuzzy_match_header(query)
+        if not match:
+            embed = Embed(
+                title=random.choice(constants.ERROR_REPLIES),
+                description=ERROR_MESSAGE,
+                colour=constants.Colours.soft_red
+            )
+        else:
+            embed = await self.make_embed(query, self.headers[match])
+        await ctx.send(embed=embed)
+
+
+def setup(bot: Bot) -> None:
+    """Load WTFPython Cog."""
+    bot.add_cog(WTFPython(bot))
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -1,6 +1,8 @@
+import functools
 import logging
 import random
 import re
+from enum import Enum
 from typing import Dict, Optional
 
 from discord import Embed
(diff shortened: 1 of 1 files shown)


bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class WTFPython(commands.Cog):
+    """Cog that allows users to retrieve issues from GitHub."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.get_wtf_python_readme.start()
+
+        self.data: dict = {}
```
ChrisLovering, line 40: I don't think you use this anymore.

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,117 @@
...
+        self.headers: Dict[str] = dict()
+
+    @tasks.loop(hours=1)
+    async def get_wtf_python_readme(self) -> None:
+        """Gets content of README.md of WTF Python Repository."""
```
ChrisLovering, line 45: Minor grammar.
```suggestion
        """Gets the content of README.md from the WTF Python Repository."""
```

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,117 @@
...
+                self.raw = await resp.text()
+                await self.parse_readme(self.raw)
+            else:
+                log.debug(f"Failed to get latest WTF Python README.md. Status code {resp.status}")
+
+    async def parse_readme(self, data: str) -> None:
+        """
+        Parses the README.md into a dict.
+
+        Key is the name of the section and link to it.
```
ChrisLovering, line 57: Did you rename `Key` to `data`?

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,117 @@
...
+            match = re.findall(r"\[▶ (.*)\]\((.*)\)", header)
+            if match:
+                self.headers.update(
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fizzy match of a query if its ratio is above 90 else returns None."""
```
ChrisLovering, line 74:
```suggestion
        """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
```

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,117 @@
...
+        keys = list(self.headers.keys())
+        match = process.extractOne(query, keys)
+        print(match)
```
ChrisLovering, line 77: Remove debug print, or make it a `log.trace()`

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,117 @@
...
+            if match:
+                self.headers.update(
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fizzy match of a query if its ratio is above 90 else returns None."""
+        keys = list(self.headers.keys())
```
ChrisLovering, line 75: Is this wrapper needed? can `process.extractOne()` take `self.headers.keys()` directly?

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,117 @@
...
+        match = process.extractOne(query, keys)
```
ChrisLovering, line 76: Maybe unpacking this type into `match, certainty` could help with readability of the if statement below.

ChrisLovering (changes requested):

bot/exts/evergreen/wtf_python.py
Shivansh-007: No, it is talking about how it parses the `README.md` into `self.headers`

ChrisLovering: Ahh, so `self.headers` is a `name: link` dict?

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,115 @@
...
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
+        match, certainty = process.extractOne(
+            query,
+            list(self.headers.keys())
```
ChrisLovering, line 76: Is the cast to list needed?

bot/exts/evergreen/wtf_python.py
Shivansh-007: > Ahh, so `self.headers` is a `name: link` dict? If so, could you update this docstring to say that it parses the readme into the `self.headers` dict, where the key is the heading and the value is the link to the heading?

Yes you are right! Alright will update the `doc-string`.

bot/exts/evergreen/wtf_python.py
Shivansh-007: Yes
```py
d = {'a':1, 'b':2, 'c':3}
print(d.keys())
print(type(d.keys()))
```
Will give:
```py
001 | dict_keys(['a', 'b', 'c'])
002 | <class 'dict_keys'>
```

ChrisLovering: Yea, I know that, but as far as I can see `process.extractOne()` will work with a `dict_keys` class directly as its an iterable.

Shivansh-007: Ah right,

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,114 @@
...
+THUMBNAIL = "https://raw.githubusercontent.com/satwikkansal/wtfpython/master/images/logo.png"
+
+ERROR_MESSAGE = f"""
+Unknown WTF Python Query. Please try to reformulate your query.
+
+**Examples**:
+```md
+{constants.Client.prefix}wtf wild imports
+{constants.Client.prefix}wtf subclass
+{constants.Client.prefix}wtf del
```
ChrisLovering, line 27: Running both `.wtf del` and `.wtf subclass` results in the error embed, do these need to be updated?

ChrisLovering (changes requested):
Code looks good, just one issue while testing locally. I think it could be a good idea to have a `.wtf list` or similar, to output all of the available pages. Maybe one for after the paginator PR in merged.

bot/exts/evergreen/wtf_python.py
Shivansh-007: Maybe I could lower down the certainty check.

Shivansh-007:
> Code looks good, just one issue while testing locally. I think it could be a good idea to have a `.wtf list` or similar, to output all of the available pages. Maybe one for after the paginator PR in merged.

Alright will do.

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,116 @@
...
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
+        match, certainty = process.extractOne(query, self.headers.keys())
+        return match if certainty > MINIMUM_CERTAINTY else None
+
+    @staticmethod
+    async def make_embed(query: str, link: str) -> Embed:
```
ToxicKidz, line 82: Why is this an async function? I don't see it `await`'ing anything.
```suggestion
    def make_embed(query: str, link: str) -> Embed:
```

ToxicKidz (changes requested):

bot/exts/evergreen/wtf_python.py
Xithrius: I don't think you can create an embed object within a sync function.

anand2312: Why not?  Sync function

Xithrius: Huh, I guess I was wrong. maybe in previous versions that was a thing? Not sure.

ToxicKidz: > I don't think you can create an embed object within a sync function.

Why not?

ToxicKidz (approved):
LGTM!

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @tasks.loop(hours=1)
+    async def get_wtf_python_readme(self) -> None:
+        """Gets the content of README.md from the WTF Python Repository."""
+        async with self.bot.http_session.get(WTF_PYTHON_RAW_URL) as resp:
+            if resp.status == 200:
+                self.raw = await resp.text()
+                await self.parse_readme(self.raw)
+            else:
+                log.debug(f"Failed to get latest WTF Python README.md. Status code {resp.status}")
```
RohanJnr, line 52: What happens if the status is not 200?
Then `self.headers` will be an empty dict, then the command fails as line 78 
```python
match, certainty = process.extractOne(query, self.headers.keys())
```
fails with the error
image
and the user does not get any output.

IMO, we should disable to command if we cannot fetch the data from the website/readme or if the response status is not 200 and retry to fetch data for a few times at regular intervals.

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,116 @@
...
+        self.get_wtf_python_readme.start()
+
+        self.headers: Dict[str] = dict()
+
+    @tasks.loop(hours=1)
+    async def get_wtf_python_readme(self) -> None:
+        """Gets the content of README.md from the WTF Python Repository."""
+        async with self.bot.http_session.get(WTF_PYTHON_RAW_URL) as resp:
+            if resp.status == 200:
+                self.raw = await resp.text()
```
RohanJnr, line 49: why is this an instance attribute instead of a local variable?

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,151 @@
...
+        ext = self.__class__.__name__
+
+        try:
+            action.value(self.bot, ext)
+        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded):
+            log.debug(f"Extension `{ext}` is already  is already {verb}ed.")
+        else:
+            log.debug(f"Extension {verb}ed: `{ext}`.")
+
+    async def parse_readme(self, data: str) -> None:
```
anand2312, line 89: This also doesn't seem to be awaiting anything. Why's it under an `async def`?

anand2312 (changes requested):

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,151 @@
...
+                        f"Failed to get latest WTF Python README.md on try {x}/{FETCH_TRIES}. Status code {resp.status}"
+                    )
+
+        if failed_tries == 3:
+            action = Action.UNLOAD
+        else:
+            action = Action.LOAD
+
+        verb = action.name.lower()
+        ext = self.__class__.__name__
```
anand2312, line 80: As for your issue about the extension not being unloaded after too many tries, I think it is possible the error is here - a [Cog](https://discordpy.readthedocs.io/en/latest/ext/commands/cogs.html) _isn't_ an [Extension](https://discordpy.readthedocs.io/en/latest/ext/commands/extensions.html). You're trying to unload the extension by using the cog name, not the extension name (the extension name will be the file name - `exts.evergreen.wtf_python`)

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,151 @@
...
+
+        try:
+            action.value(self.bot, ext)
+        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded):
+            log.debug(f"Extension `{ext}` is already  is already {verb}ed.")
```
anand2312, line 85: Grammer here - `f"Extension {ext} is already {verb}ed|`

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,151 @@
...
+        embed = Embed(
+            title=f"WTF Python Search Result For {query}",
+            colour=constants.Colours.dark_green,
+            description=f"[Go to Repository Section]({link})"
+        )
+        embed.set_thumbnail(url=f"{WTF_PYTHON_RAW_URL}images/logo.png")
+        return embed
+
+    @commands.command(aliases=("wtf",))
+    async def wtf_python(self, ctx: commands.Context, *query: str) -> None:
```
anand2312, line 128: Was there a reason you chose to use `*query` here instead of using a keyword-only argument (which discord.py uses for a "consume-rest" behaviour)?
```py
async def wtf_python(self, ctx: commands.Context, *, query: str) -> None
```
This will save you the line where you `join` them together. 
https://discordpy.readthedocs.io/en/latest/faq.html#why-do-my-arguments-require-quotes

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,151 @@
...
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
+        match, certainty = process.extractOne(query, self.headers.keys())
+        return match if certainty > MINIMUM_CERTAINTY else None
+
+    @staticmethod
+    def make_embed(query: str, link: str) -> Embed:
```
anand2312, line 117: Does this really have to be a function of it's own, when it's used only once?

anand2312 (approved):
LGTM 👍🏻

ToxicKidz (changes requested):

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+from discord import Embed
+from discord.ext import commands, tasks
+from fuzzywuzzy import process
+
+from bot import constants
+from bot.bot import Bot
+
+log = logging.getLogger(__name__)
+
+WTF_PYTHON_RAW_URL = "http://raw.githubusercontent.com/satwikkansal/wtfpython/mster/"
```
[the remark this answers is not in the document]
Shivansh-007: Whoops, that was a typo.

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+
+MINIMUM_CERTAINTY = 75
+
+
+class Action(Enum):
+    """Represents an action to perform on an extension."""
+
+    # Need to be partial otherwise they are considered to be function definitions.
+    LOAD = functools.partial(Bot.load_extension)
+    UNLOAD = functools.partial(Bot.unload_extension)
```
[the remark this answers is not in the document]
Shivansh-007: `LOAD`, and `UNLOAD` needs to partials as they would be considered as function definitions which is not what we want, if we remove the partials.

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class WTFPython(commands.Cog):
+    """Cog that allows users to retrieve issues from GitHub."""
```
laundmo, line 45: This docstring seems to be a leftover from copying the cog
```suggestion
    """Cog that allows geting WTF Python entries from the WTF Python repository."""
```

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+            match = re.findall(r"\[▶ (.*)\]\((.*)\)", header)
+            if match:
+                self.headers.update(
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
```
laundmo, line 113: Docstring mentions a ratio of 90, though the ratio is a global variable and currently 75
```suggestion
        """Returns the fuzzy match of a query if its ratio is above `MINIMUM_CERTAINTY` else returns None."""
```

laundmo (changes requested):

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+
+                if resp.status == 200:
+                    raw = await resp.text()
+                    self.parse_readme(raw)
+                    log.debug("Successfully fetched the latest WTF Python README.md, breaking out of retry loop")
+                    break
+
+                else:
+                    failed_tries += 1
+                    log.debug(
```
Kronifer, line 70: Seems to me this should be `log.warning`, but thats just me

bot/exts/evergreen/wtf_python.py
Kronifer, line 113: Could we reword this to "Returns the fuzzy match of a query if its ratio is above 90, otherwise returns None", using else sounds bad

bot/exts/evergreen/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
+        match, certainty = process.extractOne(query, self.headers.keys())
+        return match if certainty > MINIMUM_CERTAINTY else None
+
+    @commands.command(aliases=("wtf",))
+    async def wtf_python(self, ctx: commands.Context, *, query: str) -> None:
+        """
+        Search wtf python.
```
Kronifer, line 120: wtf is capitalised in other areas

Kronifer (changes requested):
Few nitpicks, but looks good!

bot/exts/evergreen/wtf_python.py
Shivansh-007: I am logging a `error` level messages if it fails to fetch it 3 times. I don't see the need to do it for every fetch.

bot/exts/evergreen/wtf_python.py
Shivansh-007: Explain "bad", I think `else` is good and makes its functionality clear.

Shivansh-007:
I don't think I am having enough time to look into this and it has been stale for quite a while now, so whoever wants can take over the PR, and deliver the feature!

Xithrius (changes requested):
@decorator-factory has noticed that this PR still uses [`fuzzywuzzy`](https://pypi.org/project/fuzzywuzzy/), which [contains the GPLv2 license](https://github.com/seatgeek/fuzzywuzzy/blob/master/LICENSE.txt). We now use [`rapidfuzz`](https://pypi.org/project/rapidfuzz/) because it uses the MIT license, so this new feature needs to reflect this change.

brad90four:
I think I can take this one, looks like rapidfuzz also has an extractOne method and the change from `fuzzy_wuzzy` to `rapidfuzz` is just the import statement.

Xithrius:
@brad90four I shall assign you. `rapidfuzz` is currently what we're going with for the default replacement of `fuzzy_wuzzy`. Be sure to re-lock `poetry.lock`.

Xithrius:
Given that the restructure of this bot occurred, the main file of `bot/exts/evergreen/wtf_python.py` will need to be moved to `bot/exts/utilities/wtf_python.py`.

brad90four:
Should I get perms to Shivansh's repo in order to push changes to the feature branch or should I make a new PR with a branch from my own repo?

Xithrius:
Make a new PR you will.

PR #607, opened 2021-03-02
Fixes Startup Channel Check Bug

### LINKED ISSUE #606: Startup Channel Check Bug
## Description
#482 Was merged a couple hours ago, and in it, the following change was introduced:
https://github.com/python-discord/sir-lancebot/blob/53af90f92258905c2bfc4c2db2341dfd1fdc0124/bot/bot.py#L74-L82

This logs an error for every single channel we have into sentry.

## Steps to Reproduce

Startup the bot.

## Expected Behaviour

It should only log missing channels.

## Actual Behaviour

All channels are logged as missing.

## Possible Solutions

So this bug is due to the check on line 81:
https://github.com/python-discord/sir-lancebot/blob/53af90f92258905c2bfc4c2db2341dfd1fdc0124/bot/bot.py#L81

All channels is a set of channel objects. Channel ID is an int, so it'll always return false. We should instead compare it to a set of IDs.

## Additional Details

While you're at it, could you please kaizen in a change to ignore this check on dev environments? A lot of these channels can't be set without directly modifying the constants, so invetiably someone will change constants to suppress it, and inventiably it'll be committed and we'll have to clean it up.

Shivansh-007:
## Relevant Issues
Closes #606

## Description
Uses `channel.id`s instead of `channel` object while checking for missing channels. Skips the check if `debug` is set to `False` in the `.env`.

## Did you:

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -74,11 +74,16 @@ class Bot(commands.Bot):
     async def check_channels(self) -> None:
         """Verifies that all channel constants refer to channels which exist."""
         await self.wait_until_guild_available()
-        all_channels = set(self.get_all_channels())
+
+        if constants.Client.debug:
+            log.info("Skipping Channels Check.")
+            return
+
+        all_channels_ids = [channel.id for channel in set(self.get_all_channels())]
         for name, channel_id in vars(constants.Channels).items():
             if name.startswith('_'):
                 continue
-            if channel_id not in all_channels:
+            if channel_id not in all_channels_ids:
                 log.error(f'Channel "{name}" with ID {channel_id} missing')
 
     async def send_log(self, title: str, details: str = None, *, icon: str = None) -> None:
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -79,7 +79,7 @@ class Bot(commands.Bot):
             log.info("Skipping Channels Check.")
             return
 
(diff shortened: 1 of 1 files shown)


bot/bot.py
```diff
@@ -74,11 +74,16 @@ async def on_command_error(self, context: commands.Context, exception: DiscordEx
     async def check_channels(self) -> None:
         """Verifies that all channel constants refer to channels which exist."""
         await self.wait_until_guild_available()
-        all_channels = set(self.get_all_channels())
+
+        if constants.Client.debug:
+            log.info("Skipping Channels Check.")
+            return
+
+        all_channels_ids = [channel.id for channel in set(self.get_all_channels())]
```
HassanAbouelela, line 82: We probably don't have to use a set here, but everything else is fine. Thanks 👍

HassanAbouelela (approved):

mbaruh (approved):
I'll approve for the purpose of the hotfix, but I think there should be a split between channels which are constant are actually required, and ones which are temporary.

An even better approach would be for each cog to specify which channels it needs.

Shivansh-007:
Alright @mbaruh I will have a look at that. Can u tell how should i progress with the second method of implemented. Like how should the channel - cog map be stored/

HassanAbouelela:
@Shivansh-007 This shouldn’t be part of this hot fix. There will be another issue for it.

PR #618, opened 2021-03-09
Wikiguess Game

### LINKED ISSUE #446: Wikiguess game
## Description

A new `.wikiguess` game, where you get points for guessing the correct article name.

This would be a typical game:
- Send `.wikigame` or `.wikigame start` to start a new game.
- Get an embed with instructions about the game (click green reaction tick box to start game)
- X (5?) rounds are done
- The user receives a score

This is how a normal round would look:
- Get the first few sentences from a not too uncommon Wikipedia article
- Get X (10?) seconds to answer what the article is called
- Get result for the round ("was it correct?")

We could *possibly* have a multi player mode as well, where players compete to guess it first.

## Reasoning

This would be a quite fun game to play for people in the community. They may even find some new and interesting topics to check out, since the articles will be linked on the bottom of the pages.

## Proposed Implementation

- Checking if the inputted article is the right one would be done with fuzzy matching, so "Donald Trumpp" would be accepted for "Donald J. Trump" (intended typo)
- The game would probably be played in #seasonalbot-commands. A DM version *could* be available as well, with an option to let moderators choose which one is available, if it gets spammy
- The articles to choose from would be hand picked, in order to keep them free from inappropriate topics
- First sentences may or may not be requested from Wikipedia during the game, or be cached from when they were picked. Preferably live fetching

If anyone would like to implement it, though, I'm not strictly against it.

Shivansh-007:
## Relevant Issues
Closes #446 

## Description & Reasoning
Tis PR adds the wikiguess game. To implement this I have used the random wikipedia article API, to get the title, extract. The questions are formed using the extract, which is then run through a replace function, to remove the title if it is present in the extract with `[redacted]`. 
Another check I have added while fetching the wikiguess question, is to check whether the title contains any non alpha-numeric character (`-+-*/[]{}()!@#$%^&`). As it becomes difficult to answer when the answer is like "Felsőkelecsény" or " Camp Lake, Wisconsin" and you answer "Camp Lake".

## Screenshots
Screenshot from 2021-03-09 14-18-44

## Did you:

### DIFF
--- a/bot/exts/fun/trivia_quiz.py
+++ b/bot/exts/fun/trivia_quiz.py
@@ -3,9 +3,10 @@ import json
 import logging
 import operator
 import random
-from typing import Callable, Optional
+from collections import defaultdict
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Callable, Optional
 
 import discord
 from discord.ext import commands, tasks
@@ -209,7 +210,7 @@ class TriviaQuiz(commands.Cog):
         self.questions = self.load_questions()
         self.question_limit = 0
 
-        self.player_scores = {}  # A variable to store all player's scores for a bot session.
+        self.player_scores = defaultdict(int)  # A variable to store all player's scores for a bot session.
         self.game_player_scores = {}  # A variable to store temporary game player's scores.
 
         self.categories = {
--- a/bot/exts/fun/trivia_quiz.py
+++ b/bot/exts/fun/trivia_quiz.py
@@ -3,12 +3,12 @@ import json
 import logging
 import operator
 import random
+from typing import Callable, Optional
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Callable, Optional
 
 import discord
-from discord.ext import commands
+from discord.ext import commands, tasks
 from rapidfuzz import fuzz
 
 from bot.bot import Bot
@@ -25,41 +25,15 @@ WRONG_ANS_RESPONSE = [
     "Better luck next time...",
 ]
 
-N_PREFIX_STARTS_AT = 5
-N_PREFIXES = [
-    "penta", "hexa", "hepta", "octa", "nona",
-    "deca", "hendeca", "dodeca", "trideca", "tetradeca",
-]
+RULES = (
+    "No cheating and have fun!",
+    "Points for each question reduces by 25 after 10s or after a hint. Total time is 30s per question"
+)
 
-PLANETS = [
-    ("1st", "Mercury"),
-    ("2nd", "Venus"),
-    ("3rd", "Earth"),
-    ("4th", "Mars"),
-    ("5th", "Jupiter"),
-    ("6th", "Saturn"),
-    ("7th", "Uranus"),
-    ("8th", "Neptune"),
-]
-
-TAXONOMIC_HIERARCHY = [
-    "species", "genus", "family", "order",
-    "class", "phylum", "kingdom", "domain",
-]
-
-UNITS_TO_BASE_UNITS = {
-    "hertz": ("(unit of frequency)", "s^-1"),
-    "newton": ("(unit of force)", "m*kg*s^-2"),
-    "pascal": ("(unit of pressure & stress)", "m^-1*kg*s^-2"),
-    "joule": ("(unit of energy & quantity of heat)", "m^2*kg*s^-2"),
-    "watt": ("(unit of power)", "m^2*kg*s^-3"),
-    "coulomb": ("(unit of electric charge & quantity of electricity)", "s*A"),
-    "volt": ("(unit of voltage & electromotive force)", "m^2*kg*s^-3*A^-1"),
-    "farad": ("(unit of capacitance)", "m^-2*kg^-1*s^4*A^2"),
-    "ohm": ("(unit of electric resistance)", "m^2*kg*s^-3*A^-2"),
-    "weber": ("(unit of magnetic flux)", "m^2*kg*s^-2*A^-1"),
-    "tesla": ("(unit of magnetic flux density)", "kg*s^-2*A^-1"),
-}
+RANDOM_WIKI_URL = "https://en.wikipedia.org/api/rest_v1/page/random/summary"
+TRIVIA_QUIZ_ICON = (
+    "https://raw.githubusercontent.com/python-discord/branding/main/icons/trivia_quiz/trivia-quiz-dist.png"
+)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


bot/exts/evergreen/trivia_quiz/_cog.py
```diff
@@ -2,47 +2,91 @@
...
         self.bot = bot
-        self.questions = self.load_questions()
+
+        self.wiki_questions: t.List = []
+
         self.game_status = {}  # A variable to store the game status: either running or not running.
         self.game_owners = {}  # A variable to store the person's ID who started the quiz game in a channel.
-        self.question_limit = 4
         self.player_scores = {}  # A variable to store all player's scores for a bot session.
         self.game_player_scores = {}  # A variable to store temporary game player's scores.
```
RohanJnr, line 57: So currently, we are using 4 variables to store the state of a quiz game.
Would it be better to have a class representing a game? like 
```py
class Game:
    def __init__(self, author, *args *kwargs):
        ...
    
```

and then in the Cog init method, 
we can have a variable `games` (a dict).
the `games` dict would look something like this
```py
{
channel_id: Game(...) class instance,
}
```

Also, another question is if we still want to store game instances, because this was done as the plan was to have this command run in multiple channels.
If the bot is restricted to only 1 channel, then it could become simpler.

bot/exts/evergreen/trivia_quiz/_cog.py
```diff
@@ -0,0 +1,137 @@
...
+        start_id = 201
+
+        while len(self.wiki_questions) != 10:
+            async with self.bot.http_session.get(url=_constants.RANDOM_WIKI_URL) as r:
+                if r.status != 200:
+                    continue
+
+                raw_data = await r.json()
+
+                if any(not c.isalnum() for c in raw_data["title"]):
```
ToxicKidz, line 49:
```suggestion
                if all(c.isalnum() for c in raw_data["title"]):
```
I think this looks better, unless you want it to continue if `raw_data["title"]` is empty (if that's even possible).

bot/exts/evergreen/trivia_quiz/_cog.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @commands.group(name="quiz", aliases=["trivia"], invoke_without_command=True)
+    async def quiz_game(self, ctx: commands.Context, category: str = None) -> None:
+        """
+        Start a quiz!
+
+        Questions for the quiz can be selected from the following categories:
+        - general : Test your general knowledge. (default)
+        """
+        if ctx.channel.id in self.games.keys():
```
ToxicKidz, line 70:
```suggestion
        if ctx.channel.id in self.games:
```

bot/exts/evergreen/trivia_quiz/_constants.py
```diff
@@ -0,0 +1,27 @@
...
+)
+WINNER_RESPONSE = (
+    "Congratulations {winners_mention} :tada: "
+    "{winner_word} have won this quiz game with a grand total of {highest_points} points!"
+)
+
+RULES = {
+    1: "No cheating and have fun!",
+    2: "Points for each question reduces by 25 after 10s or after a hint. Total time is 30s per question"
+}
```
ToxicKidz, line 17: Can this be made a tuple (and use `enumerate` instead later)?

bot/exts/evergreen/trivia_quiz/_helpers.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # Check if more than 1 player has highest points.
+        if no_of_winners > 1:
+            word = "You guys"
+            winners = []
+            points_copy = list(player_data.values()).copy()
+
+            for _ in range(no_of_winners):
+                index = points_copy.index(highest_points)
+                winners.append(list(player_data.keys())[index])
```
ToxicKidz, line 65:
```suggestion
                winners.append(list(player_data)[index])
```

bot/exts/evergreen/trivia_quiz/_cog.py
```diff
@@ -0,0 +1,137 @@
...
+            await ctx.send("No quiz running.")
+
+    @quiz_game.command(name="leaderboard")
+    async def leaderboard(self, ctx: commands.Context) -> None:
+        """View everyone's score for this bot session."""
+        leaderboard = {}
+
+        for game in self.games.values():
+            for player, points in game.player_scores.items():
+                if player in leaderboard.keys():
```
ToxicKidz, line 127:
```suggestion
                if player in leaderboard:
```

bot/exts/evergreen/trivia_quiz/_helpers.py
```diff
@@ -0,0 +1,109 @@
...
+                points_copy[index] = 0
+
+            winners_mention = " ".join(winner.mention for winner in winners)
+        else:
+            word = "You"
+            author_index = list(player_data.values()).index(highest_points)
+            winner = list(player_data.keys())[author_index]
```
ToxicKidz, line 72:
```suggestion
            winner = list(player_data)[author_index]
```

ToxicKidz (changes requested):
Mostly looks good.

bot/exts/evergreen/trivia_quiz/_cog.py
ToxicKidz: Oops it might need a `not` too.

bot/exts/evergreen/trivia_quiz/_game.py
```diff
@@ -0,0 +1,129 @@
...
+        # A variable to store the person's ID who started the quiz game in a channel.
+        self.game_owner = ctx.author.id
+        # # A variable to store all player's scores for a the trivia quiz session.
+        self.player_scores = {}
+
+    async def play(self) -> None:
+        """Start and handle game."""
+        # Start game if not running.
+        if not self.over:
+            self.game_owner = self.ctx.author
```
ToxicKidz, line 33: In `__init__`, `self.game_owner` is an `int` object, yet when `play` is called, `self.game_owner` is a `Member` object.

ToxicKidz (changes requested):

ChrisLovering (changes requested):
Hi, I started to review the code, but found it harder than I should to follow it, mostly due to the lack of comments. Could you add some useful comments around this new cog that explains why you chose to do things in certain ways?

ChrisLovering:
Also, could you update the reason section of your PR description? This section is supposed to be used to explain why you made the implementation decisions you did. This is not a place to describe the whole reason for the cog, or who approved it.

bot/exts/evergreen/trivia_quiz/_game.py
```diff
@@ -0,0 +1,138 @@
...
+        # A variable to store the questions which have been sent to the user during the quiz.
+        done_question = []
+        # A variable to store the number of hints used in a quiz question.
+        hint_no = 0
+        # A variable to store the answer to a quiz question.
+        answer = None
+
+        while not self.over:
+            # Exit quiz if number of questions for a round are already sent.
+            if len(done_question) > _constants.QUESTION_LIMIT and hint_no == 0:
```
RohanJnr, line 48: I think checking for `hint_no` is not required here. What do you think?
```suggestion
            if len(done_question) > _constants.QUESTION_LIMIT:
```

bot/exts/evergreen/trivia_quiz/_game.py
```diff
@@ -0,0 +1,138 @@
...
+                # if number of hints sent or time alerts sent is less than 2, then send one.
+                if hint_no < _constants.HINTS_LIMIT:
+                    hint_no += 1
+                    if "hints" in question_dict:
+                        hints = question_dict["hints"]
+                        await self.ctx.send(f"**Hint #{hint_no + 1}\n**{hints[hint_no]}")
+                    else:
+                        # If hints are not available for the question, then send time alerts
+                        await self.ctx.send(
+                            f"{(_constants.HINTS_LIMIT * 10) - (hint_no * 10)}s left!"
```
RohanJnr, line 96: This line is causing this issue, with wrong time left being sent.
image


Solution is to increment the `hint_no` value after sending the hint or the time (after line 97).

bot/exts/evergreen/trivia_quiz/_game.py
```diff
@@ -0,0 +1,138 @@
...
+        # Start game if not running.
+        if not self.over:
+            self.game_owner = self.ctx.author.id
+            start_embed = _helpers.make_start_embed(self.category)
+
+            await self.ctx.send(embed=start_embed)  # send an embed with the rules
+            await asyncio.sleep(1)
+
+        # A variable to store the questions which have been sent to the user during the quiz.
+        done_question = []
```
RohanJnr, line 40:
```suggestion
        done_questions = []
```

bot/exts/evergreen/trivia_quiz/_game.py
```diff
@@ -0,0 +1,138 @@
...
+        self.ctx = ctx
+        self.questions = questions
+        self.category = category
+
+        # A variable to store the game status: either running or not running.
+        self.over = False
+        # A variable to store the person's ID who started the quiz game in a channel.
+        self.game_owner = ctx.author.id
+        # # A variable to store all player's scores for a the trivia quiz session.
+        self.player_scores = {}
```
RohanJnr, line 27:
```suggestion
        self.player_scores = defaultdict(int)
```
don't forget to import default dict.

bot/exts/evergreen/trivia_quiz/_game.py
RohanJnr: Also in constants, the `MESSAGE_TIMEOUT` is 15 seconds, but the time left is calculated by taking the value 10 seconds.

bot/exts/evergreen/trivia_quiz/_constants.py
```diff
@@ -0,0 +1,30 @@
...
+MESSAGE_TIMEOUT = 15
+
+
+WRONG_ANS_RESPONSE = (
+    "No one answered correctly!", "Better luck next time"
+)
+WINNER_RESPONSE = (
+    "Congratulations {winners_mention} :tada: "
+    "{winner_word} have won this quiz game with a grand total of {highest_points} points!"
+)
```
RohanJnr, line 13:
```suggestion
WINNER_RESPONSE = (
    "{winners_mention} Congratulations "
    "on winning this quiz game with a grand total of {highest_points} points :tada:"
)
```
This will also get rid of the `word` variable in _helpers.py

bot/exts/evergreen/trivia_quiz/_cog.py
```diff
@@ -0,0 +1,146 @@
...
+                                # giving away the answers, to remove then, they are replaced
+                                # with '[redacted]'.
+                                f"{raw_data['extract'].replace(raw_data['title'], '[redacted]')}",
+                    "answer": raw_data["title"],
+                    "info": raw_data["extract"]
+                }
+                start_id += 1
+                self.wiki_questions.append(formatted_data)
+
+    @commands.group(name="quiz", aliases=["trivia"], invoke_without_command=True)
```
RohanJnr, line 69:
```suggestion
    @commands.group(name="quiz", aliases=("trivia",), invoke_without_command=True)
```

bot/exts/evergreen/trivia_quiz/_cog.py
```diff
@@ -0,0 +1,146 @@
...
+    async def quiz_game(self, ctx: commands.Context, category: str = None) -> None:
+        """
+        Start a quiz!
+
+        Questions for the quiz can be selected from the following categories:
+        - general : Test your general knowledge. (default)
```
RohanJnr, line 75:
```suggestion
        Questions for the quiz can be selected from the following categories:
        A random category will be chosen if provided with none.
        - general : Test your general knowledge.
        - wikipedia/wiki : Guess the title of random wikipedia passages.
```

RohanJnr (changes requested):
Excellent work 💯 ! Just a few changes and bug encounters.

kosayoda (changes requested):
The current game loop is fragile.

- `DEFAULT_QUESTION_LIMIT` is hardcoded to 6 in 4430760c, which seems to be because there are only 7 questions in the `retro` category.
- Using the default question limit bypasses the topic length check
- The game loop only ever exits if the number of done questions go above the question limit
- The current question is retrieved in an infinite loop calling `random.choice` on the available questions and checking if it's done

These points result in the game (and the bot) hanging if the number of available questions for a topic is lower than the hardcoded `DEFAULT_QUESTION_LIMIT`. Since the wikipedia questions are auto-populated (again with an arbitrary hardcoded value of 10), take note of the above scenario when adding the ratelimit to the API.

Ideally, the game loop should be refactored to better retrieve a question for a topic, the default question limit should have a comment describing why it's set to 6, and the default question limit should not bypass the topic questions length check. I don't feel like the fixes for these should go in this PR, so just keep in mind the above quirks for this game so the bot still works for this topic until the fix is made. A separate issue should be created after this PR.

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -420,30 +486,27 @@ def contains_correct_answer(m: discord.Message) -> bool:
...
+
         start_embed = discord.Embed(
-            colour=Colours.blue,
-            title="Quiz game starting!",
+            title="Quiz game Starting!!",
             description=(
-                f"This game consists of {self.question_limit + 1} questions.\n"
-                "**Rules: **No cheating and have fun!\n"
-                f"**Category**: {category}"
+                "Each game consists of 5 questions.\n"
```
RohanJnr, line 503: The number 5 is hard coded but the number of questions are 7 in some cases.

RohanJnr:
image

A few questions have the info attribute set to a blank string in the json file (example: [here](https://github.com/python-discord/sir-lancebot/blob/0a6a355419d9382b35ac651117287d907e98af0c/bot/resources/evergreen/trivia_quiz.json#L186) and [here](https://github.com/python-discord/sir-lancebot/blob/0a6a355419d9382b35ac651117287d907e98af0c/bot/resources/evergreen/trivia_quiz.json#L180))
We might want to remove those or we can tweak the code on line 642
```python
if info is not None:
```
can be changed to 
```python
if info:
```
to filter out empty strings.

wookie184 (approved):
Nice stuff, I'm approving this as it does seem to work, although there are two comments I would like to make.



Firstly, while the randomness of the articles chosen is nice, there are a *lot* of wikipedia articles, to the extent that usually picking a random wikipedia page will give you some obscure topic you have no idea about. For example:
> [redacted] is a village in Nowkand Kola Rural District, in the Central District of Qaem Shahr County, Mazandaran Province, Iran. At the 2006 census, its population was 24, in 10 families.

and as well as this some also seem like they could have multiple answers, e.g.
> [redacted] is a village in Kozhikode district in the state of Kerala, India.

In my testing I couldn't find any questions which I could answer and this makes the game not particularly interesting. A solution to this would be to instead store a list of manually chosen wikipedia pages that can be chosen from.


I also think the current hourly refreshing of questions isn't great. Most days the `.quiz` command will not be used at all, meaning the bot is making 24 unnecessary API requests. As well as this, often when it is used it will be used multiple times in the same hour, meaning you will get repeated questions. 
I think it would have been ok just to make a request per command ran, or if the api ratelimits aren't high enough to do that it should only refresh the list of questions if the current ones have been used. 
Also there is currently a race condition where if you run the `.quiz wikipedia` on startup before `get_wiki_questions` has finished you'll get a KeyError as the questions aren't there yet.

It could be worth opening an issue for these changes once this PR is merged if you don't want to tackle them now.

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -20,46 +21,22 @@
...
-    "weber": ("(unit of magnetic flux)", "m^2*kg*s^-2*A^-1"),
-    "tesla": ("(unit of magnetic flux density)", "kg*s^-2*A^-1"),
-}
+RANDOM_WIKI_URL = "https://en.wikipedia.org/api/rest_v1/page/random/summary"
+TRIVIA_QUIZ_ICON = (
+    "https://raw.githubusercontent.com/python-discord/branding/main/icons/trivia_quiz/trivia-quiz-dist.png"
+)
 
 
 @dataclass(frozen=True)
```
decorator-factory, line 42: 1. Why is `answer` not a set/list of allowed answers?
2. I would remove words like "dataclass" and repeating the class name, which are already clear without the docstring.

decorator-factory (changes requested):

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -578,7 +643,7 @@ async def send_answer(
             description="",
         )
 
-        if info is not None:
+        if info:
```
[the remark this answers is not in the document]
decorator-factory: Oh I see, it's because of https://github.com/python-discord/sir-lancebot/pull/618#issuecomment-857994321
I suppose you could add a comment as to why, or just use `null` instead of an empty string in the questions.

wookie184 (changes requested):
Nice stuff, I think the gameplay with these changes is *much* more fun, better than i'd expected. Just a few things i've commented and things from other reviews (mainly `fuzzywuzzy` -> `rapidfuzz`, should just be just a drop in replacement) and i'll be ready to approve.

Also, on the line with `self.question_limit = DEFAULT_QUESTION_LIMIT`, this is not necessarily correct with the wikipedia cog as there may not be as many questions as `DEFAULT_QUESTION_LIMIT`. I think this line should be changed to `self.question_limit = min(DEFAULT_QUESTION_LIMIT, topic_length)` to prevent the infinite loop when it runs out of questions

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -238,40 +240,56 @@ def cog_unload(self) -> None:
...
+                    # multi letter words are formatted this way: Jason_Terror
+                    title = article["normalizedtitle"].replace("_", " ")
+
+                    if not question:
+                        continue
+                    elif not all(c.isalnum() for c in title):
+                        # If the title contains non alphabet/numerical(s) then
+                        # don't append it to the question and move on to the next.
+                        # Eg. "Jurassic World: Dominion (2022)" would be skipped.
+                        continue
```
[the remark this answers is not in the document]
Shivansh-007: Umm, yeah I will change this to special characters like `~!@#$%^&*(){}[]<>?/:";'` otherwise you get answers as `Jurassic Park (Dominion)`, which would be marked incorrect when you answer it without them.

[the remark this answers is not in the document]
wookie184: shouldn't the fuzzy matching of answers take care of that?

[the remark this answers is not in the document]
Shivansh-007: For some cases it does, for some, it doesn't.

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -238,40 +240,56 @@ def cog_unload(self) -> None:
...
+
+                    for word in question.split(" "):
+                        word = word.translate(str.maketrans('', '', string.punctuation))
+                        if word.lower() in [_word.lower() for _word in title.split(" ")]:
+                            question = question.replace(word, len(word) * "\*")
```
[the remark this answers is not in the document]
wookie184: might want to extract punctuation from `word` before the `re.sub` first like you did I guess. Although I guess if the title was `ice-cream` we'd want to sub out both word separately so maybe the split should take into account punctuation as well.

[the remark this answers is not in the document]
Shivansh-007: I could replace `-` with ` `, like I do for `_`, but replacing all punctuations would space could spoil words I guess?

bot/exts/evergreen/trivia_quiz.py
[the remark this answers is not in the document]
wookie184: I think tweaking the fuzzy match tolerance to make it a bit more forgiving may be a better solution than this, I reckon false matches would be fairly rare so we could probably be relatively forgiving. Thoughts?

bot/exts/evergreen/trivia_quiz.py
Shivansh-007: Doing the first change falls out of the scope of the PR, and requires a bit amount of refactoring so I am not sure if I should include that here.

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -264,7 +348,7 @@ async def quiz_game(self, ctx: commands.Context, category: Optional[str], questi
         topic_length = len(topic)
 
         if questions is None:
-            self.question_limit = DEFAULT_QUESTION_LIMIT
+            self.question_limit = min(DEFAULT_QUESTION_LIMIT, topic_length)
```
wookie184, line 351: This still doesn't protect against the infinite loop, seemingly because of an off-by-one bug throughout this cog. A simple fix for this would probably be just to use `min(DEFAULT_QUESTION_LIMIT, topic_length-1)`, or you could look into this further and try and fix this off-by-one in the rest of the cog (I think the root cause is line 389 `if len(done_questions) > self.question_limit and hint_no == 0:` which should instead be `>=`, as if the number of completed questions is greater than the limit, we've already done too many).

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -202,27 +222,91 @@ def __init__(self, bot: Bot) -> None:
...
+                        continue
+                    elif not all(c.isalnum() for c in title):
+                        # If the title contains non alphabet/numerical(s) then
+                        # don't append it to the question and move on to the next.
+                        # Eg. "Jurassic World: Dominion (2022)" would be skipped.
+                        continue
+
+                    for word in re.split("[\s-]", title):  # noqa: W605
+                        word = word.strip(string.punctuation)
+                        secret_word = "\*"*len(word)  # noqa: W605
```
wookie184, line 280: You should avoid using `noqa`s if you can, as usually they make a good point (as they do here). Here you can use raw strings, e.g. `r"[\s-]"`  to make it clear `\s` isn't an escape sequence meant to be interpreted by python

wookie184 (changes requested):
accidentally posted other comment individually rather than part of a review

bot/exts/evergreen/trivia_quiz.py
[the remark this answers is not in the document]
wookie184: I'm not sure this was resolved. If we think it's hard to guess because of the punctuation we could give the punctuation with the words redacted before the description, this would also help people know the order/length of words, could maybe do that the same as how we do it from the extract
```python
cleaned_title = title
for word in re.split("[\s-]", title):
    word = word.strip(string.punctuation)
    secret_word = "\*"*len(word)
    cleaned_title = re.sub(rf'\b{word}\b', f"**{secret_word}**", cleaned_title, flags=re.IGNORECASE) 
```
Haven't tested this or thought if there might be a cleaner way though

wookie184 (approved):
Thanks for making those changes, this seems good to me now!

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -421,30 +502,29 @@ def contains_correct_answer(m: discord.Message) -> bool:
...
-                    self.question_limit - len(done_question) + 1,
+                    self.question_limit - len(done_questions),
                 )
                 await self.send_score(ctx.channel, self.game_player_scores[ctx.channel.id])
 
                 await asyncio.sleep(2)
 
     def make_start_embed(self, category: str) -> discord.Embed:
         """Generate a starting/introduction embed for the quiz."""
+        rules = "\n".join([f"`{index}`: {rule}" for index, rule in enumerate(RULES, start=1)])
```
janine9vn, line 516: This is minor, but I'm not sure why the rules index is kept in a code block format. It's very odd visually.
image

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -3,190 +3,211 @@
 import logging
 import operator
 import random
+import re
+import string
+import typing as t
```
janine9vn, line 8: This and some of the other typing annotations are going to conflict with the current codebase. You should pull down the recent lancebot and resolve merge conflicts.

bot/exts/evergreen/trivia_quiz.py
```diff
@@ -559,7 +640,7 @@ def category_embed(self) -> discord.Embed:
     @staticmethod
     async def send_answer(
         channel: discord.TextChannel,
-        answers: List[str],
+        answers: t.List[str],
```
janine9vn, line 643: This will conflict with the current codebase and we should use `list` since we're on 3.9.

janine9vn (dismissed):
This PR will conflict with our current code base since we merged in the Type Annotations PR.
You should pull down the recent changes and resolve those conflicts.

Otherwise from my testing the game plays just fine.

Xithrius (approved):
Yes.

Objectivitix (commented):
Impeccable.

PR #619, opened 2021-03-09
Added another alias and fixed a bug in Tic-Tac-Toe

JagTheFriend:
## Fixed bugs in the Tic-Tac-Toe
I changed the code up a bit,
So when the user tries to play the game with an opponent,
The program would first check whether the _opponent_ is a bot or not.
If the bot is a opponent, then game would not start, instead
a message saying: `"You can't play Tic-Tac-Toe with bots!"` would be sent

## Another alias
I also made a new alias called `tic`

### DIFF
--- a/bot/exts/evergreen/tic_tac_toe.py
+++ b/bot/exts/evergreen/tic_tac_toe.py
@@ -10,8 +10,8 @@ from bot.constants import Emojis
 from bot.utils.pagination import LinePaginator
 
 CONFIRMATION_MESSAGE = (
-    "{opponent}, {requester} wants to play Tic-Tac-Toe against you. React to this message with "
-    f"{Emojis.confirmation} to accept or with {Emojis.decline} to decline."
+    "{opponent}, {requester} wants to play Tic-Tac-Toe against you."
+    f"\nReact to this message with {Emojis.confirmation} to accept or with {Emojis.decline} to decline."
 )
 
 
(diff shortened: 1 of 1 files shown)


bot/exts/evergreen/tic_tac_toe.py
```diff
@@ -276,6 +276,9 @@ async def tic_tac_toe(self, ctx: Context, opponent: t.Optional[discord.User]) ->
             )
         self.games.append(game)
         if opponent is not None:
+            if opponent.bot:  # check whether the opponent is a bot or not
+                return await ctx.send("You can't play Tic-Tac-Toe with bots!")
```
ToxicKidz, line 280:
```py
async def tic_tac_toe(self, ctx: Context, opponent: t.Optional[discord.User]) -> None:
```
The return type hint here is None, yet you are returning a Message object.
```suggestion
            if opponent.bot:  # check whether the opponent is a bot or not
                await ctx.send("You can't play Tic-Tac-Toe with bots!")
                return
```

ToxicKidz (changes requested):

bot/exts/evergreen/tic_tac_toe.py
```diff
@@ -84,7 +84,7 @@ async def get_move(self, board: t.Dict[int, str], _: discord.Message) -> t.Tuple
                 board_copy = board.copy()
                 board_copy[move] = symbol
                 if check_win(board_copy):
-                    return False, move
+                    return (False, move)
```
Xithrius, line 87: I don't feel like this change is necessary. If two items are returned at the same time, it will be returned as a tuple of those two items.

JagTheFriend: _i just added to for no real reason :)_

Xithrius: Then I'm removing this change.

Xithrius (approved):

HassanAbouelela (approved):
Thanks for your contribution!

JagTheFriend:
> Thanks for your contribution!

_np x)_

PR #633, opened 2021-03-20
Only output override channels & bot commands channel on whitelist error

### LINKED ISSUE #556: Only recommend one channel in “you can’t use it here” embed
## Description

When using a command in a non-whitelisted channel, something like the following will appear.
image

It would be nice to have a single preferred channel for bot commands, instead of listing all available channels.

## Reasoning

The current embed is very long, and users have received embeds like this one in response as well:
image

It may seem counter intuitive to shoots only one option, but we don’t redirect users to off-topic or voice chat to use bot commands. We do it whenever a person uses the command in a non-whitelisted channel, so redirecting them to an on-topic channel wouldn't make sense. They're only trying to use the bot command, possibly as a part of the current conversation.

The main benefits are that the embed will be more user friendly and avoids the scenario where someone jumps into an on topic channel for bot commands, because the bot told them to.

ChrisLovering:
## Relevant Issues

Closes #556

## Description

Remove all default whitelisted channels from the channel kwarg before building the output string

## Reasoning

Previously this would output all channels, and could result in an error. This change ensures only the main bot channel & and any overridden channels are shown in the embed. We do this before the categories block as the categories kwarg itself is an override, so we want to include those in any output.

## Screenshots

Example where there are no overrides:
image
Example where whitelist includes overrides:
image

## Did you:

### DIFF
--- a/bot/utils/decorators.py
+++ b/bot/utils/decorators.py
@@ -11,7 +11,7 @@ from discord import Colour, Embed
 from discord.ext import commands
 from discord.ext.commands import CheckFailure, Command, Context
 
-from bot.constants import ERROR_REPLIES, Month
+from bot.constants import Channels, ERROR_REPLIES, Month, WHITELISTED_CHANNELS
 from bot.utils import human_months, resolve_current_month
 from bot.utils.checks import in_whitelist_check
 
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
This looks good to me 👍 ~~line 269 is annoying me but it's not part of this PR so that's ok~~

Shivansh-007 (approved):
Looks good! 👌🏻

PureFunctor (approved):
LGTM :+1:

ks129 (approved):

PR #639, opened 2021-03-26
Added 3 More Elements To The Json

NeilShah2026:
Added 3 New Elements To The .Json

PR #647, opened 2021-03-30
Revert "Added 3 More Elements To The Json"

Akarys42:
Reverts python-discord/sir-lancebot#639

The JSON format is invalid.


Akarys42:
Let’s add the commas instead.

PR #659, opened 2021-04-06
Forbid commands in DMs with in_whitelist_check

Akarys42:
This will prevent most commands from running in DMs, which would break most commands and lead to moderation issues.

### DIFF
--- a/bot/utils/checks.py
+++ b/bot/utils/checks.py
@@ -39,6 +39,7 @@ def in_whitelist_check(
     channels: Container[int] = (),
     categories: Container[int] = (),
     roles: Container[int] = (),
+    allow_dms: bool = False,
     redirect: Optional[int] = constants.Channels.community_bot_commands,
     fail_silently: bool = False,
 ) -> bool:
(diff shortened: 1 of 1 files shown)


bot/utils/checks.py
```diff
@@ -39,6 +39,7 @@ def in_whitelist_check(
     channels: Container[int] = (),
     categories: Container[int] = (),
     roles: Container[int] = (),
+    allow_dms: bool = False,
```
ToxicKidz, line 42: I don't think this parameter was used anywhere.

ToxicKidz (changes requested):

bot/utils/checks.py
Akarys42: *play Jedi tricks* you didn't see anything

bot/utils/checks.py
```diff
@@ -66,21 +67,23 @@ def in_whitelist_check(
...
 
-    if channels and ctx.channel.id in channels:
-        log.trace(f"{ctx.author} may use the `{ctx.command.name}` command as they are in a whitelisted channel.")
-        return True
-
-    # Only check the category id if we have a category whitelist and the channel has a `category_id`
-    if categories and hasattr(ctx.channel, "category_id") and ctx.channel.category_id in categories:
-        log.trace(f"{ctx.author} may use the `{ctx.command.name}` command as they are in a whitelisted category.")
-        return True
+    if allow_dms or ctx.guild is not None:
```
ChrisLovering, line 70: Do we need `is not None` here? Could it just be `or ctx.guild`?

Akarys42:
Due to our (probably too deeply) nested checks, these changes shouldn't be solving any of our issues according to sir @HassanAbouelela. Closing this PR.

PR #668, opened 2021-04-08
Add latex cog

### LINKED ISSUE #663: LaTeX Command
## Description
It would be great to have Sir Lancebot spot latex syntax in messages and send an image rendering it.

## Reasoning
There are math-related questions in the off-topic channels fairly frequently, and writing equations in plain text is quite, uh, painful.

## Proposed Implementation
We could use matplotlib for rendering the latex, it seems to do the job pretty well.
Messages with text enclosed in $ symbols would be identified, said text would be rendered, and its image would be sent.

Similar to how the TeXit bot does it:
image

ShakyaMajumdar:
## Relevant Issues
Closes #663 

## Description
Uses matplotlib's default parser for rendering tex.

## Reasoning
matplotlib's subset of tex allows nothing more than symbols and alphanumeric characters, eliminating the need for sandboxing.
It also does not require a local installation of LaTeX.

## Screenshots
image
image

## Did you:

### DIFF
--- a/bot/exts/evergreen/latex.py
+++ b/bot/exts/evergreen/latex.py
@@ -1,3 +1,4 @@
+import re
 from io import BytesIO
 from typing import Union
 
@@ -16,6 +17,16 @@ plt.rcParams.update(
     }
 )
 
+FORMATTED_CODE_REGEX = re.compile(
+    r"(?P<delim>(?P<block>```)|``?)"  # code delimiter: 1-3 backticks; (?P=block) only matches if it's a block
+    r"(?(block)(?:(?P<lang>[a-z]+)\n)?)"  # if we're in a block, match optional language (only letters plus newline)
+    r"(?:[ \t]*\n)*"  # any blank (empty or tabs/spaces only) lines before the code
+    r"(?P<code>.*?)"  # extract all code inside the markup
+    r"\s*"  # any more whitespace before the end of the code markup
+    r"(?P=delim)",  # match the exact same delimiter from the start again
+    re.DOTALL | re.IGNORECASE,  # "." also matches newlines, case insensitive
+)
+
 
 class Latex(commands.Cog):
     """Renders latex."""
@@ -26,7 +37,6 @@ class Latex(commands.Cog):
     @staticmethod
     def _render(text: str) -> Union[BytesIO, str]:
         """Return the rendered image if latex compiles without errors, otherwise return the error message."""
-        text = text.replace(r"\\", "$\n$")  # matplotlib uses \n for newlines, not \\
         fig = plt.figure()
 
         try:
@@ -40,16 +50,26 @@ class Latex(commands.Cog):
         except ValueError as e:
             return str(e)
 
+    @staticmethod
+    def _prepare_input(text: str) -> str:
+        text = text.replace(r"\\", "$\n$")  # matplotlib uses \n for newlines, not \\
+
+        if match := FORMATTED_CODE_REGEX.match(text):
+            return match.group("code")
+        else:
+            return text
+
     @commands.command()
     async def latex(self, ctx: commands.Context, *, text: str) -> None:
         """Renders the text in latex and sends the image."""
+        text = self._prepare_input(text)
         async with ctx.typing():
             image = self._render(text)
 
             if isinstance(image, BytesIO):
                 await ctx.send(file=discord.File(image, "latex.png"))
             else:
-                await ctx.send(image)
+                await ctx.send("```" + image + "```")
 
 
 def setup(bot: commands.Bot) -> None:
--- a/bot/exts/evergreen/latex.py
+++ b/bot/exts/evergreen/latex.py
@@ -1,0 +1,57 @@
+from io import BytesIO
+from typing import Union
+
+import discord
+import matplotlib.pyplot as plt
+from discord.ext import commands
+
+# configure fonts and colors for matplotlib
+plt.rcParams.update(
+    {
+        "font.size": 16,
+        "mathtext.fontset": "cm",  # Computer Modern font set
+        "mathtext.rm": "serif",
+        "figure.facecolor": "38383F",  # matches Discord's dark mode background color
+        "text.color": "white",
+    }
+)
+
+
+class Latex(commands.Cog):
+    """Renders latex."""
+
+    def __init__(self, bot: commands.Bot):
+        self.bot = bot
+
+    @staticmethod
+    def _render(text: str) -> Union[BytesIO, str]:
+        """Return the rendered image if latex compiles without errors, otherwise return the error message."""
+        text = text.replace(r"\\", "$\n$")  # matplotlib uses \n for newlines, not \\
+        fig = plt.figure()
+
+        try:
+            fig.text(0, 1, text, horizontalalignment="left", verticalalignment="top")
+
+            rendered_image = BytesIO()
+            plt.savefig(rendered_image, bbox_inches="tight", dpi=600)
+            rendered_image.seek(0)
+            return rendered_image
+
+        except ValueError as e:
+            return str(e)
+
+    @commands.command()
+    async def latex(self, ctx: commands.Context, *, text: str) -> None:
+        """Renders the text in latex and sends the image."""
+        async with ctx.typing():
+            image = self._render(text)
+
+            if isinstance(image, BytesIO):
+                await ctx.send(file=discord.File(image, "latex.png"))
+            else:
+                await ctx.send(image)
+
+
+def setup(bot: commands.Bot) -> None:
+    """Load the Latex Cog."""
+    bot.add_cog(Latex(bot))
--- a/bot/exts/evergreen/latex.py
+++ b/bot/exts/evergreen/latex.py
@@ -1,7 +1,6 @@
 import asyncio
 import re
 from concurrent.futures import ThreadPoolExecutor
-from functools import partial
 from io import BytesIO
 
 import discord
@@ -66,7 +65,7 @@ class Latex(commands.Cog):
 
             with ThreadPoolExecutor() as pool:
                 image = await asyncio.get_running_loop().run_in_executor(
-                    pool, partial(self._render, text)
+                    pool, self._render, text
                 )
 
             await ctx.send(file=discord.File(image, "latex.png"))
--- a/bot/exts/evergreen/latex.py
+++ b/bot/exts/evergreen/latex.py
@@ -13,19 +13,19 @@ plt.rcParams.update(
         "font.size": 16,
         "mathtext.fontset": "cm",  # Computer Modern font set
         "mathtext.rm": "serif",
-        "figure.facecolor": "38383F",  # matches Discord's dark mode background color
+        "figure.facecolor": "36393F",  # matches Discord's dark mode background color
         "text.color": "white",
     }
 )
 
 FORMATTED_CODE_REGEX = re.compile(
-    r"(?P<delim>(?P<block>```)|``?)"  # code delimiter: 1-3 backticks; (?P=block) only matches if it's a block
-    r"(?(block)(?:(?P<lang>[a-z]+)\n)?)"  # if we're in a block, match optional language (only letters plus newline)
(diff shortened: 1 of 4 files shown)


bot/exts/evergreen/latex.py
```diff
@@ -0,0 +1,77 @@
...
+    r"\s*"  # any more whitespace before the end of the code markup
+    r"(?P=delim)",  # match the exact same delimiter from the start again
+    re.DOTALL | re.IGNORECASE,  # "." also matches newlines, case insensitive
+)
+
+
+class Latex(commands.Cog):
+    """Renders latex."""
+
+    def __init__(self, bot: commands.Bot):
```
anand2312, line 34: We can skip overriding the `__init__` as you don't end up using the Bot instance anywhere

bot/exts/evergreen/latex.py
```diff
@@ -0,0 +1,77 @@
+import re
```
anand2312, line 1: Nitpicking but I think you'll want to rearrange those imports according to alphabetical order as well

bot/exts/evergreen/latex.py
```diff
@@ -0,0 +1,57 @@
...
+    @commands.command()
+    async def latex(self, ctx: commands.Context, *, text: str) -> None:
+        """Renders the text in latex and sends the image."""
+        async with ctx.typing():
+            image = self._render(text)
+
+            if isinstance(image, BytesIO):
+                await ctx.send(file=discord.File(image, "latex.png"))
+            else:
+                await ctx.send(image)
```
PureFunctor: Thanks for the PR! Just a small suggestion:
I think instead of returning a `Union[BytesIO, str]` after a `try-except`, it'd be better if we catch that error within `latex` instead. You could raise a `commands.BadArgument` instead of sending the exception message directly as well. So possibly something like:
```py
    @commands.command()
    async def latex(self, ctx: commands.Context, *, text: str) -> None:
        """Renders the text in latex and sends the image."""
        async with ctx.typing():
            try:
                image = self._render(text)
                await ctx.send(file=discord.File(image, "latex.png"))
            except ValueError as e:
                raise commands.BadArgument("Failed to render.")
```

bot/exts/evergreen/latex.py
ShakyaMajumdar: the imports are pretty much in the order for which flake8 doesn't complain, probably because `import...`s should go before `from ... import ...`s

bot/exts/evergreen/latex.py
ShakyaMajumdar: Done

bot/exts/evergreen/latex.py
ShakyaMajumdar: Alright, once I do that would I have to change anything in the `error_handler.py`?
Also, I think if the rendering fails, the users might want to see where the syntax error is in their latex

PureFunctor: Not entirely sure on the former, but assuming `e` has the information for that, you could interpolate it

ChrisLovering: https://github.com/python-discord/sir-lancebot/blob/964112352809bcec5b13fe5da93b15b4ade1b196/bot/exts/evergreen/error_handler.py#L94 catches BadArguments. The string that you pass when raising will be included in the output, so you could include what you want to output in that there.

bot/exts/evergreen/latex.py
```diff
@@ -64,12 +63,13 @@ async def latex(self, ctx: commands.Context, *, text: str) -> None:
...
         async with ctx.typing():
-            image = self._render(text)
 
-            if isinstance(image, BytesIO):
-                await ctx.send(file=discord.File(image, "latex.png"))
-            else:
-                await ctx.send("```" + image + "```")
+            with ThreadPoolExecutor() as pool:
+                image = await asyncio.get_running_loop().run_in_executor(
+                    pool, partial(self._render, text)
```
ChrisLovering, line 69: This doesn't need to be a partial since text is a normal arg, you'd only need a partial if using kwargs.

bot/exts/evergreen/latex.py
```diff
@@ -48,7 +47,7 @@ def _render(text: str) -> Union[BytesIO, str]:
             return rendered_image
 
         except ValueError as e:
-            return str(e)
+            raise commands.BadArgument(str(e))
```
ChrisLovering, line 50: I'm not sure if we want to output the full traceback here. What does `e` look like in this case?

ChrisLovering (changes requested):

bot/exts/evergreen/latex.py
ShakyaMajumdar: image

ShakyaMajumdar: reasonable imo

ChrisLovering: Yea, that's fine then. Wasn't sure if it would have a huge traceback.

bot/exts/evergreen/latex.py
```diff
@@ -0,0 +1,76 @@
...
+        try:
+            fig.text(0, 1, text, horizontalalignment="left", verticalalignment="top")
+
+            rendered_image = BytesIO()
+            plt.savefig(rendered_image, bbox_inches="tight", dpi=600)
+            rendered_image.seek(0)
+            return rendered_image
+
+        except ValueError as e:
+            raise commands.BadArgument(str(e))
```
ChrisLovering, line 49: Which line is actually going to raise this ValueError? Ideally we would only wrap that individual line in the try/except, rather than all of this.

ChrisLovering (approved):
Just one minor gripe which doesn't need to be done.

Rest looks good to me :)

bot/exts/evergreen/latex.py
ShakyaMajumdar: it was the `plt.savefig`, removed all other lines from the `try`.

ChrisLovering (approved):

bot/exts/evergreen/latex.py
```diff
@@ -0,0 +1,75 @@
...
+
+FORMATTED_CODE_REGEX = re.compile(
+    r"(?P<delim>(?P<block>```)|``?)"  # code delimiter: 1-3 backticks; (?P=block) only matches if it's a block
+    r"(?(block)(?:(?P<lang>[a-z]+)\n)?)"  # if we're in a block, match optional language (only letters plus newline)
+    r"(?:[ \t]*\n)*"  # any blank (empty or tabs/spaces only) lines before the code
+    r"(?P<code>.*?)"  # extract all code inside the markup
+    r"\s*"  # any more whitespace before the end of the code markup
+    r"(?P=delim)",  # match the exact same delimiter from the start again
+    re.DOTALL | re.IGNORECASE,  # "." also matches newlines, case insensitive
+)
```
Shivansh-007, line 29:
```suggestion
FORMATTED_CODE_REGEX = re.compile(
    r"(?P<delim>(?P<block>```)|``?)"        # code delimiter: 1-3 backticks; (?P=block) only matches if it's a block
    r"(?(block)(?:(?P<lang>[a-z]+)\n)?)"    # if we're in a block, match optional language (only letters plus newline)
    r"(?:[ \t]*\n)*"                        # any blank (empty or tabs/spaces only) lines before the code
    r"(?P<code>.*?)"                        # extract all code inside the markup
    r"\s*"                                  # any more whitespace before the end of the code markup
    r"(?P=delim)",                          # match the exact same delimiter from the start again
    re.DOTALL | re.IGNORECASE               # "." also matches newlines, case insensitive
)
```

bot/exts/evergreen/latex.py
```diff
@@ -0,0 +1,75 @@
...
+import matplotlib.pyplot as plt
+from discord.ext import commands
+
+# configure fonts and colors for matplotlib
+plt.rcParams.update(
+    {
+        "font.size": 16,
+        "mathtext.fontset": "cm",  # Computer Modern font set
+        "mathtext.rm": "serif",
+        "figure.facecolor": "38383F",  # matches Discord's dark mode background color
```
Shivansh-007, line 16:
```suggestion
        "figure.facecolor": "36393F",  # matches Discord's dark mode background color
```

Nitpicking, but this is the exact colour code for dark mode, can be found in `discord.colour.Colour.dark_theme`

bot/exts/evergreen/latex.py
```diff
@@ -0,0 +1,75 @@
...
+        """Renders the text in latex and sends the image."""
+        text = self._prepare_input(text)
+        async with ctx.typing():
+
+            with ThreadPoolExecutor() as pool:
+                image = await asyncio.get_running_loop().run_in_executor(
+                    pool, self._render, text
+                )
+
+            await ctx.send(file=discord.File(image, "latex.png"))
```
Shivansh-007, line 70: You would need to `import time` to run this code.

```suggestion
        text = self._prepare_input(text)
        async with ctx.typing():
            with ThreadPoolExecutor() as pool:
                start = time.time()
                image = await asyncio.get_running_loop().run_in_executor(
                    pool, self._render, text
                )
                end = time.time()

                time_taken = end - start

            embed = discord.Embed(name="Latex Output")
            embed.set_footer(text=f"Rendering the latex image took {time_taken}s.")
            file = discord.File(image, "latex.png")
            embed.set_image(url='attachment://latex.png')

            await ctx.send(embed=embed, file=file)
```

Was messing up with the output message, and I think I came up with something good, here is the result:

### Dark Mode
Screenshot from 2021-04-10 07-19-02


### Light Mode
Screenshot from 2021-04-10 07-19-54

Shivansh-007 (changes requested):

bot/exts/evergreen/latex.py
ShakyaMajumdar: Per discussion in #dev-contrib I'm not implementing this

jb3 (dismissed):
This all looks fine and works great locally, however discussion in #dev-contrib pointed towards this task taking multiple seconds, so I'm wondering whether some sort of caching should be implemented.

My suggestion would be to generate a hash of some sort of the query and then store to a `_latex_cache` directory in the current folder or something (therefore allowing it to work in Docker & on Host). The directory would not be persistent at all and should of course be gitignored.

What's the maximum observed time that one of these generations takes, and what is the rate at which that increases? We should try take steps to ensure this feature isn't open to abuse.

bot/exts/evergreen/latex.py
```diff
@@ -69,6 +80,9 @@ async def latex(self, ctx: commands.Context, *, text: str) -> None:
 
             await ctx.send(file=discord.File(image, "latex.png"))
 
+            with open(image_path, "wb") as f:
+                f.write(image.getbuffer())
```
ChrisLovering, line 84: Could we move this into the `_render` function, since it's blocking? Could be worth renaming the function or updating the docstring after too, to reflect it also saves the file.

bot/exts/evergreen/latex.py
```diff
@@ -28,6 +30,10 @@
     re.DOTALL | re.IGNORECASE,              # "." also matches newlines, case insensitive
 )
 
+CACHE_DIRECTORY = pathlib.Path("_latex_cache")
+if not CACHE_DIRECTORY.exists():
+    CACHE_DIRECTORY.mkdir()
```
ChrisLovering, line 35: This could just be `CACHE_DIRECTORY.mkdir(exist_ok=True)`

bot/exts/evergreen/latex.py
```diff
@@ -60,7 +66,12 @@ def _prepare_input(text: str) -> str:
     async def latex(self, ctx: commands.Context, *, text: str) -> None:
```
ChrisLovering, line 66: Following the discussion in #dev-contrib I think we want to add a [max_concurrency](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.max_concurrency) to this command. I'd suggest with 1 concurrent, the guild bucket, and wait=True.

ChrisLovering (changes requested):

ChrisLovering (approved):

ks129 (approved):

Xithrius (approved):

PR #672, opened 2021-04-10
add timed command for timed execution of commands

### LINKED ISSUE #671: Feature: .timed <command> command to allow any command's execution time to be displayed
## Description
Quite simply, the ability to use `.timed <command + args>` (.t as an alias), to be able to time the execution time of a command.

## Reasoning
Considering a conversation in dev-contrib about having time displayed in a latex rendering command, I figured why not have the best of both worlds; by default don't display the time since it can be considered bloat, but have the ability to time the execution of a command should you wish for the extra technical details, if it's of interest to you.

## Proposed Implementation
Implementation would be done by invoking the command from the timed command, but argument and check validation would need to be considered, since by default ctx.invoke bypasses command checks.

An example of the usage of this might be `.t 8bitify` which is a long running command, and it would show you exactly how long running it was.

vcokltfre:
## Relevant Issues
Implements and closes #671 

## Description
The changes have been implemented by creating a new execution context for the command by taking the message content, updating a copy of the message with that content, and using `bot.get_context()` to get a new context for execution. The invokation of that context is then timed by perf_counter() and displayed to the user upon the command finishing with the time in seconds to .4f precision.

## Reasoning
It's been implemented in this way so as to be able to display the time without modifying the original context,

## Screenshots
A normal timed execution:
image
An attempt at timed execution where command permissions were denied (temporarily role-gated 8bitify command):
image
An attempt at timed execution where the command is on cooldown (temporarily cooldowned 8bitify command):
image
Passing additional command arguments:
image

## Did you:

### DIFF
--- a/bot/exts/evergreen/timed.py
+++ b/bot/exts/evergreen/timed.py
@@ -16,7 +16,7 @@ class TimedCommands(commands.Cog):
 
         return await ctx.bot.get_context(msg)
 
-    @commands.command(name="timed", aliases=["t"])
+    @commands.command(name="timed", aliases=["time", "t"])
     async def timed(self, ctx: commands.Context, *, command: str) -> None:
         """Time the command execution of a command."""
         new_ctx = await self.create_execution_context(ctx, command)
--- a/bot/exts/evergreen/timed.py
+++ b/bot/exts/evergreen/timed.py
@@ -1,0 +1,36 @@
+from copy import copy
+from time import perf_counter
+
+from discord import Message
+from discord.ext import commands
+
+
+class TimedCommands(commands.Cog):
+    """Time the command execution of a command."""
+
+    def __init__(self, bot: commands.Bot) -> None:
+        self.bot = bot
+
+    @staticmethod
+    async def create_execution_context(ctx: commands.Context, command: str) -> commands.Context:
+        """Get a new execution context for a command."""
+        msg: Message = copy(ctx.message)
+        msg._update({"content": f"{ctx.prefix}{command}"})
+
+        return await ctx.bot.get_context(msg)
+
+    @commands.command(name="timed", aliases=["t"])
+    async def timed(self, ctx: commands.Context, *, command: str) -> None:
+        """Time the command execution of a command."""
+        new_ctx = await self.create_execution_context(ctx, command)
+
+        t_start = perf_counter()
+        await new_ctx.command.invoke(new_ctx)
+        t_end = perf_counter()
+
+        await ctx.send(f"Command execution for `{new_ctx.command}` finished in {(t_end - t_start):.4f} seconds.")
+
+
+def setup(bot: commands.Bot) -> None:
+    """Cog load."""
+    bot.add_cog(TimedCommands(bot))
--- a/bot/exts/evergreen/timed.py
+++ b/bot/exts/evergreen/timed.py
@@ -8,9 +8,6 @@ from discord.ext import commands
 class TimedCommands(commands.Cog):
     """Time the command execution of a command."""
 
-    def __init__(self, bot: commands.Bot) -> None:
-        self.bot = bot
-
     @staticmethod
     async def create_execution_context(ctx: commands.Context, command: str) -> commands.Context:
         """Get a new execution context for a command."""
@@ -23,6 +20,9 @@ class TimedCommands(commands.Cog):
     async def timed(self, ctx: commands.Context, *, command: str) -> None:
         """Time the command execution of a command."""
         new_ctx = await self.create_execution_context(ctx, command)
+
+        if new_ctx.command and new_ctx.command.qualified_name == "timed":
+            return await ctx.send("You are not allowed to time the execution of the `timed` command.")
 
         t_start = perf_counter()
         await new_ctx.command.invoke(new_ctx)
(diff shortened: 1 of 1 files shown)


bot/exts/evergreen/timed.py
```diff
@@ -0,0 +1,36 @@
...
+    """Time the command execution of a command."""
+
+    def __init__(self, bot: commands.Bot) -> None:
+        self.bot = bot
+
+    @staticmethod
+    async def create_execution_context(ctx: commands.Context, command: str) -> commands.Context:
+        """Get a new execution context for a command."""
+        msg: Message = copy(ctx.message)
+        msg._update({"content": f"{ctx.prefix}{command}"})
```
ToxicKidz, line 18:
```suggestion
        msg.content = f"{ctx.prefix}{command}"

**Review by ToxicKidz — CHANGES_REQUESTED** [review]

**Review by Shivansh-007 — CHANGES_REQUESTED** [review]
### Problem
Currently a user can run as many `time` commands as he wants, which would then lead to spamming in the channel, i.e.

Screenshot from 2021-04-10 23-34-33

### Solution
add `commands.max_concurrency(2)` decorator would solve this issue by allowing only two `time` calls.

Screenshot from 2021-04-10 23-45-34

**Comment by Akarys42** [comment]
The problem pointed out by @Shivansh-007 (https://github.com/python-discord/sir-lancebot/pull/672#pullrequestreview-632942620) should be fixed, although I disagree with the proposed solution. Using this wouldn't allow more than two users to use the command at the same time which doesn't make sense. The `timed` command should just make sure that it isn't executing itself.

**Comment by ToxicKidz** [comment]
> The problem pointed out by @Shivansh-007 ([#672 (review)](https://github.com/python-discord/sir-lancebot/pull/672#pullrequestreview-632942620)) should be fixed, although I disagree with the proposed solution. Using this wouldn't allow more than two users to use the command at the same time which doesn't make sense. The `timed` command should just make sure that it isn't executing itself.

I think that's only because it's using the [global bucket type](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.discord.ext.commands.BucketType.default) instead of the [user bucket type](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.discord.ext.commands.BucketType.user).

**Comment by Akarys42** [comment]
Cool, let's use the user bucket then and call it a day.

File: bot/exts/evergreen/timed.py (hunk 1 above)
Comment (vcokltfre) [reply bot/exts/evergreen/timed.py:18]: discord.py doesn't appreciate when we directly set attributes of things like messages, which is why i'm using the update function made specifically for that

**Comment by vcokltfre** [comment]
A better solution would just be to check that new_ctx.command is not `timed` which is my intended solution to the problem

File: bot/exts/evergreen/timed.py (hunk 1 above)
Comment (ToxicKidz) [reply bot/exts/evergreen/timed.py:15]: I don't like that we're using a private method here, plus, the only thing `_update` does is set `message.content = value`. So this is the same thing here.

**Comment by vcokltfre** [comment]
image

File: bot/exts/evergreen/timed.py [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+
+        return await ctx.bot.get_context(msg)
+
+    @commands.command(name="timed", aliases=["t"])
+    async def timed(self, ctx: commands.Context, *, command: str) -> None:
+        """Time the command execution of a command."""
+        new_ctx = await self.create_execution_context(ctx, command)
+
+        if new_ctx.command and new_ctx.command.qualified_name == "timed":
+            return await ctx.send("You are not allowed to time the execution of the `timed` command.")
```
ToxicKidz, line 25: The return annotation is `None`, but you're returning a `Message` object.
```suggestion
            await ctx.send("You are not allowed to time the execution of the `timed` command.")
            return
```

ToxicKidz (changes requested):

ToxicKidz:
Alright, this command works pretty nicely. However, when the command ***isn't*** found, there isn't any handling for an `AttributeError`:
```py
  File "sir-lancebot\bot\exts\evergreen\timed.py", line 29, in timed
    await new_ctx.command.invoke(new_ctx)
AttributeError: 'NoneType' object has no attribute 'invoke'
```

ToxicKidz (approved):
Thanks. LGTM. :shipit:

ChrisLovering (approved):
Looks good to me. Only limitation is that it doesn't work with commands that use `wait_for`s, such as anything using the `LinePaginator`, as the invoked command doesn't "finish" until the `wait_for` ends.

bot/exts/evergreen/timed.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @staticmethod
+    async def create_execution_context(ctx: commands.Context, command: str) -> commands.Context:
+        """Get a new execution context for a command."""
+        msg: Message = copy(ctx.message)
+        msg.content = f"{ctx.prefix}{command}"
+
+        return await ctx.bot.get_context(msg)
+
+    @commands.command(name="timed", aliases=["t"])
```
Akarys42, line 19: It could be worth adding a `time` alias, but I don't mind either way.
```suggestion
    @commands.command(name="timed", aliases=["time", "t"])
```

Akarys42 (approved):
LGTM + an alias request

Shivansh-007 (approved):
Thanks! Looks Good To Me! :shipit:

PR #679, opened 2021-04-12
fix: display help for the correct command when an error occurs in timed

vcokltfre:
## Description
When a new execution context is created in the timed command it's now added as the `subcontext` attribute of the parent context. When the error handler deals with the error the context is set to the subcontext if available, else the normal context. This means the command displays correctly.

## Reasoning
It's been implemented this way since I believe it to be the cleanest and simplest way to do this. If this turns out not to be the case I'm more than happy to change the implementation to something more suitable.

## Screenshots
Before:
image
After:
image

## Did you:

### DIFF
--- a/bot/exts/evergreen/timed.py
+++ b/bot/exts/evergreen/timed.py
@@ -21,7 +21,9 @@ class TimedCommands(commands.Cog):
         """Time the command execution of a command."""
         new_ctx = await self.create_execution_context(ctx, command)
 
-        if not new_ctx.command:
+        ctx.subcontext = new_ctx
+
+        if not ctx.subcontext.command:
             help_command = f"{ctx.prefix}help"
             error = f"The command you are trying to time doesn't exist. Use `{help_command}` for a list of commands."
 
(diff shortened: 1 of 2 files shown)


ToxicKidz (approved):
:shipit: :shipit: :shipit:

ChrisLovering (approved):
Looks good to me, tested locally.

Xithrius (commented):
Is there a way of having the bot send something like `.timed uwu <text>`?

vcokltfre:
I'll see what I can do, but I get the feeling the solution may end up being quite ugly, anyway, will see :P

bot/exts/evergreen/timed.py
```diff
@@ -21,6 +21,8 @@ async def timed(self, ctx: commands.Context, *, command: str) -> None:
         """Time the command execution of a command."""
         new_ctx = await self.create_execution_context(ctx, command)
 
+        ctx.subcontext = new_ctx
+
         if not new_ctx.command:
```
Xithrius, line 26: Why don't we do this?
```suggestion
        if not ctx.subcontext.command:
```

Xithrius (approved):
Yes.

PR #682, opened 2021-04-13
Fix Latex Cog Bug

ShakyaMajumdar:
## Description
Switched from `matplotlib.pyplot.figure` to `matplotlib.figure.Figure`.
Memory cleared manually with `del` and `gc.collect()`.
Cooldown added to the `.latex` command

## Reasoning
`matplotlib.pyplot.figure` isn't properly reference counted.
With these changes in place, memory usage doesn't increase without bound, plateaus out at ~110MB.

## Did you:

### DIFF
--- a/bot/exts/evergreen/latex.py
+++ b/bot/exts/evergreen/latex.py
@@ -8,7 +8,6 @@ from io import BytesIO
 
 import discord
 import matplotlib.pyplot as plt
-from matplotlib import figure
 from discord.ext import commands
 from matplotlib import figure
 
--- a/bot/exts/evergreen/latex.py
+++ b/bot/exts/evergreen/latex.py
@@ -1,4 +1,5 @@
 import asyncio
+import gc
 import hashlib
 import pathlib
 import re
@@ -8,6 +9,7 @@ from io import BytesIO
 import discord
 import matplotlib.pyplot as plt
 from discord.ext import commands
+from matplotlib import figure
 
 # configure fonts and colors for matplotlib
 plt.rcParams.update(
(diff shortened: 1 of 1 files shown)


bot/exts/evergreen/latex.py
```diff
@@ -44,14 +46,18 @@ def _render(text: str, filepath: pathlib.Path) -> BytesIO:
...
 
         try:
-            plt.savefig(rendered_image, bbox_inches="tight", dpi=600)
+            fig.savefig(rendered_image, bbox_inches="tight", dpi=600)
         except ValueError as e:
             raise commands.BadArgument(str(e))
+        finally:
+            fig.clf()
+            del fig
+            gc.collect()
```
ChrisLovering, line 60: I don't think we should be manually messing with memory, especially the global python GC. Scaleios is working on a mutliprocessing version of this cog, so that we can limit the memory of the spawned process.

This can be removed until that is ready.

bot/exts/evergreen/latex.py
```diff
@@ -71,6 +77,7 @@ def _prepare_input(text: str) -> str:
 
     @commands.command()
     @commands.max_concurrency(1, commands.BucketType.guild, wait=True)
+    @commands.cooldown(1, 5, commands.BucketType.guild)
```
ChrisLovering, line 80: I'm not sure this is needed, since the issue isn't with too many people using it quickly, but 1 bad actor giving it really long input. I don't think this cooldown is needed, but I'm not completely against it if someone else disagrees.

ChrisLovering (changes requested):

bot/exts/evergreen/latex.py
onerandomusername: > 
> 
> I'm not sure this is needed, since the issue isn't with too many people using it quickly, but 1 bad actor giving it really long input. I don't think this cooldown is needed, but I'm not completely against it if someone else disagrees.

I think in addition to a cooldown a limit of how many times the command can be executing at once would also help.

ChrisLovering: We're already limiting how many times the command can be executing at once with the `max_concurrency` decorator. That's my point, I don't think the cooldown adds anything on top of that.

kosayoda:
@ShakyaMajumdar Hello, are you planning on continuing work on this PR?

ShakyaMajumdar:
@kosayoda Sorry for the late, late response, but this PR wasn't going anywhere. I'm going to look into alternatives to matplotlib altogether, and open up new issues/PRs when I get there.

kosayoda:
No problem

PR #683, opened 2021-04-15
Updates Sir-lancebot To Python3.9.4

### LINKED ISSUE #549: Investigate an upgrade to Python 3.9.4
## Description
Python 3.9.1 has been released today (yay!), which means that it's mature enough for us to start looking into upgrading the Python version of sir-lancebot. We'll obviously have to make sure all our dependencies support 3.9 and everything works well, but it would be nice to use the new language features.

## Reasoning
It's better to keep up with Python version, adjusting our code base as required, than making a major leap in major versions with a potential more sizeable migration. Also, Python 3.9 introduces those neat built-in type generics, meaning that those ugly `typing.List[Typing.Dict[str, str]]` annotations can be replaced by `list[dict[str, str]]`. 

## Additional Details
We should not rush the upgrade! Take your time and have a look at our dependencies: Do they support Python 3.9? Are there know issues that are relevant for us? If too man roadblocks appear on the horizon, we can simply wait until they're solved!

Shivansh-007:
Closes #549 

This PR upgrades the Python version we use for the bot to Python 3.9.4. One of the major upgrades this offers for us is that we can now use `list[dict[str, str]]` rather than importing `typing` module and then doing those ugly `typing.List[Typing.Dict[str, str]]` annotations.

## Changes to be aware of

- Bumps Python to 3.9 in Pipfile & lock and Docker.
- Bumps all dependencies to the latest version, which support `python3.9` and solve the **breaking** changes only
- Adds super class call to `LinePaginator`, discord.py adds `linesep` argument to the `Paginator` in v1.7.0.
- `typing` module has deprecated `Iterator`, `Mapping`, `Callable`, `Container` in python3.9, and they have to be imported from `collections.abc` rather now.
- `asyncio.ensure_future` has been changed to `asyncio.create_task` as it​ is a higher level function and is more readable, and preferred over `​asyncio.ensure_future`​

- `loop.run_in_executor` has been changed to `asyncio.to_thread`. It is mainly used for running IO-bound functions in a separate thread to avoid blocking the event loop, and essentially works as a high-level version of `run_in_executor()` that can directly take keyword arguments.

## Dependency Tracker
[REMOVED] All dependencies have been tested and updated by myself and @ToxicKidz

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # bot (project-specific)
 log/*
 data/*
-
+_latex_cache/*
 
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8-slim
+FROM python:3.9-slim
 
 # Set pip to have cleaner logs and no saved cache
 ENV PIP_NO_CACHE_DIR=false \
--- a/Pipfile
+++ b/Pipfile
@@ -15,6 +15,7 @@ PyYAML = "~=5.4"
 "discord.py" = {extras = ["voice"], version = "~=1.5.1"}
 async-rediscache = {extras = ["fakeredis"], version = "~=0.1.4"}
 emojis = "~=0.6.0"
+matplotlib = "~=3.4.1"
 
 [dev-packages]
 flake8 = "~=3.8"
--- a/Pipfile
+++ b/Pipfile
@@ -4,33 +4,33 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-aiodns = "~=2.0"
-arrow = "~=0.14"
-beautifulsoup4 = "~=4.8"
-fuzzywuzzy = "~=0.17"
-pillow = "~=8.1"
-pytz = "~=2019.2"
-sentry-sdk = "~=0.19"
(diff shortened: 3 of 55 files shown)


Shivansh-007:
Merged `python-discord/sir-lancebot:main` into the PR branch to make the PR up to date with all the changes, the merge introduces no breaking change.

HassanAbouelela (changes requested):
Preliminary block on this PR (I have not reviewed the actual content yet). This PR can not be merged yet, for the same reason our other projects can not be migrated. A lot of the dependencies you've bumped do not have wheels for 3.9 yet, which means these changes are not acceptable through no fault of your own. Without wheels, you need build tools, which significantly hurts people contributing on systems that don't have them by default (such as windows), and will not work in our CI. You can test the state of these current deps using the docker image.

Shivansh-007:
> A lot of the dependencies you've bumped do not have wheels for 3.9 yet

This is not entirely correct, I tested it on docker and with pipenv, the only dependency which doesn't have python3.9 wheels is [`pycares`](https://github.com/saghul/pycares).

Xithrius:
Way too many conflicts. Best to make a new PR.

PR #704, opened 2021-04-23
Suppress Links In Certain Commands

### LINKED ISSUE #703: Catify Can Embed Images
## Description

Commands that echo user input have the vulnerability of allowing embedded images where they otherwise should not be allowed. It manifested today multiple times through the catify command, but has implications for other commands such as `uwu` and `randomcase`.

## Steps to Reproduce

1. Use the catify command with a CDN link.
2. The bot should embed the link (even if the user can't).

## Expected Behaviour

1. The link is escaped.

## Actual Behaviour

The image is embedded, even when it is not supposed to.

## Possible Solutions

I'm making a PR immediately after this report which escapes links.

## Additional Details

Why has this not manifested before? The `uwu` command messes discord CDN links by "uwufying" them, and the `randomcase` messes them up by changing the case (technically, it's possible for randomcase to echo back a completely unchanged link, which would open up the same vulnerability, but it doesn't seem to have happened yet).

Here is a full list of vulnerable commands:
- catify
- uwu
- randomcase
- eggdecorate (with invalid input)

There could be more, but these are the ones I could find just from the help menu + ones that are actually used regularly.

Not including embeds, there are 256 instances of the send function being invoked, and any are susceptible. This fix is a good **mitigation**, but it's also worth considering doing something like this in a more global manner to avoid it in the future.

HassanAbouelela:
## Relevant Issues

Closes #703

## Description

I've added a helper that does a regex search through a given string, looks for `http` or `https`, and surrounds that with `<>`. I've called it in places I could find where the bot tries to echo part of, or the full user input back to the user.

## Reasoning

Regex seemed like the most simple way to deal with finding links. This solution does not take into account that an already escaped link will end up with weird extra brackets around it, but that didn't seem to matter much.

## Did you:

### DIFF
--- a/bot/exts/easter/egg_decorating.py
+++ b/bot/exts/easter/egg_decorating.py
@@ -10,6 +10,8 @@ import discord
 from PIL import Image
 from discord.ext import commands
 
+from bot.utils import helpers
+
 log = logging.getLogger(__name__)
 
 with open(Path("bot/resources/evergreen/html_colours.json"), encoding="utf8") as f:
@@ -65,7 +67,7 @@ class EggDecorating(commands.Cog):
             if value:
                 colours[idx] = discord.Colour(value)
             else:
-                invalid.append(colour)
+                invalid.append(helpers.suppress_links(colour))
 
         if len(invalid) > 1:
             return await ctx.send(f"Sorry, I don't know these colours: {' '.join(invalid)}")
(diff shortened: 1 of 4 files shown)


Xithrius (approved):
It works well.

ToxicKidz:
LGTM. :shipit:

ToxicKidz (approved):

PR #727, opened 2021-05-09
Reddit Revoke Endpoint Status Code

HassanAbouelela:
## Relevant Issue
[Sentry error, and me dissecting it](https://discord.com/channels/267624335836053506/820712029245538314/841055222910418984).

## Description
The reddit cog has a util to revoke a token. The util checks the status code of the revoke endpoint for code 204.

The reddit [docs](https://github.com/reddit-archive/reddit/wiki/OAuth2#manually-revoking-a-token) make the following claim:
> Note: Per [RFC 7009](https://datatracker.ietf.org/doc/html/rfc7009), this request will return a success (204) response even if the passed in token was never valid.

That claim is blatantly incorrect. First, the standard specifies 200 as the success code, as seen on [page 4, section 2.2](https://datatracker.ietf.org/doc/html/rfc7009#section-2.2). Secondly, the endpoint is easy enough to test, and the return code is in fact, 200.

(If you'd like to test the endpoint, make a POST to `https://www.reddit.com/api/v1/revoke_token`, with multipart form body, with the key token, and any value, with basic auth using your app's username and token)

This PR switches the check to use 200 as it should and leaves 204 in case that is in reddit's future plans, though I doubt it.

## Did you:


vcokltfre (approved):
Yep.

ToxicKidz (approved):
:shipit:

Xithrius (approved):
Yes.

PR #728, opened 2021-05-10
Add "half" command 

devharry20:
I have created a command called "half" which will display half of a members avatar.

PR #729, opened 2021-05-10
New half command created

devharry20:
I have created a command which will display a cropped version of a members avatar.


HassanAbouelela:
I don't understand why you closed #728 then opened this one? The same things that were said there apply here.

devharry20:
> 
> 
> I don't understand why you closed #728 then opened this one? The same things that were said there apply here.

Ah shit yeah I didn't see the comment

PR #744, opened 2021-05-17
Remove superfluous dependencies

ChrisLovering:
## Description

1. Change AoC helpers to use arrow over pytz 
2. Remove bs4 and pytz 
3. Update arrow to 1.1.0

pytz is only used in one file in the whole bot, so we can remove the dependency by changing AoC to use arrow.  
I chose to use `America/Chicago` arbitrarily, for no other reason that being the first tz google returned that was in EST.

Beautifulsoup4 is no longer used, it used to be used in AoC code a while ago.

I also updated arrow to 1.1.0 while I was here as the breaking changes don't impact us.

## Did you:


ToxicKidz (changes requested):
I get this error when I run `.aoc countdown`
```py
05/17/21 18:41:12 - bot.exts.evergreen.error_handler ERROR: Unhandled command error: tzinfo argument must be None or of a tzinfo subclass, not type 'str'
  File "python3.9\lib\site-packages\discord\ext\commands\core.py", line 85, in wrapped 
  File "~\sir-lancebot\bot\exts\christmas\advent_of_code\_cog.py", line 103, in aoc_countdown
    datetime_now = datetime.now(_helpers.EST)
TypeError: tzinfo argument must be None or of a tzinfo subclass, not type 'str'
```

ToxicKidz (approved):
:shipit:

jb3:
> I chose to use America/Chicago arbitrarily, for no other reason that being the first tz google returned that was in EST.

Will this be consistent? Arrow will presumably implement daylight savings for actual locations.

bot/exts/christmas/advent_of_code/_helpers.py
```diff
@@ -48,7 +48,7 @@
 )
 
 # Create an easy constant for the EST timezone
-EST = pytz.timezone("EST")
+EST = "America/Chicago"
```
Shivansh-007, line 51: According to @janine9vn 
> Chicago is CST/CDT. So one hour before EST/EDT

[New York](https://www.zeitverschiebung.net/en/timezone/america--new_york), should be the right timezone.

ChrisLovering:
> > I chose to use America/Chicago arbitrarily, for no other reason that being the first tz google returned that was in EST.
> 
> Will this be consistent? Arrow will presumably implement daylight savings for actual locations.

Arrow will shift for daylight savings, which I assume is what we want, rather than a static GMT offset.

Shivansh-007 (approved):
Looks Good To Me! :shipit:

jb3 (approved):

Issue #705, opened 2021-04-23
Preserve The State Of Unloaded Cogs


HassanAbouelela:
## Description

Unloaded cogs should be saved in redis, and should not be loaded after restarts.

## Reasoning

From time to time, we may end up with problematic code that we choose to unload. This can happen for many reasons, from slight oversights to huge breaking bugs. The main point is, when we unload something, we usually don't want it loaded again until the problem is addressed.

Unloading right now works well enough, but it does not persist across bot restarts. There is no good way to force users to unload cogs whenever the bot restarts, so we should implement a cache on unloaded cogs.

I've been told this had been considered before, but not actioned because lancebot did not have a cache. It does now, this feature seems like a pretty important addition.

## Proposed Implementation

An identifier, such as cog name should be saved in a redis cache upon unloading. During restarts, each module should be looked for in the cache, and should not be loaded if found.

Certain edge cases may occur if a module changes names while unloaded, and in such cases, it may be best to ignore it in code, and send a warning in a dev channel.

## Would you like to implement this yourself?

- [ ] I'd like to implement this feature myself
- [x] Anyone can implement this feature

Shivansh-007:
Cool, I will go ahead and implement this, thanks!

HassanAbouelela:
It's not approved yet, I want discussion from people who rejected it previously.

kwzrd:
> Certain edge cases may occur if a module changes names while unloaded, and in such cases, it may be best to ignore it in code, and send a warning in a dev channel.

Do you have any ideas for how this could work?

The bot could walk modules in the cache and verify membership in the [extensions](https://github.com/python-discord/sir-lancebot/blob/d02c79e569092dbd6823f2edb6c78ed8ab1bcae4/bot/utils/extensions.py#L44) constant. This would mean looking into the cache before the bot connects, which is probably ok, as by the time the bot instance exists, we should already [be connected to Redis and have a loop](https://github.com/python-discord/sir-lancebot/blob/d02c79e569092dbd6823f2edb6c78ed8ab1bcae4/bot/bot.py#L161).

[...]

PR #745, opened 2021-05-18
Add logic for preserving unloaded Cogs.

Shivansh-007:
This PR makes a Redis cache in `bot.bot` to store all the unloaded extensions, on every unload and load this cache would be edited in the desired manner, either removing it from it or adding it in the cache.

Now, coming to not loading the unloaded commands, on bot startup, just after Redis gets connected, the bot would run `bot.get_unloaded_extensions` and get all the unloaded extensions from the cache. These extensions would be skipped and not loaded, the skipping message would be logged. Once all these extensions are loaded, we compare and check which
extensions are present in the cache and not the bot. This means these cogs are either removed or renamed.

A message containing all these "extra" cogs are sent to #dev-alerts and these cogs are removed from the cache.

Resolves: #705

##### 1: Cache containing `'bot.exts.evergreen.catify | bot.exts.evergreen.xkd'`
Screenshot from 2021-05-18 15-37-50

##### 2: After bot sends ^
Screenshot from 2021-05-18 15-43-34

##### 3: On doing `.exts unload ping`
Screenshot from 2021-05-18 15-44-57

##### 4: On doing `.exts load ping`
Screenshot from 2021-05-18 15-57-27

### Todo
- [x] There is a small bug in this, I will fix that soon, since the bot removes the cogs from the cache first and then performs the action, it will also remove the cogs which have been failed to auction.
- [ ] Remove testing code
- [ ] Add core-developers role to allowed mention, need to discuss if it should be allowed overall or just while the alert message is sent.


bot/bot.py
```diff
@@ -61,8 +76,14 @@ def add_cog(self, cog: commands.Cog) -> None:
...
-        super().add_cog(cog)
-        log.info(f"Cog loaded: {cog.qualified_name}")
+        cog_name = inspect.getmodule(cog).__name__
+
+        if cog_name in self.unloaded_extensions:
+            self.unloaded_extensions.remove(cog_name)
+            log.info(f"Skipping cog {cog.qualified_name}, found in unloaded cache.")
+        else:
+            super().add_cog(cog)
+            log.info(f"Cog loaded: {cog.qualified_name}")
```
ToxicKidz, line 86: Maybe we can just override `load_extension` here

bot/constants.py
```diff
@@ -270,7 +271,7 @@ class Roles(NamedTuple):
     moderator = 267629731250176001
     owner = 267627879762755584
     helpers = int(environ.get("ROLE_HELPERS", 267630620367257601))
-    core_developers = 587606783669829632
+    core_developers = 814190307319087172
```
ToxicKidz, line 274: Did you forget to change this back to the correct ID?

ToxicKidz (changes requested):

bot/constants.py
Shivansh-007: Whoops yeah

bot/bot.py
Shivansh-007: Hmm, we could but I don’t see why is that better than this.

ToxicKidz: `Cog.__module__` is not the same as its extension. For example, `InternalEval.__module__` would be `bot.exts.internal_eval._internal_eval`, yet the corresponding extension is `bot.exts.internal_eval`.

Shivansh-007: I am using `Cog.__name__` here not `.__module__`

Screenshot from 2021-05-22 07-42-11

bot/bot.py
```diff
@@ -55,6 +69,14 @@ async def close(self) -> None:
         if self.redis_session:
             await self.redis_session.close()
 
+    def load_extension(self, name: str, *, package: Any = None) -> None:
```
ToxicKidz, line 72: This should be `Optional[str]`.
```suggestion
    def load_extension(self, name: str, *, package: Optional[str] = None) -> None:
```

bot/bot.py
```diff
@@ -180,6 +202,35 @@ def _remove_root_aliases(self, command: commands.Command) -> None:
...
+        Process extensions which are in unloaded cache but not found on the bot.
+
+        Send a message to #dev-alerts with the extensions, and remove them from the cache,
+        """
+        await self.wait_until_guild_available()
+
+        if not self.unloaded_extensions:
+            return
+
+        extensions_msg = '- ' + '\n- '.join(self.unloaded_extensions)
```
ToxicKidz, line 216:
```suggestion
        extensions_msg = "- " + "\n- ".join(self.unloaded_extensions)
```

ToxicKidz (changes requested):

kosayoda (changes requested):
From a cursory look at the PR:

I don't think this is a suitable way to store unloaded cogs in the cache. Currently, you manually serialize cog names by joining them with a pipe character, then setting it to a single key `unloaded` in the cache.

This is redundant when we have a cache to work with: Simply set the cog name as the cache key. That way, there is no need to even deserialize the cache into a dictionary on every time you want to modify the cache: Simply set and get the cog names from the cache.

That way, the value for each key in the cache is also freed up. One useful value for that is the message ID used to unload the cogs. In the event of an unloaded cog not found on startup, the embed can provide that message ID/link to give more context to core developers.

bot/bot.py
```diff
@@ -36,6 +41,16 @@ def __init__(self, redis_session: RedisSession, **kwargs):
         self.redis_session = redis_session
         self.loop.create_task(self.check_channels())
         self.loop.create_task(self.send_log(self.name, "Connected!"))
+        self.loop.create_task(self.process_ext_not_found())
+
+    async def get_unloaded_extensions(self) -> None:
+        """Get all the unloaded extensions from cache."""
+        # For testing sir-lancebot#705
+        await self.unloads_cache.set("bot.exts.evergreen.catify", "https://dummy-msg-link.com")
+        await self.unloads_cache.set("bot.exts.evergreen.xkd", "https://dummy-msg-link.com")
```
kosayoda, line 50: Are these for local testing? They should not be left in production.

bot/bot.py
```diff
@@ -36,6 +41,16 @@ def __init__(self, redis_session: RedisSession, **kwargs):
         self.redis_session = redis_session
         self.loop.create_task(self.check_channels())
         self.loop.create_task(self.send_log(self.name, "Connected!"))
+        self.loop.create_task(self.process_ext_not_found())
```
kosayoda, line 44: Is this guaranteed to run after all the extensions have been loaded?

bot/bot.py
```diff
@@ -180,6 +203,34 @@ def _remove_root_aliases(self, command: commands.Command) -> None:
...
+
+        Send a message to #dev-alerts with the extensions, and remove them from the cache,
+        """
+        await self.wait_until_guild_available()
+
+        if not self.unloaded_extensions:
+            return
+
+        cache_dict = await self.unloads_cache.to_dict()
+        extensions_msg = "\n- ".join(f"`{ext}`: {cache_dict.get(ext)}" for ext in self.unloaded_extensions)
```
kosayoda, line 218: You should be able to get the message link for an extension straight from the rediscache without having to get the whole cache as a dictionary.

kosayoda (changes requested):
With your current approach, you store the cached unloads into a dict, then remove successfully unloaded cogs from the dict in `load_extension`, then compare the dict with the cached unloads to find missing cogs.

A better solution would be to just go ahead with loading the extension (and checking presence in the cache before loading), and storing a set of successfully unloaded cogs instead. Then, you can get all the unloads from the cache and clean up the ones that were not in the successful dict.

This can be done with a `load_extensions` method on the bot that keep tracks of extensions that were successfully unloaded locally. That way, there's no need to store a redundant instance attribute after startup. It would also reduce the chances of race conditions with getting the unloaded cogs/loading cogs/checking for not found cogs.

bot/bot.py
Shivansh-007: Yeah these are for local testing, I have hence left a comment over that and in the PR description. I will remove that before it gets merged.

bot/bot.py
Shivansh-007: > Is this guaranteed to run after all the extensions have been loaded?

It should, yes, and while testing it did.

ToxicKidz: Yeah, the event loop gets ran after the bot instance is created

Shivansh-007:
> With your current approach, you store the cached unloads into a dict, then remove successfully unloaded cogs from the dict in `load_extension`, then compare the dict with the cached unloads to find missing cogs.
> 
> A better solution would be to just go ahead with loading the extension (and checking presence in the cache before loading), and storing a set of successfully unloaded cogs instead. Then, you can get all the unloads from the cache and clean up the ones that were not in the successful dict.
> 
> This can be done with a `load_extensions` method on the bot that keep tracks of extensions that were successfully unloaded locally. That way, there's no need to store a redundant instance attribute after startup. It would also reduce the chances of race conditions with getting the unloaded cogs/loading cogs/checking for not found cogs.
 
That would be possible if we make `load_extensions` an async function, since to check whether it exists within the async redis-cache or not it would be awaited. Hence I didn't do that.

kosayoda:
Would that be an issue? It wouldn't be much unlike how you're getting the unloaded extensions after all. I'm not an async expert though

Shivansh-007:
> Would that be an issue? It wouldn't be much unlike how you're getting the unloaded extensions after all. I'm not an async expert though

Not really but I would need to add `await`/`async` in the some places (The extensions cog, and `bot/__main__.py` probably).

bot/__main__.py
```diff
@@ -27,7 +28,16 @@
 
 bot.add_check(whitelist_check(channels=WHITELISTED_CHANNELS, roles=STAFF_ROLES))
 
-for ext in walk_extensions():
-    bot.load_extension(ext)
+
+async def load_extensions() -> None:
+    """Async wrapper to load all the extensions."""
+    await bot.unloads_cache.set("bot.exts.evergreen.catify", "https://dummy-msg-link.com")
+    await bot.unloads_cache.set("bot.exts.evergreen.xkd", "https://dummy-msg-link.com")
```
brad90four, line 35:
```py
    await bot.unloads_cache.set("bot.exts.evergreen.catify", "https://dummy-msg-link.com")
    await bot.unloads_cache.set("bot.exts.evergreen.xkd", "https://dummy-msg-link.com")
```

Are these lines the "testing" code that you mentioned?

bot/exts/utils/extensions.py
```diff
@@ -215,21 +223,25 @@ def batch_manage(self, action: Action, *extensions: str) -> str:
...
 
-        log.debug(f"Batch {verb}ed extensions.")
+        successful = [ext for ext in extensions if ext not in failures]
+        log.debug(f"Batch {verb}ed extensions: {successful}")
 
-        return msg
+        return msg, successful
 
-    def manage(self, action: Action, ext: str) -> t.Tuple[str, t.Optional[str]]:
+    async def manage(self, action: Action, ext: str) -> t.Tuple[str, t.Optional[str]]:
```
brad90four, line 231:
```suggestion
    async def manage(self, action: Action, ext: str) -> tuple[str, t.Optional[str]]:
```

brad90four (commented):
Just giving it a read through for the content. I haven't set it up or tested it out locally and I only saw two things to comment on. The first one is potentially the "testing" code as a place holder, and the other is related to the fix type annotations PR that is being worked on.

bot/__main__.py
Shivansh-007: Yes.

Xithrius:
File restructure has occurred, it may have been the cause of these conflicts.

bot/bot.py
```diff
@@ -72,6 +80,14 @@ async def close(self) -> None:
         if self.redis_session:
             await self.redis_session.close()
 
+    async def load_extension(self, name: str, *, package: Optional[str] = None) -> None:
+        """Load the extension only if it is not present in unloaded cache."""
+        if await self.unloads_cache.get(name):
+            self.successfully_unloaded.append(name)
+            log.info(f"Skipping cog {name}, found in unloaded cache.")
+        else:
+            super().load_extension(name, package=package)
```
Bluenix2, line 89: Shouldn't you also override `unload_extension()` to add the extension to the unloaded cache?

bot/bot.py
```diff
@@ -197,6 +213,38 @@ def _remove_root_aliases(self, command: commands.Command) -> None:
...
+        """
+        Process extensions which are in unloaded cache but not found on the bot.
+
+        Send a message to #dev-alerts with the extensions, and remove them from the cache,
+        """
+        await self.wait_until_guild_available()
+        cache_dict = await self.unloads_cache.to_dict()
+
+        unsuccesfully_unloaded = {
+            ext: link for ext, link in cache_dict.items() if ext not in self.successfully_unloaded
```
Bluenix2, line 226: What is this link?

bot/bot.py
```diff
@@ -197,6 +213,38 @@ def _remove_root_aliases(self, command: commands.Command) -> None:
...
+        }
+
+        if not unsuccesfully_unloaded:
+            return
+
+        extensions_msg = "\n- ".join(f"`{ext}`: {msg}" for ext, msg in unsuccesfully_unloaded.items())
+
+        dev_alerts_channel = self.get_channel(constants.Channels.dev_alerts)
+        core_dev_role = self.get_guild(constants.Client.guild).get_role(constants.Roles.core_developers)
```
Bluenix2, line 235: This does not sucessfully handle the case of someone having bad configuration and these returning None.

bot/bot.py
```diff
@@ -197,6 +213,38 @@ def _remove_root_aliases(self, command: commands.Command) -> None:
...
+        msg = (
+            f"\u26a0 {core_dev_role.mention}\n"
```
Bluenix2, line 237: What is `\u26a0`?

Can you use the [`\N{NAME}`](https://docs.python.org/3/howto/unicode.html#the-string-type) style?

bot/bot.py
```diff
@@ -197,6 +213,38 @@ def _remove_root_aliases(self, command: commands.Command) -> None:
...
+            "The following extensions were found in the cache but not on the bot:\n\n"
+            f"- {extensions_msg}"
+            "\n\nClearing them from the cache."
+        )
+
+        for ext in unsuccesfully_unloaded:
+            await self.unloads_cache.delete(ext)
```
Bluenix2, line 244: What happens if you delete half of these then one of them fail? You won't send the message that you have cleared the cache because it errors. *That said it will send a sentry error and appear in the log I believe, but my point still stands*

Can you move this down until after you have sent the message?

Bluenix2 (changes requested):

bot/bot.py
```diff
@@ -72,6 +89,20 @@ async def close(self) -> None:
         if self.redis_session:
             await self.redis_session.close()
 
+    async def load_extension(
+        self,
+        name: str,
+        *,
+        package: Optional[str] = None,
+        ignore_cache: Optional[bool] = False
```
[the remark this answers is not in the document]
Bluenix2: Why is the default False?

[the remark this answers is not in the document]
Shivansh-007: Since we load the cog earlier and then remove them from the cache (when using the `.c load` command, it would never be loaded as the cog would be present in the cache, so this option was added to force load the cog without checking the cache.

Since we don't want to force reload the cog when the bot initially loads the cogs, it is set to `False`.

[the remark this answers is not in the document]
Bluenix2: But this means that when `start()` calls `load_extension()` it will always load all cogs?

Because you do `if ignore_cache and not await self.unloads_cache.get(name):`?

I am a bit confused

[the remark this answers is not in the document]
Shivansh-007: Ah yeah, I had fixed that locally during my last testing... forgot to push it. Should be fixed now in 8193125.

bot/bot.py
```diff
@@ -72,6 +89,20 @@ async def close(self) -> None:
...
+    ) -> None:
+        """Load the extension only if it is not present in unloaded cache."""
+        if not ignore_cache and not await self.unloads_cache.get(name):
+            self.successfully_unloaded.append(name)
+            log.info(f"Skipping cog {name}, found in unloaded cache.")
```
Bluenix2, line 102:
```suggestion
            log.info(f"Skipping cog {name}, found in unloaded cache.")
```
```suggestion
            log.info(f"Skipping ext '{name}', found in unloaded cache.")
```

bot/bot.py
```diff
@@ -72,6 +89,20 @@ async def close(self) -> None:
...
 
+    async def load_extension(
+        self,
+        name: str,
+        *,
+        package: Optional[str] = None,
+        ignore_cache: Optional[bool] = False
+    ) -> None:
+        """Load the extension only if it is not present in unloaded cache."""
+        if not ignore_cache and not await self.unloads_cache.get(name):
```
Bluenix2, line 100: This if-statement is still wrong, but now for the wrong reason. `await self.unloads_cache.get(name)` becomes None and then you invert this using `not`.

What ends up happening is the opposite of what we want.

Bluenix2 (changes requested):

bot/bot.py
```diff
@@ -74,6 +74,8 @@ async def start(self, *args, **kwargs) -> None:
         """Async wrapper to load all the extensions."""
         from bot.utils.extensions import walk_extensions
 
+        await bot.unloads_cache.set("bot.exts.evergreen.xkcd", "https://dummy-msg-link.com")
+
```
Bluenix2, line 78: Can this be removed please, as a rule of thumb unless a PR is marked as a Draft or has requested changes you should feel comfortable having it merged at any time.

Shivansh-007: Yes, I thought I removed it from my commit, was testing... so needed that.

Bluenix2 (approved):
Let's get this merged! ✨ Thanks for this @Shivansh-007

bot/bot.py
```diff
@@ -62,6 +70,15 @@ async def on_thread_join(self, thread: Thread) -> None:
         with suppress(Forbidden):
             await thread.join()
 
+    async def start(self, *args, **kwargs) -> None:
+        """Async wrapper to load all the extensions."""
+        from bot.utils.extensions import walk_extensions
```
onerandomusername, line 75: why is this imported here? is it possible to import it at the top of the file?

bot/bot.py
```diff
@@ -121,7 +152,7 @@ async def check_channels(self) -> None:
             if name.startswith("_"):
                 continue
             if channel_id not in all_channels_ids:
-                log.error(f'Channel "{name}" with ID {channel_id} missing')
+                log.error(f"Channel '{name}' with ID {channel_id} missing")
```
onerandomusername, line 155: good change imo

bot/bot.py
onerandomusername: This may be worth not assuming, given that a user testing pulls may come across this error more than the referenced one.

bot/constants.py
```diff
@@ -108,6 +108,7 @@ class Channels(NamedTuple):
     organisation = 551789653284356126
     devlog = int(environ.get("CHANNEL_DEVLOG", 622895325144940554))
     dev_contrib = 635950537262759947
+    dev_alerts = int(environ.get("CHANNEL_DEVALERTS", 820712029245538314))
```
onerandomusername, line 111: `CHANNEL_DEV_ALERTS`?

onerandomusername (changes requested):
There is somewhat a problem here, conceptually.

Internally on dpy, load_extension, unload_extension, and reload_extension are all *not* awaitables.

However, the new load_extension for the bot is defined as a coro. There may be internal methods that call load_extension, but I don't think there are, so it should be good.

However, because of these three lines, 

https://github.com/python-discord/sir-lancebot/blob/029a6324823d4d8e934c12edbc41d221e59d809a/bot/exts/core/extensions.py#L24-L30

These need to all be awaited too. The problem arises in unload_extension and reload_extension. Having not been overridden, they are still normal functions, not coros. As reading the code, I can see that they are all awaited where they are used, but the functions themselves are still normal functions, and not coros.

bot/bot.py
Shivansh-007: Circular imports.

bot/exts/core/extensions.py
```diff
@@ -216,21 +224,25 @@ def batch_manage(self, action: Action, *extensions: str) -> str:
...
         """Apply an action to an extension and return the status message and any error message."""
         verb = action.name.lower()
         error_msg = None
 
         try:
-            action.value(self.bot, ext)
+            if action is action.LOAD:
+                await action.value(self.bot, ext, ignore_cache=True)
+            else:
+                action.value(self.bot, ext)
```
onerandomusername, line 241: I may have missed this code, ignore my previous write-up

onerandomusername (approved):
er, lgtm

Xithrius:
Per [this message](https://canary.discord.com/channels/267624335836053506/635950537262759947/940807906264219678) from the author, this PR is now up for grabs.

Xithrius:
This PR is now up for grabs. Would anyone that's currently subscribed to this thread or just happened to see this comment like to take over this feature?

ToxicKidz:
Hello, I’d like to have a go at this in a separate PR.

Shivansh-007:
@Xithrius Currently this PR is complete, only needs another core-dev approval.

Shivansh-007:
Cool okay 👌🏻 , let me know if you need any help while going through the code @ToxicKidz.

Xithrius:
Whoops, didn't mean to request another review from you, @Bluenix2. Sorry about that.

Xithrius:
@ToxicKidz When you have the time, please resolve the file conflicts within this PR. Thanks!

Xithrius:
Due to the staleness of this PR, I'll be closing it. If anyone wants to pick it back up, feel free to do so.

PR #760, opened 2021-06-01
Adds `.reverse`

### LINKED ISSUE #749: Add a .reverse command
## Description

Add a .reverse command that reverses a submitted string.

## Reasoning

Idk, as a joke I guess, could go with fun.py.

## Proposed Implementation

Basically, something like this 
```py
@commands.command()
async def reverse(self, ctx: commands.Context, *, text: str = "reverse time"):
    text = text[::-1]
    await ctx.send(f"> {text}")
``` 

But if no args are given, we could reverse the users profile picture and send it in an embed.

Kronifer:
## Relevant Issues
Closes #749 

## Description
Implemented the reverse command, which flips the users profile picture if no text is given, or the given text.

## Images

Profile Reversal:
image

Text Reversal:

image

## Did you:


bot/exts/evergreen/avatar_modification/_effects.py
```diff
@@ -96,6 +96,18 @@ def eight_bitify_effect(image: Image.Image) -> Image.Image:
...
         return image.quantize()
 
+    @staticmethod
+    def reverse_effect(image: Image.Image) -> Image.Image:
+        """
+        Flips the image horizontally.
+
+        This is done by just using ImageOps.flip().
+        """
+        image = image.resize((1024, 1024))
```
ChrisLovering, line 106: This line isn't needed, since you're already getting the avatar in this size

bot/exts/evergreen/avatar_modification/_effects.py
```diff
@@ -96,6 +96,18 @@ def eight_bitify_effect(image: Image.Image) -> Image.Image:
         image = image.resize((1024, 1024), resample=Image.NEAREST)
         return image.quantize()
 
+    @staticmethod
+    def reverse_effect(image: Image.Image) -> Image.Image:
```
ChrisLovering, line 100: I feel like this is better named as `flip_effect`

bot/exts/evergreen/avatar_modification/avatar_modify.py
```diff
@@ -121,6 +121,34 @@ async def eightbit_command(self, ctx: commands.Context) -> None:
 
         await ctx.send(embed=embed, file=file)
 
+    @avatar_modify.command(name="reverse", root_aliases=("reverse",))
+    async def reverse(self, ctx: commands.Context, *, text: t.Optional[str]) -> None:
+        """Either flips your profile picture or the submitted text."""
+        if not text:
+            async with ctx.typing():
+                user = await self._fetch_user(ctx.author.id)
```
ChrisLovering, line 129: Use may be none here if this call fails, we should deal with that similar to how the other commands do.

Objectivitix (changes requested):
Everything looks good and our testing worked! Some commas missing here:

image
image

ChrisLovering (changes requested):

bot/exts/evergreen/avatar_modification/_effects.py
Kronifer: alright

bot/exts/evergreen/avatar_modification/_effects.py
Kronifer: okay

bot/exts/evergreen/avatar_modification/avatar_modify.py
Kronifer: use if text is None instead?

ChrisLovering: No, I mean `self._fetch_user()` could return `None`, so you'd need to handle that case, otherwise the next line with error. The other commands in this cog already do it.

https://github.com/python-discord/sir-lancebot/blob/6dc4dcd52d7dd697a6c0ad8a7cc941eeb1d102b7/bot/exts/evergreen/avatar_modification/avatar_modify.py#L99-L102

Kronifer:
@Objectivitix the leading commas are not needed.

Kronifer:
to my knowledge at least.

bot/exts/evergreen/avatar_modification/avatar_modify.py
ChrisLovering: This hasn't been resolved.

Kronifer: OH

ChrisLovering (dismissed):

bot/exts/evergreen/avatar_modification/_effects.py
```diff
@@ -96,6 +96,17 @@ def eight_bitify_effect(image: Image.Image) -> Image.Image:
         image = image.resize((1024, 1024), resample=Image.NEAREST)
         return image.quantize()
 
+    @staticmethod
+    def flip_effect(image: Image.Image) -> Image.Image:
+        """
+        Flips the image horizontally.
+
+        This is done by just using ImageOps.flip().
```
ChrisLovering, line 104: This should be mirror

```suggestion
        This is done by just using ImageOps.mirror().
```

Objectivitix (changes requested):
The end-trailing commas are used almost everywhere in Python Discord's open-source projects, and is considered good practice. I would do it, at least, for the sake of consistency.

ToxicKidz:
> The end-trailing commas are used almost everywhere in Python Discord's open-source projects
This isn't true, and I think that it's fine the way it is.

Objectivitix:
Wat- am i hallucinating then? I remember getting reviews telling me to do that on one of the pr's

Objectivitix:
Ok, so some are and some aren't; the inconsistency can be very misleading (just like as I got misleaded). If no trailing commas is considered the better way, we should modify that all over Lance

vcokltfre:
After the spring cleanup ~~mess~~ PR we concluded that things would be changed as and when needed, rather than in a batch of 60 files at a time (although I cant say I think removing trailing commas is a good idea, since they're useful and best practice I thought)

Objectivitix:
@vcokltfre Then we should have a vote. In any case, this inconsistency is not good. It's either "PURGE THEM ALL" or "TRAILING COMMAS 4 LIFE"

Here are my thoughts on these end-trailing commas:
1. Some argue that such a miniscule change won't affect readability and instead simply adds extra stuff, but I think that it makes code more symmetrical and pleasant on the eyes (all lines will have the same format).
2. Easy extendability: if one needs to add one more entry, this will largely decrease the chances that they forget to add a comma before newlining.
3. There are multiple articles on the Internet giving even more reasons, such as easy copy-paste code manipulation.

This might not be the best place to talk about this, thus I will open an issue and discuss in further detail :D

janine9vn:
Please see what PEP8 has to say about trailing commas: https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas. Our style guidelines falls back to PEP8 and regardless trailing commas are good since we use version control and they ensure the git blame is accurate.

Objectivitix:
alright, but regarding that code refactoring... currently there are a lot of places where the dangling comma is not used

ChrisLovering (changes requested):

bot/exts/evergreen/avatar_modification/avatar_modify.py
```diff
@@ -121,6 +121,50 @@ async def eightbit_command(self, ctx: commands.Context) -> None:
...
+        if not ctx.invoked_subcommand:
+            await invoke_help_command(ctx)
+
+    @reverse.command(name="text")
+    async def reverse_text(self, ctx: commands.Context, *, text: str) -> None:
+        """Sends the given text backwards."""
+        await ctx.send(f"> {text[::-1]}", allowed_mentions=discord.AllowedMentions.none())
+
+    @reverse.command(name="image")
+    async def reverse_photo(self, ctx: commands.Context) -> None:
```
ChrisLovering, line 136: I feel like this should be the default behaviour of `.reverse` so if a user doesn't so `.reverse text <some text>` it defaults to reversing their avatar, rather than requiring them to do `.reverse image`

ChrisLovering (changes requested):
Two small comments left, rest looks good :)

Shivansh-007 (approved):
Other than the changes mentioned by Chris, this looks good!

bot/exts/evergreen/avatar_modification/avatar_modify.py
```diff
@@ -121,6 +121,50 @@ async def eightbit_command(self, ctx: commands.Context) -> None:
...
+            file = await in_executor(
+                PfpEffects.apply_effect,
+                image_bytes,
+                PfpEffects.flip_effect,
+                filename
+            )
+
+            embed = discord.Embed(
+                title="Your reversed avatar.",
+                description="Here is your reversed avatar. I think it is a spitting image of you."
```
Vthechamp22, line 160: Did you mean `splitting`?

wookie184: `spitting image` is correct, https://www.merriam-webster.com/dictionary/spitting%20image. `splitting image` would be a great pun for the `.mosaic` command though, thinking about it 😄

Kronifer:
I'm gonna be getting back to this after an eternity, sorry for keeping y'all hanging

bot/exts/evergreen/avatar_modification/avatar_modify.py
Kronifer: Merged reverse into one command, resolved

bot/exts/evergreen/avatar_modification/avatar_modify.py
```diff
@@ -121,6 +121,42 @@ async def eightbit_command(self, ctx: commands.Context) -> None:
...
         await ctx.send(embed=embed, file=file)
 
+    @avatar_modify.command(name="reverse", root_aliases=("reverse",))
+    async def reverse(self, ctx: commands.Context, *, text: t.Optional[str]) -> None:
+        """
+        Reverses the sent text.
+
+        If no text is provided, the user's profile picture will be reversed.
+        """
+        if not text:
```
ChrisLovering, line 131: The changes look good now :) One small comment I have is that we can avoid this huge if statement here by doing the unhappy path first IE
```py
if text:
    await ctx.send(f"> {text[::-1]}", allowed_mentions=discord.AllowedMentions.none())
    return
``` and then doing the rest of the code. This will avoid the large nested block

Kronifer: alright, gimme a few seconds

ChrisLovering (approved):
Nice, looks good!

wookie184 (approved):
Works well

PR #785, opened 2021-07-07
Correct an off-by-one bug due to bisect.bisect()

### LINKED ISSUE #784: Wrong title in love calc when 0% love
## Description
In love calculator when love is found 0% it does not show "There's no real connection between you two" like in [json file](https://github.com/python-discord/sir-lancebot/blob/main/bot/resources/valentines/love_matches.json).

## Steps to Reproduce
Found it while getting love as 0%

## Expected Behaviour
It should be "There's no real connection between you two" as per json(I did try to check if it was sarcasm for 0% but that's not the case)

## Actual Behaviour

image
image

## Known Impacted Platforms
- [x] Web
- [x] Desktop
- [x] Android App
- [x] iOS App

## Possible Solutions
Something should be wrong here:
```
        m = hashlib.sha256(who.encode() + whom.encode())
        # Mod 101 for [0, 100]
        love_percent = sum(m.digest()) % 101

        # We need the -1 due to how bisect returns the point
        # see the documentation for further detail
        # https://docs.python.org/3/library/bisect.html#bisect.bisect
        index = bisect.bisect(LOVE_DATA, (love_percent,)) - 1
```
in [file](https://github.com/python-discord/sir-lancebot/blob/main/bot/exts/valentines/lovecalculator.py)

PythonTryHard:
## Relevant Issues

Closes #784 

## Description

I removed the `-1` from the quote index calculation at <https://github.com/python-discord/sir-lancebot/blob/3218ff7d59c9db211039414b8c84ea098fd784d6/bot/exts/valentines/lovecalculator.py#L75> This means the command `[p]love` (`[p]` refers to Lancebot's prefix) will now return the correct quote for `love_percent = 0` edge-case, and scores in the range [96, 100] will now correctly use the 100 quote from `bot/resources/valentines/love_matches.json` instead of the 95 quote.

## Did you:

### DIFF
--- a/bot/exts/holidays/valentines/lovecalculator.py
+++ b/bot/exts/holidays/valentines/lovecalculator.py
@@ -74,7 +74,8 @@ class LoveCalculator(Cog):
         # We need the -1 due to how bisect returns the point
         # see the documentation for further detail
         # https://docs.python.org/3/library/bisect.html#bisect.bisect
-        index = bisect.bisect(LOVE_DATA, (love_percent,)) - 1
+        love_threshold = [threshold for threshold, _ in LOVE_DATA] 
+        index = bisect.bisect(love_threshold, love_percent) - 1
         # We already have the nearest "fit" love level
         # We only need the dict, so we can ditch the first element
         _, data = LOVE_DATA[index]
--- a/bot/exts/holidays/valentines/lovecalculator.py
+++ b/bot/exts/holidays/valentines/lovecalculator.py
@@ -74,7 +74,7 @@ class LoveCalculator(Cog):
         # We need the -1 due to how bisect returns the point
         # see the documentation for further detail
         # https://docs.python.org/3/library/bisect.html#bisect.bisect
-        love_threshold = [threshold for threshold, _ in LOVE_DATA] 
+        love_threshold = [threshold for threshold, _ in LOVE_DATA]
         index = bisect.bisect(love_threshold, love_percent) - 1
         # We already have the nearest "fit" love level
         # We only need the dict, so we can ditch the first element


Akarys42:
Converted to draft as the issue isn't approved yet

Xithrius:
I've approved.

Xithrius (approved):

wookie184 (changes requested):
While this works fine, I don't think the current comment is entirely clear. It says that `-1` is needed, but it isn't very clear why, and if we aren't using it then surely it *isn't* needed.

From some testing it seems that `-1` was there to prevent values triggering on the number before. e.g. without `-1` the message for `100` in the JSON also appears on `99`. I think this behaviour isn't a problem, but the current comment should be changed to explain this behaviour, or at least remove the bit that says `-1` is needed.

PythonTryHard:
I'll try get back on my dev environment, though probably not until 2021-08-25 due to irl complications. Reading back, I kinda with you, comment isn't very clear

Xithrius:
@PythonTryHard How's it going?

lxnn:
Hello @PythonTryHard 

Can I just clarify, is the intended behaviour that you should get:
- the `0` message for a score in the range `0-4` inclusive,
- the `5` message for `5-19`,
- ...
- the `95` message for `95-99`, and
- the `100` message for just `100`?

If that is the case, then you would want to use `bisect.bisect_left` rather than `bisect.bisect_right` (which is what `bisect.bisect` is an alias of). It's actually only due to a quirk of tuple-comparison (shorter tuples are lesser) that the number `bisect.bisect(LOVE_DATA, (love_percent,))` lies within the valid range of indices for `LOVE_DATA`. The number `bisect.bisect_left(LOVE_DATA, (love_percent,))` is guaranteed to be a valid index assuming that the largest value in `LOVE_DATA` is `(100, X)` (although I believe this assumption should be made explicit with an `assert` statement).

I'm not sure I've worded this very clearly, so let me know if I can explain anything.

lxnn:
Hello @PythonTryHard 

Can I just clarify, is the intended behaviour that you should get:
- the `0` message for a score in the range `0-4` inclusive,
- the `5` message for `5-19`,
- ...
- the `95` message for `95-99`, and
- the `100` message for just `100`?

I believe there might be a cleaner way to implement this. Let me know if you want me to elaborate.

Xithrius:
The file that is changed within this PR has been moved to `bot/exts/holidays/valentines/lovecalculator.py` due to the file restructuring that has occurred.

Please resolve the file conflict to reflect these changes.

PythonTryHard:
> @PythonTryHard How's it going?

Barely opened GitHub. Turns out the semester hits me harder than expected. I'll need to set up a temp VM for dev work now install of constant rebooting to my Linux dual-boot.

> [...] Let me know if you want me to elaborate.

@lxnn Yes please

Kronifer (approved):
Looks good

PythonTryHard:
After some diving into the source code of `bisect..bisect()` and playing around with `LOVE_DATA` I found that tuple comparison was the bug. The fix I came up with was to pull the love percent thresholds out and bisect there instead of a half-arsed `- 1` and a bad comment. I also force-pushed my branch to address the file restructuring.

vcokltfre (approved):

lxnn:
Sorry, I totally forgot I'd commented here. (I need to read my emails more often 😅) 

> After some diving into the source code of `bisect.bisect()` and playing around with `LOVE_DATA` I found that tuple comparison was the bug. The fix I came up with was to pull the love percent thresholds out and bisect there instead of a half-arsed `- 1` and a bad comment. I also force-pushed my branch to address the file restructuring.

Yes, this is roughly what I was going to suggest. That is, separating the thresholds from the data, then bisecting the thresholds.

There's a neat trick for "unzipping" a list of tuples:
```python
THRESHOLDS, DATA = zip(*LOVE_DATA)
```

I was also going to suggest using `bisect.bisect_left` rather than `bisect.bisect_right` (which `bisect.bisect` is an alias of).  As it stood, the code relied on a quirk of tuple comparison (shorter tuples are lesser when tied on the last element of the shorter tuple). You have sorted that, but I think using `bisect_left` would make it simpler for the reason that it will return an index of the list so long as the final threshold is `100` or greater. It does change the behaviour slightly, but not in a way that really matters. It would behave as I described in my last message.

It's totally up to you whether you use this suggestion.

wookie184 (approved):
LGTM

wookie184:
Merged this now as the current code seems to work well, although anyone can feel free to open a new issue/PR if they have ideas for further changes

PR #786, opened 2021-07-09
Duckduckduckgoose

### LINKED ISSUE #782: A Competitive Duck Game
## Description
Implement a game inspired by the classic card game, Set, using pydis ducks.

After running the command, the bot posts an embed with an image of the board consisting of 12 cards. Players race to claim [valid sets](https://en.wikipedia.org/wiki/Set!) of 3 cards. Correct answers earn them a point, incorrect answers deduct a point. After a time limit the game ends, and the scores are displayed in the initial embed along with any valid sets that went unclaimed. Alternatively, a player may call "goose" if they think there are no valid sets remaining. If they are right, they get 2 points and the game ends; if they're wrong they lose a point.

A card needs to have 4 easily-identifiable features, each with 3 possible values. I thought ducks would be appropriate, so for the proof of concept I chopped and screwed the random duck generation code from the [pydis staff's PyWeek entry](https://github.com/python-discord/aaaaAAAA).

### Copyright concerns
The only commonality with Set is the mathematical definition of a valid set. The scoring system, structure of the game, and art are all different. As far as I understand, game mechanics are not copyrightable anyway. 

## Reasoning
The game is super fun and challenging.

## Proposed Implementation

dddg example

## Additional Details
I'm not very experienced with graphic design. I've already implemented the image generation for the proof of concept, but I'm sure I committed some sins. So if anyone is interested in taking on that part I would not stand in their way.

camcaswell:
## Relevant Issues
Closes #782 

## Description
Adds a cog to run Duck Duck Duck Goose games. It randomly chooses cards for the game, then crops the corresponding card images from a pregenerated PNG file of every possible card and pastes them together into a board with index labels.

It listens for comments in channels with a currently running game for those that contain three numbers, and parses them as potential answers. The game ends either when the time limit is up (set at 3 minutes), or when a player correctly calls that all the valid answers have been claimed. The board embed is edited to add claimed answers, and the scoreboard at the end of the game. It reacts to comments that contain three numbers which are valid indices but which don't make a valid set with an "x" (and also incorrect game-end calls).

There is also a help command that explains how to play the game, and a stop command only available to mods to force-stop the game.

## Did you:


camcaswell:
Maybe something to think about: the large-ish image file is currently opened at load time. I don't have a good sense of the bot's resource limitations, so is this bad? fine? I don't know.

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+
+    @start_game.command(name="help")
+    async def show_rules(self, ctx: commands.Context) -> None:
+        """Explain the rules of the game."""
+        await self.send_help_embed(ctx)
+
+    @start_game.command(name="stop")
+    async def stop_game(self, ctx: commands.Context) -> None:
+        """Stop a currently running game. Only available to mods."""
+        if not any(role in ctx.author.roles for role in MODERATION_ROLES):
```
wookie184, line 291: You can use the [`with_role`](https://github.com/python-discord/sir-lancebot/blob/e6e280cd13236647ffcaa35f522baeb747fbf97e/bot/utils/decorators.py#L147) decorator instead of this check. Example of it's usage [here](https://github.com/python-discord/sir-lancebot/blob/e6e280cd13236647ffcaa35f522baeb747fbf97e/bot/exts/internal_eval/_internal_eval.py#L153). 

Also, I'm not sure if this current code would work as `MODERATION_ROLES` contains id's (ints) but you're comparing against a list of role objects (ctx.author.roles)

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+
+""" Distribution of minimum acceptable solutions at board generation.
+    This is for gameplay reasons, to shift the number of solutions per board up,
+    while still making the end of the game unpredictable.
+    Note: this is *not* the same as the distribution of number of solutions.
+"""
+SOLN_DISTR = 0, 0.05, 0.05, 0.1, 0.15, 0.25, 0.2, 0.15, .05
+
+image_path = Path("bot", "resources", "evergreen", "all_cards.png")
+font_path = Path("bot", "resources", "evergreen", "LuckiestGuy-Regular.ttf")
```
wookie184, line 36: could probably be capitalised as they're constants.

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+CORRECT_SOLN = 1
+INCORRECT_SOLN = -1
+CORRECT_GOOSE = 2
+INCORRECT_GOOSE = -1
+
+""" Distribution of minimum acceptable solutions at board generation.
+    This is for gameplay reasons, to shift the number of solutions per board up,
+    while still making the end of the game unpredictable.
+    Note: this is *not* the same as the distribution of number of solutions.
+"""
```
wookie184, line 32: Block comments should be lines starting with `#`s rather than a docstring, as per pep 8 https://www.python.org/dev/peps/pep-0008/#block-comments

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+        """Calculate valid solutions and cache to avoid redoing work."""
+        if self._solutions is None:
+            self._solutions = set()
+            for idx_a, card_a in enumerate(self.board):
+                for idx_b, card_b in enumerate(self.board[idx_a+1:], start=idx_a+1):
+                    """
+                        Two points determine a line, and there are exactly 3 points per line in {0,1,2}^4.
+                        The completion of a line will only be a duplicate point if the other two points are the same,
+                        which is prevented by the triangle iteration.
+                    """
```
wookie184, line 132: Same as comment about comments above, I wouldn't use a docstring here

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+            return
+
+        # Also be forgiving for answers that have already been claimed (and avoid penalizing for racing conditions).
+        if answer in game.claimed_answers:
+            return
+
+        if answer in game.solutions:
+            game.claimed_answers[answer] = msg.author
+            game.scores[msg.author] += CORRECT_SOLN
+            await self.display_claimed_answer(game, msg.author, answer)
```
wookie184, line 222: I think a tick reaction should be added here, to make it more clear it was accepted if the board was buried and save having to look up at it, and also as I think the user would expect it as a cross is added for not accepted answers

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+        self.current_games = {}
+
+    @commands.group(name='duckduckduckgoose', aliases=['dddg', 'duckgoose'], invoke_without_command=True)
+    @commands.cooldown(rate=1, per=2, type=commands.BucketType.channel)
+    async def start_game(self, ctx: commands.Context) -> None:
+        """Generate a board, send the game embed, and end the game after a time limit."""
+        if ctx.channel.id in self.current_games:
+            return
```
wookie184, line 161: Maybe send a message saying there is already a game running

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+        if not (match := re.match(ANSWER_REGEX, msg.content)):
+            return
+        answer = tuple(sorted(int(m) for m in match.groups()))
+
+        # Be forgiving for answers that use indices not on the board.
+        if not all(0 <= n < len(game.board) for n in answer):
+            return
+
+        # Also be forgiving for answers that have already been claimed (and avoid penalizing for racing conditions).
+        if answer in game.claimed_answers:
```
wookie184, line 216: I'm wondering if it would be worth adding a reaction to this to indicate that it was reused, probably something different to the cross as they didn't lose a point. Maybe `:recycle:`?

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+class DuckGame:
+    """A class for a single game."""
+
+    def __init__(self,
+                 rows: int = 4,
+                 columns: int = 3,
+                 minimum_solutions: int = 1,
+                 ) -> None:
```
wookie184, line 88: I would usually format something like this like
```python
    def __init__(
        self,
        rows: int = 4,
        columns: int = 3,
        minimum_solutions: int = 1,
    ) -> None:
```
this is what the style guide suggests too https://pythondiscord.com/pages/guides/pydis-guides/contributing/style-guide/#bracket-and-item-arrangement

wookie184 (changes requested):
This is a fun game, I found it quite difficult😅. The images and handling of the images is really neat.

A few thoughts:
 - I think it would be a good idea to send the scoreboard as a new message after the game has finished, imo editing the original message isn't a very clear UI, especially as channels can get very busy.
 - I think the help command could maybe be a bit clearer, it could probably do with being a bit more verbose to ensure it is clear for somebody who has no idea what the game is.
 - The mod `stop` command should send a message saing the game was cancelled, or send a message saying there was no active game. In general i'd say any user action should result in a response from the bot in either a message or reaction.
 - RE you wondering about resource usage, I'm pretty sure it should be fine, although someone else will know better.

bot/exts/evergreen/duck_game.py
```diff
@@ -0,0 +1,329 @@
...
+            color=discord.Color.dark_purple(),
+        )
+        embed.description = (
+            "**Each card has 4 features**\nColor, Number, Hat, and Accessory\n"
+            "\n**A valid flight**\n3 cards where each feature is either all the same or all different\n"
+            '\n**Call "GOOSE"**\nif you think there are no more flights\n'
+            "\n**1**   for each valid flight\n"
+            '**2**   for a correct "GOOSE" call\n'
+            "**-1** for any wrong answer"
+        )
```
wookie184, line 313: I think this might be clearer as a multiline string constant defined at the top, then you can use actual newline characters instead of `\n`, making it more clear what the result will actually look like. Would also mean you don't have to use a mixture of single and double quoted strings.

Also not sure why there are 3 spaces, e.g. between `**1**` and  `for` on line 310, seems to render with just 1 in discord though, not sure if i'm missing something

Xithrius:
@camcaswell will you be fixing up this PR?

camcaswell:
> 
> 
> @camcaswell will you be fixing up this PR in the near future?

Yes, sorry I've been busy lately but I'll get to it

Xithrius:
Thank you for the update.

bot/exts/evergreen/duck_game.py
camcaswell: So this was the way I was doing it originally, but reacting to every message has the effect of burying the board even faster. My reasoning was that, in terms of gameplay, the player needs to know if their answer was incorrect (so they can correct it, or see where they went wrong), but doesn't immediately need to know whether their answer was accepted or was a repeat.

Currently I have it updating the list of claimed solutions below the board. I found that to be a good compromise in testing, since the biggest problem is the board disappearing off-screen.

bot/exts/evergreen/duck_game.py
camcaswell: My reasoning for this is the same as here: https://github.com/python-discord/sir-lancebot/pull/786#discussion_r692440022

ChrisLovering (approved):
Looks great, had some good fun playing it while testing!

PR #795, opened 2021-08-02
Handle Wikipedia API errors and warnings in the Wikipedia Cog

### LINKED ISSUE #793: KeyError: 'query' (wikipedia cog)
Sentry Issue: [SIR-LANCEBOT-5T](https://sentry.io/organizations/python-discord/issues/2548014774/?referrer=github_integration)

```
KeyError: 'query'
  File "discord/ext/commands/core.py", line 85, in wrapped
  File "bot/exts/evergreen/wikipedia.py", line 78, in wikipedia_search_command
    contents = await self.wiki_request(ctx.channel, search)
  File "bot/exts/evergreen/wikipedia.py", line 42, in wiki_request
    number_of_results = raw_data["query"]["searchinfo"]["totalhits"]

Unhandled command error: 'query'
```

This looks like the request didn't return an expected result, even with a 200 response code.

D0rs4n:
## Relevant Issues
Closes #793 
## Description
I read the Wikipedia API docs, and noticed that when the Wikipedia API encounters some kind of error or warning, it'll send back the response with the proper error message, but regardless of that the HTTP Reponse Code still remains 200, so I added an additional check to avoid such issues in the future. [Docs](https://www.mediawiki.org/wiki/API:Errors_and_warnings) (the Wikipedia API uses the same API system)

## Did you:


ChrisLovering (changes requested):
I think we should refactor this functionality entirely.

Ideally we should check the bad path first and then return, so that we don't havea  huge if/else block.

For example:
```
if resp.status != 200:
   ... send error message ...
   return
... carry on as normal ...
```

Also ideally for errors, we would raise an error and then let the error_handler cog deal with sending an error embed. Having a quick look at the error handler, we don't have a good error being handled there for API issues. So consider this an optional thing if you want to take it on!

bot/exts/evergreen/wikipedia.py
```diff
@@ -39,31 +39,42 @@ async def wiki_request(self, channel: TextChannel, search: str) -> Optional[List
...
+                    else:
+                        log.info("There was an issue when trying to communicate with the Wikipedia API")
 
-                else:
                     await channel.send(
-                        "Sorry, we could not find a wikipedia article using that search term."
-                    )
+                        "There was an issue processing your Wikipedia request, please try again later.")
                     return
+                else:
```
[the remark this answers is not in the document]
D0rs4n: As for the first part, I think that wouldn't work since as the docs says the Wikipedia API "usually" returns a 200 status code even for errors and warnings, [(Docs)](https://www.mediawiki.org/wiki/API:Errors_and_warnings) so checking the status is actually not useful. I just left it there in case the API goes down.
As for the second part, yes, I'm not sure why I left it there! 😄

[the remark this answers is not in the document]
D0rs4n: Also, the API should never even send back an error, just in case of malformed query URLs, and I'm not sure why this error was raised in the first place, I tested out the payload that caused the error, but I could only reproduce it once.

bot/exts/evergreen/wikipedia.py
```diff
@@ -37,37 +38,42 @@ async def wiki_request(self, channel: TextChannel, search: str) -> Optional[List
...
-                        lines.append(line)
-
-                    return lines
+            if resp.status != 200:
+                log.info(f"Unexpected response `{resp.status}` while searching wikipedia for `{search}`")
+                raise ExternalAPIError("Wikipedia API")
 
+            raw_data = await resp.json()
+            if raw_data.get("query", None) is None:
+                if raw_data.get("errors", None) is not None:
```
Shivansh-007, line 47: `dict.get()` sets the second parameter i.e. the default value as `None` by default so that can be ignored, another thing you can change is `is None` and `is not none` using `not`.

```suggestion
            if not raw_data.get("query"):
                if raw_data.get("errors"):
```

bot/utils/exceptions.py
```diff
@@ -1,4 +1,14 @@
...
     pass
+
+
+class ExternalAPIError(Exception):
+    """Raised when an external API(eg. Wikipedia) returns an error."""
+
+    def __init__(self, api: str):
+        super().__init__()
+        self.api = api
+    pass
```
Shivansh-007, line 14: Good addition, can we add `status_code` and an optional argument that contains the error message received by the API too?

bot/utils/exceptions.py
```diff
@@ -1,4 +1,14 @@
+
 class UserNotPlayingError(Exception):
     """Raised when users try to use game commands when they are not playing."""
 
     pass
+
+
+class ExternalAPIError(Exception):
+    """Raised when an external API(eg. Wikipedia) returns an error."""
```
Shivansh-007, line 9: Nitpick
```suggestion
    """Raised when an external API(eg. Wikipedia) returns an error response."""
```

Shivansh-007 (changes requested):

bot/utils/exceptions.py
D0rs4n: Good point, right now it's too specific to the Wikipedia API
(that returns 200 even if it sends back an error :D )

bot/exts/evergreen/error_handler.py
```diff
@@ -113,13 +113,23 @@ async def on_command_error(self, ctx: commands.Context, error: commands.CommandE
             return
 
         if isinstance(error, commands.CheckFailure):
-            await ctx.send(embed=self.error_embed("You are not authorized to use this command.", NEGATIVE_REPLIES))
```
Shivansh-007, line 116: There wasn't a need to change this line as it is in one line and doesn't fall over to the next line. Basically what I meant was, instead of:
```py
await ctx.send(embed=self.error_embed("This is a long message",
               argument_2)
```
And it should rather be:
```py
await ctx.send(
    embed=self.error_embed("This is a long message",
    argument_2
)
```

bot/exts/evergreen/error_handler.py
```diff
@@ -113,13 +113,23 @@ async def on_command_error(self, ctx: commands.Context, error: commands.CommandE
...
 
         if isinstance(error, UserNotPlayingError):
             await ctx.send("Game not found.")
             return
 
+        if isinstance(error, ExternalAPIError):
+            await ctx.send(
+                embed=self.error_embed(
+                    f"There was an error when communicating with the {error.api}",
+                    NEGATIVE_REPLIES))
```
Shivansh-007, line 130: This is still not following the style guide, look at the above comment.

bot/exts/evergreen/wikipedia.py
```diff
@@ -46,37 +47,59 @@ async def wiki_request(self, channel: TextChannel, search: str) -> Optional[List
...
+                            re.sub(
+                                WIKI_SNIPPET_REGEX, "", article["snippet"]
+                            )
+                        ),
+                        url=f"https://en.wikipedia.org/?curid={article['pageid']}"
                     )
-                    return
+                    lines.append(line)
+
+                return lines
```
Shivansh-007, line 98: Looks like when you where resolving git conflicts, you merged both the changes. Kindly remove one.

Shivansh-007 (changes requested):

Shivansh-007 (approved):
As a pure code review, this looks good to me :shipit:

D0rs4n:
Thank you for you review! I should've paid more attention to the style guide! 👍 
Learned my lesson!

bot/utils/exceptions.py
```diff
@@ -1,4 +1,17 @@
+from typing import Optional
+
+
 class UserNotPlayingError(Exception):
     """Raised when users try to use game commands when they are not playing."""
 
     pass
+
+
+class ExternalAPIError(Exception):
```
RohanJnr, line 10: I would just call it `APIError`

bot/exts/evergreen/wikipedia.py
```diff
@@ -46,37 +47,39 @@ async def wiki_request(self, channel: TextChannel, search: str) -> Optional[List
...
+
             else:
-                log.info(f"Unexpected response `{resp.status}` while searching wikipedia for `{search}`")
                 await channel.send(
-                    "Whoops, the Wikipedia API is having some issues right now. Try again later."
+                    "Sorry, we could not find a wikipedia article using that search term."
                 )
                 return
```
RohanJnr, line 84: Passing an empty list here instead of having the else block here would be better as you are already checking if `contents` is populated in the `wikipedia_search_command`. 
```suggestion
            return lines
```
The else block can be used in the `wikipedia_search_command`. This will require you to define the `lines` variable (line 64) outside the if-else scope.
And also don't forget to change the function return annotation to only `List[str]` instead of `Optional[List[str]]`.

bot/exts/evergreen/wikipedia.py
```diff
@@ -46,37 +47,39 @@ async def wiki_request(self, channel: TextChannel, search: str) -> Optional[List
...
-
-                else:
-                    await channel.send(
-                        "Sorry, we could not find a wikipedia article using that search term."
+            if (status := resp.status) != 200:
+                log.info(f"Unexpected response `{status}` while searching wikipedia for `{search}`")
+                raise ExternalAPIError("Wikipedia API", status)
+
+            raw_data = await resp.json()
+            if not raw_data.get("query"):
```
RohanJnr, line 55:
```suggestion
            raw_data = await resp.json()
      
            if not raw_data.get("query"):
```
An empty line to keep things spaced.

RohanJnr (changes requested):

RohanJnr:
We should handle APIErrors and ResponseCode errors (if the response code is not 200). APIError could mean there is an error with the feed of the response which this PR is solving. We might have unexpected errors where the response code is not 200, do we have any instances of this situation of non 200 responses?

bot/exts/evergreen/wikipedia.py
```diff
@@ -96,6 +93,11 @@ async def wikipedia_search_command(self, ctx: commands.Context, *, search: str)
             await LinePaginator.paginate(
                 contents, ctx, embed
             )
+        else:
+            await ctx.send(
+                "Sorry, we could not find a wikipedia article using that search term."
+            )
+            return
```
RohanJnr, line 100: A return statement is not required.
```suggestion
```

RohanJnr (approved):

D0rs4n:
> We should handle APIErrors and ResponseCode errors (if the response code is not 200). APIError could mean there is an error with the feed of the response which this PR is solving. We might have unexpected errors where the response code is not 200, do we have any instances of this situation of non 200 responses?
> 
> This issue can be delt separately from the issues being solved by this PR but I would like to know what others think about this.

The Wikipedia API will always send back a response code of 200 according to the MediaWiki API specification [docs](https://www.mediawiki.org/wiki/API:Errors_and_warnings) (Wikipedia uses the same). So I don't think there's a case, the only reason I added the check is for the sake of consistency, but there shouldn't be any case when it is 200, and if it's not 200 then it must be an APIError.

bot/exts/evergreen/wikipedia.py
```diff
@@ -42,43 +43,39 @@ class WikipediaSearch(commands.Cog):
...
-                            url=f"https://en.wikipedia.org/?curid={article['pageid']}"
-                        )
-                        lines.append(line)
-
-                    return lines
-
-                else:
-                    await channel.send(
-                        "Sorry, we could not find a wikipedia article using that search term."
+            if (status := resp.status) != 200:
```
Bluenix2, line 50: Unnecessary variable in my opinion

bot/exts/evergreen/wikipedia.py
```diff
@@ -42,43 +43,39 @@ class WikipediaSearch(commands.Cog):
...
+            raw_data = await resp.json()
+
+            if not raw_data.get("query"):
+                if error := raw_data.get("errors"):
+                    log.error(f"There was an error while communicating with the Wikipedia API: {error}")
+                raise APIError("Wikipedia API", status, error)
+
+            number_of_results = raw_data["query"]["searchinfo"]["totalhits"]
+            lines = []
+            if number_of_results:
```
Bluenix2, line 63: `number_of_results` is not used anywhere else?
```suggestion
            lines = []
            if raw_data["query"]["searchinfo"]["totalhits"]:
```

bot/utils/exceptions.py
```diff
@@ -1,4 +1,17 @@
...
+
+
 class UserNotPlayingError(Exception):
     """Raised when users try to use game commands when they are not playing."""
 
     pass
+
+
+class APIError(Exception):
+    """Raised when an external API(eg. Wikipedia) returns an error response."""
```
Bluenix2, line 11:
```suggestion
    """Raised when an external API (eg. Wikipedia) returns an error response."""
```

bot/exts/evergreen/wikipedia.py
D0rs4n: It is passed to the APIErro exception as an argument. 
(Since, different APIs return different status codes for errors, so I added a field to the error to indicate those status codes)

Bluenix2: Right, but you have `resp.status` right there haha

D0rs4n: Alright, valid point. I'll remove it!

bot/exts/evergreen/wikipedia.py
```diff
@@ -42,43 +43,38 @@ class WikipediaSearch(commands.Cog):
...
+            if not raw_data.get("query"):
+                if error := raw_data.get("errors"):
+                    log.error(f"There was an error while communicating with the Wikipedia API: {error}")
+                raise APIError("Wikipedia API", resp.status, error)
+
+            lines = []
+            if raw_data["query"]["searchinfo"]["totalhits"]:
+                results = raw_data["query"]["search"]
+
+                for article in results:
```
Bluenix2, line 65:
```suggestion
                for article in raw_data["query"]["search"]:
```

D0rs4n: I pushed the changes, although the only thing I changed in the code was a few checks, that was the base code, it made it a bit more consistent, but I understand that those variables were unnecessary.

Bluenix2 (approved):
This looks good now to me, thank you

Xithrius (approved):

PR #806, opened 2021-08-10
Love command checks

ChrisLovering:
## Description

This makes a few changes to the `.love` command to make it slightly less creepy, more consistent with other lovefest commands and also makes it give consistent output, regardless of input order.

The changes I have made are:
 - Added and `in_month` check, to ensure the command can only run during February (valentine season)
 - Require that both members have opted into the love fest event (by checking roles)
 - Sort the members to be checked before hashing, to ensure symmetrical results.
 - Removed some now un-used checks since we don't accept arbitrary strings as input

## Did you:

### DIFF
--- a/bot/exts/valentines/lovecalculator.py
+++ b/bot/exts/valentines/lovecalculator.py
@@ -12,7 +12,7 @@ from discord.ext import commands
 from discord.ext.commands import BadArgument, Cog, clean_content
 
 from bot.bot import Bot
-from bot.constants import Client, Lovefest, Month
+from bot.constants import Channels, Client, Lovefest, Month
 from bot.utils.decorators import in_month
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


bot/exts/valentines/lovecalculator.py
```diff
@@ -22,45 +24,41 @@
...
-            else:
-                # Otherwise normalise case and remove any leading/trailing whitespace
-                arg = arg.strip().title()
-            # This has to be done manually to be applied to usernames
-            return clean_content(escape_markdown=True).convert(ctx, arg)
+        if not all((
+            Lovefest.role_id in [role.id for role in who.roles],
+            Lovefest.role_id in [role.id for role in whom.roles]
+        )):
+            raise BadArgument("Both members must have the love fest role!")
```
Akarys42, line 54:
```suggestion
            raise BadArgument("Both members must have the lovefest role!")
```

Akarys42 (changes requested):
Sorry 😔

bot/exts/valentines/lovecalculator.py
ChrisLovering: How about this?
```suggestion
            raise BadArgument("Both members must have the `Lovefest` role!")
```

ChrisLovering: actually, never mind. It's referred to as lovefest in other commands, so lets stay consistent.

bot/exts/valentines/lovecalculator.py
```diff
@@ -22,45 +24,41 @@
...
-            else:
-                # Otherwise normalise case and remove any leading/trailing whitespace
-                arg = arg.strip().title()
-            # This has to be done manually to be applied to usernames
-            return clean_content(escape_markdown=True).convert(ctx, arg)
+        if not all((
+            Lovefest.role_id in [role.id for role in who.roles],
+            Lovefest.role_id in [role.id for role in whom.roles]
+        )):
+            raise BadArgument("Both members must have the lovefest role!")
```
wookie184, line 54: Might be worth mentioning how to get the lovefest role, something like
```suggestion
            raise BadArgument("Both members must have the lovefest role! You can get this by running the `.lovefest sub` command.")
```
if the `who` or `whom` is the invoker we probably wouldn't need to require the lovefest role, but the current behaviour is fine.

wookie184 (changes requested):
Could add a footer to the embed sent saying you can unsubscribe with `.lovefest unsub`, but you could also not do that, I don't' mind.

wookie184 (approved):
lgtm

Xithrius (approved):
Yes.

PR #816, opened 2021-08-23
Whitelist `.bm` command in occupied help channels

### LINKED ISSUE #811: Allow bookmark command  `.bm` by users in help channels
## Description

The proposal is to allow users to run the bookmark `.bm` command in help channels. Currently the command can be run in some other channels such as off-topic, but it is not allowed in the help channels. 

## Reasoning

Sometimes I wish to bookmark a particularly helpful response I receive in a help channel without having to go to #bot-commands or #sir-lancebot-playground. It would be very nice for users to be able to run `.bm` within their current help channel.

## Proposed Implementation

Whitelist `.bm` command in channels under the help category. 

Objectivitix:
## Relevant Issues
Closes #811

## Description
Allows the execution of the `.bm` command by users in help channels.

## Did you:

### DIFF
--- a/bot/exts/evergreen/bookmark.py
+++ b/bot/exts/evergreen/bookmark.py
@@ -7,14 +7,16 @@ import discord
 from discord.ext import commands
 
 from bot.bot import Bot
-from bot.constants import Colours, ERROR_REPLIES, Icons
+from bot.constants import Categories, Colours, ERROR_REPLIES, Icons, WHITELISTED_CHANNELS
 from bot.utils.converters import WrappedMessageConverter
+from bot.utils.decorators import whitelist_override
 
 log = logging.getLogger(__name__)
 
 # Number of seconds to wait for other users to bookmark the same message
 TIMEOUT = 120
 BOOKMARK_EMOJI = "📌"
+WHITELISTED_CATEGORIES = (Categories.help_in_use,)
 
 
 class Bookmark(commands.Cog):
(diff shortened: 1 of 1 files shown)


Objectivitix:
I would like to expand upon the original issue, but since that needs separate explicit approval I'll put it down here:

## Description
I believe categories such as discussion (ID: 356012535780081676) and development (ID: 411199786025484308) should also be whitelisted. The implementation would be simple: adding the two IDs into the whitelist tuple.

vcokltfre (approved):
Looks good. I'd also like to voice my support for members being able to use .bm in the other categories mentioned by Objectivix

Objectivitix:
@vcokltfre I agree. `.bm` seems like the type of command which we should allow everywhere - perhaps we could even rewrite `whitelist_override` to be able to accept a boolean parameter `everywhere` which, when set to `True`, makes the command available everywhere?

Akarys42:
Hello, thanks for the PR @Objectivitix! We agreed to allow the bookmark command in every channel, please go forth and obliterate the check! :shipit:

Objectivitix:
Alright, thanks @Akarys42! I'm thinking of implementing an `everywhere` parameter to facilitate future use if another command like this needs to be completely whitelisted - I just need to find the part where the bot checks the channel and sends the rejection embed so I can add a new condition :D

wookie184:
I sort of think that if we're thinking of allowing it everywhere that's an indicator it would be better as a feature for Python bot than sir-lancebot, maybe not worth moving though.

RE the `everywhere` parameter, would adding the `everyone` role to the `whitelist_override` be equivalent to that?

ChrisLovering (approved):
Looks good, nice easy fix 🎉

minalike:
@ChrisLovering I tried running `.bm` in an occupied help channel today but I got the message that I had to use #sir-lancebot-playground. Is that expected behavior because I wasn't the user that claimed the channel?

Objectivitix:
@minalike It was an error on my part; the decorators' order was inversed. A new PR is coming up right now as we speak.

PR #820, opened 2021-08-27
Escape markdown properly in post body to fix rendering on android

### LINKED ISSUE #746: Escape square brackets in reddit post bodies
On mobile (Android, not sure about iOS), an opening bracket without a matching closing bracket causes the Markdown renderer to interpret everything in between as a link title. The mobile renderer ostensibly uses the earliest opening bracket it finds while the desktop renderer uses the latest. It renders as a link because later on, there is link Markdown which is used for the following post's title. This is what that looks like on mobile:

bild

On desktop, it's fine:

bild

They should be escaped similarly to how it was done in python-discord/bot#1108. They can't actually be escaped, so they must be replaced with some similar characters.

wookie184:
## Relevant Issues
Closes #746 

## Description

Some of the results from the command aren't that beautiful, but I think any other changes aren't really within the scope of this PR and are difficult to do anyway because of how unreliable discord markdown rendering is.

## Did you:


ChrisLovering (approved):
One day rendering will be consistent across platforms, but today is not that day!

PR #831, opened 2021-08-30
Only send articles to user - Closes Issue#828

### LINKED ISSUE #828: Real Python command lists more than articles
## Description
When using the Real Python command and inspecting the links, you'll find that some links are courses, lessons, or quizes.

## Steps to Reproduce
Use the `.rp` command with `concurrency` as search words.

## Expected Behaviour
The point of the Real Python command is to provide articles, it should only bring up articles per its purpose.

## Actual Behaviour
Currently you may receive other medias that Real Python produces, which is unwanted.

## Possible Solutions
The fix may be pretty easy, we should be filtering out any articles where the URL is `BASE/courses/*`, `BASE/quizzes/*`, `BASE/lessons/*`.

In my opinion I don't think we should be pulling more articles, as the further down we go the less accurate the search will be.

All articles are under the `BASE/*` URL so it should be easy to only exclude the non-article content.

brad90four:
Added the "kind": "article" parameter to the API request. This should only return articles to the user instead of courses, lessons or quizzes.

Also updated the "Here are the top x results" message to handle a variable amount of articles.
[Ticket: python-discord#828]

## Relevant Issues

Closes #828

## Did you:

### DIFF
--- a/bot/exts/evergreen/realpython.py
+++ b/bot/exts/evergreen/realpython.py
@@ -53,10 +53,15 @@ class RealPython(commands.Cog):
             await ctx.send(embed=no_articles)
             return
 
+        if len(articles) == 1:
+            article_description = "Here is the result:"
+        else:
+            article_description = f"Here are the top {len(articles)} results:"
+
         article_embed = Embed(
             title="Search results - Real Python",
             url=SEARCH_URL.format(user_search=quote_plus(user_search)),
-            description=f"Here are the top {len(articles)} results:",
+            description=article_description,
             color=Colours.orange,
         )
 
--- a/bot/exts/evergreen/realpython.py
+++ b/bot/exts/evergreen/realpython.py
@@ -33,7 +33,7 @@ class RealPython(commands.Cog):
     @commands.cooldown(1, 10, commands.cooldowns.BucketType.user)
     async def realpython(self, ctx: commands.Context, *, user_search: str) -> None:
         """Send 5 articles that match the user's search terms."""
-        params = {"q": user_search, "limit": 5}
+        params = {"q": user_search, "limit": 5, "kind": "article"}
         async with self.bot.http_session.get(url=API_ROOT, params=params) as response:
             if response.status != 200:
                 logger.error(
@@ -56,7 +56,7 @@ class RealPython(commands.Cog):
         article_embed = Embed(
             title="Search results - Real Python",
             url=SEARCH_URL.format(user_search=quote_plus(user_search)),
-            description="Here are the top 5 results:",
+            description=f"Here are the top {max(5, len(articles))} results:",
             color=Colours.orange,
         )
 
--- a/bot/exts/evergreen/realpython.py
+++ b/bot/exts/evergreen/realpython.py
@@ -56,7 +56,7 @@ class RealPython(commands.Cog):
         article_embed = Embed(
             title="Search results - Real Python",
             url=SEARCH_URL.format(user_search=quote_plus(user_search)),
-            description=f"Here are the top {max(5, len(articles))} results:",
+            description=f"Here are the top {len(articles)} results:",
             color=Colours.orange,
         )
 


bot/exts/evergreen/realpython.py
```diff
@@ -56,7 +56,7 @@ async def realpython(self, ctx: commands.Context, *, user_search: str) -> None:
         article_embed = Embed(
             title="Search results - Real Python",
             url=SEARCH_URL.format(user_search=quote_plus(user_search)),
-            description="Here are the top 5 results:",
+            description=f"Here are the top {max(5, len(articles))} results:",
```
wookie184, line 59: This could actually just be 
```suggestion
            description=f"Here are the top {len(articles)} results:",
```
as we're limiting the number returned elsewhere. This change would save us having to update the code in multiple places if we decide to change the number returned

wookie184 (commented):
Thanks for the PR, this is a good change :)

Just one minor comment

wookie184 (approved):
LGTM

bot/exts/evergreen/realpython.py
```diff
@@ -56,7 +56,7 @@ async def realpython(self, ctx: commands.Context, *, user_search: str) -> None:
         article_embed = Embed(
             title="Search results - Real Python",
             url=SEARCH_URL.format(user_search=quote_plus(user_search)),
-            description="Here are the top 5 results:",
+            description=f"Here are the top {len(articles)} results:",
```
ventaquil, line 59: Special case for `len(articles) == 1`?

```suggestion
            description=f"Here are the top {len(articles)} results:" if len(articles) > 1 else 'Here is only one result:',
```

ventaquil (changes requested):

bot/exts/evergreen/realpython.py
brad90four: It is over the 120 character limit for the code style, and I'm not sure how to break / wrap that inline if / else statement. Also I would use " instead of '

Would having a dedicated if / else just for the article_embed be too bulky?

brad90four:
```py
        if len(articles) == 1:
            article_embed = Embed(
                title="Search result - Real Python",
                url=SEARCH_URL.format(user_search=quote_plus(user_search)),
                description=f"Here is the result:",
                color=Colours.orange,
            )
            
        article_embed = Embed(
            title="Search results - Real Python",
            url=SEARCH_URL.format(user_search=quote_plus(user_search)),
            description=f"Here are the top {len(articles)} results:",
            color=Colours.orange,
        )
```

ventaquil: This was just a suggestion 👍 
You can also do something similar to the following code.

```py
        if len(articles) == 1:
            description="Here is the result:"
        else:
            description=f"Here are the top {len(articles)} results:"
            
        article_embed = Embed(
            title="Search results - Real Python",
            url=SEARCH_URL.format(user_search=quote_plus(user_search)),
            description=description,
            color=Colours.orange,
        )
```

brad90four: Yeah that actually looks way better and makes more sense. Are you able to make that in your suggested change? Then I can commit it directly from your comment.

ventaquil: I will try :+1:

bot/exts/evergreen/realpython.py
```diff
@@ -56,7 +56,7 @@ async def realpython(self, ctx: commands.Context, *, user_search: str) -> None:
...
             color=Colours.orange,
         )
```
ventaquil, line 61:
```suggestion
        if len(articles) == 1:
            description = "Here is the result:"
        else:
            description = f"Here are the top {len(articles)} results:"

        article_embed = Embed(
            title="Search results - Real Python",
            url=SEARCH_URL.format(user_search=quote_plus(user_search)),
            description=description,
            color=Colours.orange,
        )
```

ventaquil (changes requested):

bot/exts/evergreen/realpython.py
ventaquil: GitHub doesn't want to cooperate 👀

brad90four: LGTM

Bluenix2: image

It is better if @brad90four simply makes the commit locally.

ventaquil: @Bluenix2 Yeah, this is what I was talking about GitHub and cooperation :eyes:

ventaquil (approved):
Look awesome!

Bluenix2 (approved):
Thank you

brad90four:
> 
> 
> Thank you

No, thank you! Everyone who has submitted a requested change or reviewed has been awesome.

ChrisLovering (approved):
Looks good, nice one 👍

PR #610, opened 2021-03-03
Teapot support

sam-heller:
## Relevant Issues
Closes #428 
Closes #500
Closes #608

## Description
Removed HTTPStatus from the equation entirely in favor of just relying on the cat/dog servers to send us back errors if we get bad input. 

Minor refactor to pull strings out of the executable code into variables at the top. 

## Reasoning
Implemented design and strategy suggested by staff in #608 discussion. 

## Screenshots
Screen Shot 2021-03-03 at 8 16 13 AM
Screen Shot 2021-03-03 at 8 16 18 AM

## Did you:

- [X] Join the [**Python Discord Community**](https://discord.gg/python)?
- [X] If dependencies have been added or updated, run `pipenv lock`?
- [X] **Lint your code** (`pipenv run lint`)?
- [ ] Set the PR to **allow edits from contributors**? *Edits allowed from maintainers, contributors doesn't appear to be an option*

PR #833, opened 2021-08-31
Teapot support - Continues PR #610

### LINKED ISSUE #428: Statuscat 418 doesn't work
## Description
Http code 418 is reported to be invalid, even though its defined [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418). A [cat](https://http.cat/418.jpg) exists for the status code too.

## Steps to Reproduce
Send `.statuscat 418` to Seasonalbot.

## Expected Behaviour
Sends an embed with [this](https://http.cat/418.jpg) image in it, and a header of `Status: 418`.

## Actual Behaviour
Code not found:
image

## Known Impacted Platforms

- [x] Web
- [x] Desktop
- [x] Android App
- [x] iOS App

## Possible Solutions
Check for status code 418 before checking for the other codes the http library.

### LINKED ISSUE #500: httpcat missing codes
## Description

The `.http_cat` command is supposed to send the picture associated with the status code the user enters. To check if the user has entered a valid status code, it checks if it is in the [`HTTPStatus` enum](https://docs.python.org/3/library/http.html#http.HTTPStatus).  The problem with this, is that three of the status codes, `103 EARLY_HINTS`, `418 IM_A_TEAPOT` and `425 TOO_EARLY`, are new to python version 3.9, though they are in [http.cat](https://http.cat/). We should fix it so that these codes can be used.

## Reasoning

More codes for the `.http_cat` command.

## Proposed Implementation

There are two ways I can think of implementing this. 

- Create a list of valid codes from  [http.cat](https://http.cat/).
- Or make a new class deriving from [`HTTPStatus`](https://docs.python.org/3/library/http.html#http.HTTPStatus) with the extra codes.

I think the first way would be easier to implement and maybe even slightly more performant.

### LINKED ISSUE #608: HTTP Status Code 418 not recognized by .http_status command
## Description
The .http_status .status and .httpstatus commands do not recognize HTTP status code 418 (I am a Teapot) as described in [RFC-2324 : Hyper Text Coffee Pot Control Protocol (HTCPCP/1.0)
](https://tools.ietf.org/html/rfc2324)

> 2.3.2 418 I'm a teapot
>   Any attempt to brew coffee with a teapot should result in the error
>   code "418 I'm a teapot". The resulting entity body MAY be short and
>   stout.

## Steps to Reproduce
Input any of the following commands in the sir-lancebot-playground channel
```
.httpstatus dog 418
.httpstatus cat 418

.status cat 418
.status dog 418

.http_status cat 418
.http_status dog 418
```

## Expected Behaviour
The relevant image from either https://http.cat/418 
http-cat
or https://httpstatusdogs.com/418-im-a-teapot

418-dogs
Should be displayed

## Actual Behaviour
An error message is displayed indicating that the status code 418 does not exist:
Screen Shot 2021-03-02 at 3 10 55 PM

## Known Impacted Platforms

- [X] Web
- [X] Desktop
- [x] Android App
- [x] iOS App

## Possible Solutions

Support for HTTP Status Code 418 Is currently only available in [Python 3.9](https://docs.python.org/3.9/library/http.html). 

Screen Shot 2021-03-02 at 3 29 43 PM

Screen Shot 2021-03-02 at 3 30 04 PM

 Implement manual workaround for this status code until default interpreter is upgraded to 3.9
## Additional Details

Relevant Python Bug : https://bugs.python.org/issue39507

- [X] I'd like to implement the bug fix

brad90four:
## Relevant Issues

Issues
Closes #428
Closes #500
Closes #608

PR
Closes #610

## Description

Initial commit, starting PR from original forked branch. 

## Did you:


vcokltfre:
There appears to be an issue where it will display the status fine, and yet still shows an error:
image
image
image

bot/exts/evergreen/status_codes.py
```diff
@@ -20,62 +22,61 @@ class HTTPStatusCodes(commands.Cog):
...
-            embed.set_footer(text="Inputted status code does not exist.")
-
-        except NotImplementedError:
-            embed.set_footer(text="Inputted status code is not implemented by http.cat yet.")
-
-        finally:
-            await ctx.send(embed=embed)
+        """Send a cat version of the requested HTTP status code."""
+        if code in range(100, 599):
+            await self.build_embed(url=HTTP_CAT_URL.format(code=code), ctx=ctx, code=code)
```
vcokltfre, line 39: You need to return after this point to prevent the error I commented about

bot/exts/evergreen/status_codes.py
```diff
@@ -20,62 +22,61 @@ class HTTPStatusCodes(commands.Cog):
...
-            embed.set_footer(text="Inputted status code does not exist.")
-
-        except NotImplementedError:
-            embed.set_footer(text="Inputted status code is not implemented by httpstatusdogs.com yet.")
-
-        finally:
-            await ctx.send(embed=embed)
+        """Send a dog version of the requested HTTP status code."""
+        if code in range(100, 599):
+            await self.build_embed(url=HTTP_CAT_URL.format(code=code), ctx=ctx, code=code)
```
vcokltfre, line 51: Same here as above.

vcokltfre (changes requested):

bot/exts/evergreen/status_codes.py
```diff
@@ -20,62 +22,63 @@ class HTTPStatusCodes(commands.Cog):
...
-        finally:
-            await ctx.send(embed=embed)
+        """Send a cat version of the requested HTTP status code."""
+        if code in range(100, 599):
+            await self.build_embed(url=HTTP_CAT_URL.format(code=code), ctx=ctx, code=code)
+            return
+        await ctx.send(
+            embed=discord.Embed(
+                title="Input status code does not exist",
+                description="The range of accepted status codes that are ok is from 100 to 599",
```
wookie184, line 44: "accepted status codes that are ok" seems a bit verbose, "The range of valid status codes is 100 to 599" might be a bit more concise. Same would apply to the other one, you could probably create a function that takes the url and runs the command to get rid of the repetition here, although having `build_embed` does remove most of the repetition so that's ok

wookie184 (changes requested):
Good stuff, just small things.

Btw, this is extremely picky so I wouldn't ask you to change it, but as python's range ends are exclusive, if for example you were looking for 2xx response codes you would check in `range(200, 300)`  rather than `range(200, 299)`. That said, 299 isn't a valid status code, so in practice this would make no difference at all!

bot/exts/evergreen/status_codes.py
```diff
@@ -20,62 +22,63 @@ class HTTPStatusCodes(commands.Cog):
...
 
     @http_status_group.command(name="dog")
     async def http_dog(self, ctx: commands.Context, code: int) -> None:
-        """Sends an embed with an image of a dog, portraying the status code."""
-        # These codes aren't server-friendly.
-        if code in (304, 422):
-            await self.http_cat(ctx, code)
+        """Send a dog version of the requested HTTP status code."""
+        if code in range(100, 599):
+            await self.build_embed(url=HTTP_CAT_URL.format(code=code), ctx=ctx, code=code)
```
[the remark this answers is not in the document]
brad90four: 🤦‍♂️ Yeah....

bot/exts/evergreen/status_codes.py
brad90four: I decided to create a standalone embed that would be sent in the case that the status code is outside of the range. I used your wording since it reads much easier.

brad90four:
> 
> 
> Good stuff, just small things.
> 
> Btw, this is extremely picky so I wouldn't ask you to change it, but as python's range ends are exclusive, if for example you were looking for 2xx response codes you would check in `range(200, 300)` rather than `range(200, 299)`. That said, 299 isn't a valid status code, so in practice this would make no difference at all!

I went ahead and changed it, doesn't hurt to keep things proper (and I am already making changes, so what's one more small one?).

bot/exts/evergreen/status_codes.py
```diff
@@ -20,62 +26,53 @@ class HTTPStatusCodes(commands.Cog):
...
+                    embed=discord.Embed(
+                        title=ERR_404.format(code=code)
+                    ).set_image(url=url)
+                )
+            else:
+                await ctx.send(
+                    embed=discord.Embed(
+                        title=STATUS_TEMPLATE.format(code=code)
+                    ).Embed.set_footer(ERR_UNKNOWN.format(code=code))
+                )
```
wookie184, line 75: This doesn't seem quite right at the moment, the `.Embed` shouldn't be that, and `ERR_UNKNOWN.format(code=code)` needs to be passed in as a keyword argument, so `set_footer(text=ERR_UNKNOWN.format(code=code))`.

What I noticed to get this result is that instead of returning 404 when the image isn't foundlike the cat url, httpstatusdogs.com seems to return a 302, meaning it isn't caught in the `elif` and gets to the else. We could probably change line 64 to check for 302 first, `elif response.status in (302, 404):`, maybe with a comment as to why this is necessary as well.

This is the result I got after fixing the `else`, catching the 302 in the `elif` should make the responses more consistent:

Discord_TDrt14y8HD

wookie184 (changes requested):
Just noticed a sort of weird thing with the statusdogs api, explained in comment. Otherwise this looks good!

bot/exts/evergreen/status_codes.py
```diff
@@ -20,62 +26,53 @@ class HTTPStatusCodes(commands.Cog):
...
+                await ctx.send(
+                    embed=discord.Embed(
+                        title=STATUS_TEMPLATE.format(code=code)
+                    ).set_image(url=url)
+                )
+            elif response.status in (302, 404):  # dog URL returns 302 instead of 404
+                await ctx.send(
+                    embed=discord.Embed(
+                        title=ERR_404.format(code=code)
+                    ).set_image(url=url)
```
wookie184, line 68: Whilst the cat api returns the 404 image if the result isn't found, meaning the URL can be used anyway, the dog api does not, meaning discord will try to use an invalid URL and fail to load the image. Explicitly using the relevant 404 image would be a way to fix this. Example to test this is `.status dog 478`

bot/exts/evergreen/status_codes.py
```diff
@@ -20,62 +26,53 @@ class HTTPStatusCodes(commands.Cog):
...
+                )
+            else:
+                await ctx.send(
+                    embed=discord.Embed(
+                        title=STATUS_TEMPLATE.format(code=code)
+                    ).Embed.set_footer(text=ERR_UNKNOWN.format(code=code))
```
wookie184, line 74: The `.Embed` before `.set_footer` on this line shouldn't be there, I don't think?

bot/exts/evergreen/status_codes.py
brad90four: Would this be the correct behavior if the status code is not found?

I hard-coded in the 404 pictures for the dog and cat urls if there is a 404 error. 

image

wookie184: That looks good to me 👍

brad90four:
Thanks for the patience guys, I think it should be working now!

wookie184 (approved):
This seems to work as expected now, thanks!

vcokltfre (approved):
Cool cool, seems to be working well!

ChrisLovering (approved):

PR #835, opened 2021-08-31
Migrate to Discord 2.0a0

Akarys42:
Since the Discord.py repository has been archived, we can switch to the latest commit of 2.0a0, knowing no breaking change will occur (still pinned to the commit just in case).

This commit also solves two small problems due to that fix, the avatar interface changing and Embed.name disappearing. Quite a painless migration.

I took the liberty of adding two QoL things:
- Adding mentions as a valid trigger for the bot
- Making the bot automatically join new threads

### DIFF
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -8,6 +8,7 @@ from aiohttp import AsyncResolver, ClientSession, TCPConnector
 from async_rediscache import RedisSession
 from discord import DiscordException, Embed
 from discord.ext import commands
+from discord.ext.commands import when_mentioned_or
 
 from bot import constants
 
@@ -203,7 +204,7 @@ loop.run_until_complete(redis_session.connect())
 
 bot = Bot(
     redis_session=redis_session,
-    command_prefix=constants.Client.prefix,
+    command_prefix=when_mentioned_or(constants.Client.prefix),
     activity=discord.Game(name=f"Commands: {constants.Client.prefix}help"),
     allowed_mentions=discord.AllowedMentions(everyone=False, roles=_allowed_roles),
     intents=_intents,
--- a/bot/bot.py
+++ b/bot/bot.py
@@ -120,7 +120,7 @@ class Bot(commands.Bot):
                 return
 
         if not icon:
-            icon = self.user.avatar_url_as(format="png")
+            icon = self.user.display_avatar.url
 
         embed = Embed(description=details)
         embed.set_author(name=title, icon_url=icon)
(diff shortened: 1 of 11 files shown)


bot/exts/evergreen/snakes/_utils.py
```diff
@@ -461,7 +461,7 @@ async def _add_player(self, user: Member) -> None:
         self.players.append(user)
         self.player_tiles[user.id] = 1
 
-        avatar_bytes = await user.avatar_url_as(format="jpeg", size=PLAYER_ICON_IMAGE_SIZE).read()
+        avatar_bytes = await user.display_avatar.read()
```
ChrisLovering, line 464: Should we still replace this with `PLAYER_ICON_IMAGE_SIZE`?

ChrisLovering (commented):
Just one comment for now, will give it a full look through & test later

bot/exts/evergreen/snakes/_utils.py
Akarys42: Indeed, indeed, forgot about that one

bot/bot.py
```diff
@@ -45,6 +47,17 @@ def member(self) -> Optional[discord.Member]:
...
+    @Cog.listener()
+    async def on_thread_join(self, thread: Thread) -> None:
+        """
+        Try to join newly created threads.
+
+        Despite the event name being misleading, this is dispatched when new threads are created.
+        We want our bots to automatically join threads in order to answer commands using their prefixes.
+        """
+        with suppress(Forbidden):
+            await thread.join()
```
ChrisLovering, line 59:
```suggestion
        if thread.me:
           # Already in this thread, return early
           return

        with suppress(Forbidden):
            await thread.join()
```
on_thread_join is triggered on creation of a thread, and when ever someone joins a thread.

Since `thread.join()` makes an API request, I think we should check if the bot is already in the thread before trying to join it again.

ChrisLovering (changes requested):
Only thing I found while doing some tests. Didn't want to spend lots of time testing every area of the bot, since issues raised from this bot aren't critical.

vcokltfre:
I've just tested about 90% of the commands, as far as I can see they are working without issue

ChrisLovering (approved):

jb3 (approved):
Proxy approval for my brother from another mother, ChrisJL.

vcokltfre (approved):

PR #849, opened 2021-09-05
`.quack`

### LINKED ISSUE #652: .quack - Procedural Duck Generator!
## Description
A command to fetch a random duck from QuackStack.

https://github.com/python-discord/quackstack

## Proposed Implementation
This command can be called in two ways:
- With `.quack`, you simply get a completely random duck. Different every time.
- By using `.quack <seed>`, you'll get a duck that's always the same if the seed is the same.

## Is this blocked?
Probably. We need to get Quackstack to a more finished state before it will be possible to do this issue. If you're wondering if this issue is blocked, check if we've completed the [Quackstack v1.0.0 milestone](https://github.com/python-discord/quackstack/milestone/1).

Kronifer:
## Relevant Issues

Closes #652 

## Description

Implemented a `.quack` command that can take a seed and optional ducky type if you want to generate a duck or manduck. If the seed given is a word, that word is converted into a seed by converting the string bytes to an integer

image
image

## Did you:


bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,65 @@
...
+        self.bot = bot
+
+    @commands.command()
+    async def quack(self, ctx: commands.Context, seed: Optional[str], ducktype: Optional[str] = None) -> None:
+        """
+        Pulls a random URL from the Quackstack API.
+
+        If a seed is provided, a duck is generated based on the given seed.
+        If a duck type is provided, either a duck or manduck will be generated.
+        """
```
wookie184, line 23: Would be good to make it clear what this does to somebody who doesn't know what the quackstack API is, something like
```suggestion
        """
        Uses the Quackstack API to generate a random duck image,

        If a seed is provided, a duck is generated based on the given seed.
        If a duck type is provided, either a duck or manduck will be generated.
        """
```

bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class Quack(commands.Cog):
+    """Cog for the .quack command."""
+
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+
+    @commands.command()
+    async def quack(self, ctx: commands.Context, seed: Optional[str], ducktype: Optional[str] = None) -> None:
```
wookie184, line 17: This current signature doesn't let the user generate a manduck without specifying a seed from what I can tell. Maybe we could swap the ducktype and seed around? Any other ideas for how to fix this?

Also wouldn't it make sense for `ducktype` to default as `duck`, not `None`

bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Pulls a random URL from the Quackstack API.
+
+        If a seed is provided, a duck is generated based on the given seed.
+        If a duck type is provided, either a duck or manduck will be generated.
+        """
+        quackstack_url = "https://quackstack.pythondiscord.com/"
+
+        image_url = "https://quackstack.pythondiscord.com"
```
wookie184, line 26: We should define the API URL once, as a constant (all caps), at the start of the file.

bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,65 @@
...
+            quackstack_url += f"?seed={seed}"
+
+        async with self.bot.http_session.get(quackstack_url) as r:
+            data = await r.json()
+            file = data.get("file")
+            image_url += file
+
+        embed = discord.Embed(
+            title=f"Quack! Here's a {ducktype.lower() if ducktype else 'duck'} for you.",
+            description=f"A {ducktype.lower() if ducktype else 'duck'} from \
```
wookie184, line 53: Should try and avoid the repetition of `ducktype.lower() if ducktype else 'duck'` here. If we make ducktype `duck` instead of `None` we could just use the value directly

bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,65 @@
...
+
+        If a seed is provided, a duck is generated based on the given seed.
+        If a duck type is provided, either a duck or manduck will be generated.
+        """
+        quackstack_url = "https://quackstack.pythondiscord.com/"
+
+        image_url = "https://quackstack.pythondiscord.com"
+
+        if not ducktype:
+            quackstack_url += "duck"
```
wookie184, line 29: if ducktype defaulted to duck not none this bit could be removed

bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,65 @@
...
+        else:
+            if ducktype.lower() not in ["duck", "manduck"]:
```
wookie184, line 31: since we always want ducktype as lowercase we could just do `ducktype = ducktype.lower()` at the start of the function

bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,65 @@
...
+            try:
+                seed = int(seed)
+            except ValueError:
+                newseed = ""
+                for e in seed:
+                    newseed += str(ord(e))
+                seed = int(newseed)
+            quackstack_url += f"?seed={seed}"
+
+        async with self.bot.http_session.get(quackstack_url) as r:
```
wookie184, line 46: I think using the `params` argument of `.get` would be preferred here to pass in the seed. You could do something like `params={}` and then in the `if seed` set `params["seed"] = seed` so it is only defined if you use it. Haven't tried that but seems like it should work

bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,65 @@
...
+            data = await r.json()
+            file = data.get("file")
+            image_url += file
+
+        embed = discord.Embed(
+            title=f"Quack! Here's a {ducktype.lower() if ducktype else 'duck'} for you.",
+            description=f"A {ducktype.lower() if ducktype else 'duck'} from \
+                [Quackstack](https://quackstack.pydis.org/docs).",
```
wookie184, line 54: It might be better to link the github page, as it should give a bit more context about what it is, although then again the README isn't really very detailed currently. Maybe it's fine to leave it as this. Thoughts?

wookie184 (changes requested):
All works well, just some minor comments on a few things. Thanks for the PR!

bot/exts/evergreen/quack.py
Kronifer: I think I'm gonna leave it like this for now

bot/exts/evergreen/quack.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        Uses the Quackstack API to generate a random duck.
+
+        If a seed is provided, a duck is generated based on the given seed.
+        If a duck type is provided, either a duck or manduck will be generated.
+        """
+        ducktype = ducktype.lower()
+        if ducktype not in ["duck", "manduck"]:
+            raise commands.BadArgument(
+                "The given duck type is invalid. The duck type can either be `duck` or `manduck`."
```
wookie184, line 29: Just noticed, as it adds on a bit to the front here it's a little repetitive, could probably remove the "The given duck type is invalid" bit 

Discord_ueMwIE1iRa

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @commands.command()
+    async def quack(self, ctx: commands.Context, ducktype: Optional[str] = "duck", seed: Optional[str] = None) -> None:
+        """
+        Uses the Quackstack API to generate a random duck.
+
+        If a seed is provided, a duck is generated based on the given seed.
+        If a duck type is provided, either a duck or manduck will be generated.
+        """
+        ducktype = ducktype.lower()
```
doublevcodes, line 26: Even though this doesn't adversely affect the function in this case, this is a case of mutating default arguments (I think?), check out `!mutable-default-args` in PyDis

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,62 @@
...
+        quackstack_url = f"{API_URL}/{ducktype}"
+        params = {}
+        if seed is not None:
+            try:
+                seed = int(seed)
+            except ValueError:
+                newseed = ""
+                for e in seed:
+                    newseed += str(ord(e))
+                seed = int(newseed)
```
doublevcodes, line 41:
```suggestion
                seed = int(''.join([str(ord(e) for e in seed]))
```
This is also a more optional thing, but it's just a more concise way to write this

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,62 @@
...
+            params["seed"] = seed
+
+        async with self.bot.http_session.get(quackstack_url, params=params) as r:
+            data = await r.json()
+            file = data.get("file")
+            image_url = f"{API_URL}{file}"
```
doublevcodes, line 47: What happens if the request fails?

doublevcodes (commented):
None of this is solid stuff that needs to necessarily happen, just a few pointers.

bot/exts/fun/quack.py
Bluenix2, line 47: This doesn't handle down-time, even though it should be little and we should know when that is. Unneccessary to blow up the console about it.

bot/exts/fun/quack.py
Bluenix2, line 41: I don't understand this, why the intermediate `str(ord())` stuff?

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,62 @@
...
+        Uses the Quackstack API to generate a random duck.
+
+        If a seed is provided, a duck is generated based on the given seed.
+        If a duck type is provided, either a duck or manduck will be generated.
+        """
+        ducktype = ducktype.lower()
+        if ducktype not in ["duck", "manduck"]:
+            raise commands.BadArgument(
+                "The duck type can only be `duck` or `manduck`."
+            )
```
Bluenix2, line 30: This could be a converter that returns an enum.

bot/exts/fun/quack.py
Bluenix2: No, strings are immutable. Besides, even if they were this would not hurt anything.

bot/exts/fun/quack.py
Numerlor: Literal should work as a converter with the new d.py

bot/exts/fun/quack.py
Kronifer: Just in case the user submits a string as a seed

Bluenix2: Right, and how does work and why did you do this rather than something else like skipping the `str()` step.

Or is this just the first-best way to get a number from a string with characters?

Kronifer: Just the best way

bot/exts/fun/quack.py
Kronifer: I- what

bot/exts/fun/quack.py
Bluenix2: Yeah so the implementation doesn't matter, can you add a comment for that for future uses?

Objectivitix: Perhaps using double quotes `"` are better, for consistency

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,63 @@
...
+        if seed is not None:
+            try:
+                seed = int(seed)
+            except ValueError:
+                # Converts any words or words with numbers into an integer using ord()
+                seed = int(''.join([str(ord(e)) for e in seed]))
+            params["seed"] = seed
+
+        try:
+            async with self.bot.http_session.get(quackstack_url, params=params) as r:
```
Bluenix2, line 44: Still doesn't handle non-200 responses, which was mostly what I thought we should handle. We don't catch `ClientConnectorError` elsewhere afaik.

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,63 @@
...
+                data = await r.json()
+                file = data.get("file")
+                image_url = f"{API_URL}{file}"
+        except client_exceptions.ClientConnectorError:
+            raise commands.CommandError("Request failed, please try again later.")
```
Bluenix2, line 49: It's a bit cramped here after you added the try/except.

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,63 @@
...
+                seed = int(seed)
+            except ValueError:
+                # Converts any words or words with numbers into an integer using ord()
+                seed = int(''.join([str(ord(e)) for e in seed]))
+            params["seed"] = seed
+
+        try:
+            async with self.bot.http_session.get(quackstack_url, params=params) as r:
+                data = await r.json()
+                file = data.get("file")
```
Bluenix2, line 46: This uses `dict.get()` implying `file` may be missing, yet you don't handle the case of it being None.

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,63 @@
...
+from discord.ext import commands
+
+from bot.bot import Bot
+from bot.constants import Colours
+
+API_URL = 'https://quackstack.pythondiscord.com'
+
+
+class Quack(commands.Cog):
+    """Cog for the .quack command."""
```
Bluenix2, line 14: Perhaps we could make it more ambiguous (name of the class included) and say that it's for wrapping Quackstack.

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,63 @@
...
+            except ValueError:
+                # Converts any words or words with numbers into an integer using ord()
+                seed = int(''.join([str(ord(e)) for e in seed]))
+            params["seed"] = seed
+
+        try:
+            async with self.bot.http_session.get(quackstack_url, params=params) as r:
+                data = await r.json()
+                file = data.get("file")
+                image_url = f"{API_URL}{file}"
```
Bluenix2, line 47:
```suggestion
                image_url = API_URL + file
```

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Uses the Quackstack API to generate a random duck.
+
+        If a seed is provided, a duck is generated based on the given seed.
+        If a duck type is provided, either a duck or manduck will be generated.
+        """
+        error_embed = discord.Embed(
+            title=random.choice(NEGATIVE_REPLIES),
+            description="The request failed. Please try again later.",
+            color=Colours.soft_red
```
Shom770, line 31: This part could use a trailing comma but it doesn't really matter, see the reply below near line 65

Shom770 (commented):
I love this command! It's pretty neat and it would be nice to have the Quackstack command exist in Python Discord itself! I have a few nitpicks here and there, but otherwise, style-wise, this looks good!

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+            )
+
+        quackstack_url = f"{API_URL}/{ducktype}"
+        params = {}
+        if seed is not None:
+            try:
+                seed = int(seed)
+            except ValueError:
+                # Converts any words or words with numbers into an integer using ord()
+                seed = int(''.join([str(ord(e)) for e in seed]))
```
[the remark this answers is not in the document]
Objectivitix: I would personally prefer ```py
seed = int(
    "".join(
        str(ord(letter) for letter in seed)
    )
)
``` (double string and generator expression for better readability). I am aware that using a genexpr is several tens of microseconds slower than using a list comprehension in this case, but that's negligable IMO and it's easier to understand without the `[]`.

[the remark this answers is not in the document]
Comment (Numerlor) [reply bot/exts/fun/quack.py:47]: `int.from_bytes` on the bytes from the encoded string should have the same behaviour with unique seeds for strings, something along the lines of `int.from_bytes(seed.encode(), "big")`, I think that's a bit less convoluted way to get it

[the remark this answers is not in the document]
Comment (Shom770) [reply bot/exts/fun/quack.py:47]: Yeah, that sounds like a better way

File: bot/exts/fun/quack.py [hunk 21]
```diff
@@ -0,0 +1,75 @@
...
+            params["seed"] = seed
+
+        try:
+            async with self.bot.http_session.get(quackstack_url, params=params) as r:
```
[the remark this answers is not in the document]
Kronifer: good idea!

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+                file = data.get("file")
+                image_url = API_URL + file
+        except client_exceptions.ClientConnectorError:
+            await ctx.send(embed=error_embed)
+            return
+
+        embed = discord.Embed(
+            title=f"Quack! Here's a {ducktype} for you.",
+            description=f"A {ducktype} from [Quackstack](https://quackstack.pydis.org/docs).",
+            color=Colours.grass_green
```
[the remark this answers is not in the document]
Kronifer: i dont tend to use trailing commas, I'm gonna leave this

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+        If a duck type is provided, either a duck or manduck will be generated.
+        """
+        error_embed = discord.Embed(
+            title=random.choice(NEGATIVE_REPLIES),
+            description="The request failed. Please try again later.",
+            color=Colours.soft_red
+        )
+
+        ducktype = ducktype.lower()
+        if ducktype not in ["duck", "manduck"]:
```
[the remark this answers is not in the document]
Kronifer: i tend to use lists, im gonna keep this the same

bot/exts/fun/quack.py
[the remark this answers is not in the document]
Kronifer: sounds better, ill do that

bot/exts/fun/quack.py
Kronifer: not using trailings

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+from bot.bot import Bot
+from bot.constants import Colours, NEGATIVE_REPLIES
+
+API_URL = 'https://quackstack.pythondiscord.com'
+
+
+class Quackstack(commands.Cog):
+    """Cog used for wrapping Quackstack."""
+
+    def __init__(self, bot: Bot) -> None:
```
Xithrius, line 17: We no longer have this annotation across this repo for the dunder of `__init__`.

```suggestion
    def __init__(self, bot: Bot):
```

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+                data = await response.json()
+                file = data.get("file")
+                image_url = API_URL + file
+        except client_exceptions.ClientConnectorError:
+            await ctx.send(embed=error_embed)
+            return
+
+        embed = discord.Embed(
+            title=f"Quack! Here's a {ducktype} for you.",
+            description=f"A {ducktype} from [Quackstack](https://quackstack.pydis.org/docs).",
```
Xithrius, line 64: Apparently `.org` and `.com` for this specific domain both go to `.com`, so we can just replace this URL with the API constant.

```suggestion
            description=f"A {ducktype} from [Quackstack]({API_URL}/docs).",
```

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+                "The duck type can only be `duck` or `manduck`."
+            )
+
+        quackstack_url = f"{API_URL}/{ducktype}"
+        params = {}
+        if seed is not None:
+            try:
+                seed = int(seed)
+            except ValueError:
+                # Converts any words or words with numbers into an integer using ord()
```
Xithrius, line 46: Is this comment talking about how `str.encode()` converts strings into integers? Either way, this isn't using the `ord()` function, but the string's `encode()` method.

```suggestion
                # Converts any word(s) with numbers into an integer
```

Xithrius (changes requested):

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class Quackstack(commands.Cog):
+    """Cog used for wrapping Quackstack."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @commands.command()
+    async def quack(self, ctx: commands.Context, ducktype: str = "duck", *, seed: Optional[str] = None) -> None:
```
brad90four, line 21:
```suggestion
    async def quack(
        self,
        ctx: commands.Context,
        ducktype: str = "duck",
        *,
        seed: Optional[str] = None,
    ) -> None:
```
Just a suggestion on whether or not to wrap the long function definition

Kronifer: Don't think so, it fits the line limit

Bluenix2, line 21:
```suggestion
    async def quack(self, ctx: commands.Context, ducktype: Literal["duck", "manduck"] = "duck", *, seed: Optional[str] = None) -> None:
```
According to @Numerlor this should work as a converter. (don't forget to import `Literal` from `typing`)

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Uses the Quackstack API to generate a random duck.
```
Bluenix2, line 23:
```suggestion
        Use the Quackstack API to generate a random duck.
```

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+
+        If a seed is provided, a duck is generated based on the given seed.
+        If a duck type is provided, either a duck or manduck will be generated.
```
Bluenix2, line 26: This makes it sound like if someone provides a duck type the command will randomly pick one of the two, not that you're supposed to send one of the two.
```suggestion
        Either "duck" or "manduck" can be provided to change the duck type generated.
```

bot/exts/fun/quack.py
Bluenix2: Can you respond to this @Kronifer?

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+            params["seed"] = seed
+
+        try:
+            async with self.bot.http_session.get(quackstack_url, params=params) as response:
+                if response.status != 200:
+                    await ctx.send(embed=error_embed)
+                    return
+                data = await response.json()
+                file = data.get("file")
+                image_url = API_URL + file
```
Bluenix2, line 57: We can just move this down to the `set_image` because we still have the `file` variable there.

bot/exts/fun/quack.py
Numerlor: This can be removed as it's now validated by the literal converter

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            async with self.bot.http_session.get(quackstack_url, params=params) as response:
+                if response.status != 200:
+                    await ctx.send(embed=error_embed)
+                    return
+                data = await response.json()
+                file = data.get("file", None)
+                if file is None:
+                    await ctx.send(embed=error_embed)
+                    return
```
Bluenix2, line 65: Still confused at this, I even looked at Quackstack's source. There's no case where `file` is None.
```suggestion
                file = data["file"]
```
But even now we can just move this down to the `set_image()` call.

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,83 @@
...
+        except client_exceptions.ClientConnectorError:
+            await ctx.send(embed=error_embed)
+            return
```
Bluenix2, line 68: I know we talked about this before, and I may have been a bit confusing which I see now. But the status code handling should be good enough.

We don't do this anywhere else and it seems unnecessary.

Kronifer: Alright, I'll remove this when I get home

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,80 @@
+import random
+from typing import Literal, Optional
+
+import discord
+from aiohttp import client_exceptions
```
Bluenix2, line 5: btw, I think you're importing a file here? `ClientConnectorError` is most likely revealed under the `aiohttp` namespace meaning `aiohttp.ClientConnectorError` should be enough.

That said this should be remove per our previous comments

bot/exts/fun/quack.py
Bluenix2: @Kronifer like @Numerlor said here, this can be removed as we have moved validation into discord.py's hands.

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,80 @@
...
+                "The duck type can only be `duck` or `manduck`."
+            )
+
+        quackstack_url = f"{API_URL}/{ducktype}"
+        params = {}
+        if seed is not None:
+            try:
+                seed = int(seed)
+            except ValueError:
+                # Converts any word(s) with numbers into an integer
```
Bluenix2, line 52:
```suggestion
                # We just need to turn the string into an integer any way possible
```
May sound better like this? Feel free to keep the old comment though.

bot/exts/fun/quack.py
Bluenix2: Bumping this as well @Kronifer

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,80 @@
...
+        Use the Quackstack API to generate a random duck.
+
+        If a seed is provided, a duck is generated based on the given seed.
+        Either "duck" or "manduck" can be provided to change the duck type generated.
+        """
+        error_embed = discord.Embed(
+            title=random.choice(NEGATIVE_REPLIES),
+            description="The request failed. Please try again later.",
+            color=Colours.soft_red,
+        )
```
Bluenix2, line 38: There should now only be used once, so we can move it down into where it's used.

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,68 @@
...
+            params["seed"] = seed
+        async with self.bot.http_session.get(quackstack_url, params=params) as response:
+            error_embed = discord.Embed(
+                title=random.choice(NEGATIVE_REPLIES),
+                description="The request failed. Please try again later.",
+                color=Colours.soft_red,
+            )
+            if response.status != 200:
+                await ctx.send(embed=error_embed)
+                return
```
D0rs4n, line 51: I believe you should move the `error_embed` creation under the if condition, since it'll be only used if the response status is not 200.

D0rs4n (changes requested):
In addition to the changes requested above, that is what I found so far. Great PR, regardless. 👍

bot/exts/fun/quack.py
Kronifer: Sounds good, I'll move once I get home

Vthechamp22 (changes requested):
Looks good!

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,68 @@
...
+                # We just need to turn the string into an integer any way possible
+                seed = int.from_bytes(seed.encode(), "big")
+            params["seed"] = seed
+        async with self.bot.http_session.get(quackstack_url, params=params) as response:
+            error_embed = discord.Embed(
+                title=random.choice(NEGATIVE_REPLIES),
+                description="The request failed. Please try again later.",
+                color=Colours.soft_red,
+            )
+            if response.status != 200:
```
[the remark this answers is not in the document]
Vthechamp22: You could also mention it in the embed so people can know the issue.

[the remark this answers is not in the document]
Kronifer: @Vthechamp we decided against catching asyncio.ClientConnectorError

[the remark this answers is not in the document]
Vthechamp22: but the status code check remains right? So I was wondering if you could log the status code for future debugging?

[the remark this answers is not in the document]
Bluenix2: Yes this check should remain, and I do agree @Kronifer should add a `log.error`. This is what the rest of the bot does.

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,72 @@
...
+        quackstack_url = f"{API_URL}/{ducktype}"
+        params = {}
+        if seed is not None:
+            try:
+                seed = int(seed)
+            except ValueError:
+                # We just need to turn the string into an integer any way possible
+                seed = int.from_bytes(seed.encode(), "big")
+            params["seed"] = seed
+        async with self.bot.http_session.get(quackstack_url, params=params) as response:
```
Bluenix2, line 46:
```suggestion
            params["seed"] = seed

        async with self.bot.http_session.get(quackstack_url, params=params) as response:
```

Bluenix2, line 46:
```suggestion
            params["seed"] = seed

        async with self.bot.http_session.get(quackstack_url, params=params) as response:
```

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,72 @@
...
+            error_embed = discord.Embed(
+                title=random.choice(NEGATIVE_REPLIES),
+                description="The request failed. Please try again later.",
+                color=Colours.soft_red,
+            )
+            if response.status != 200:
+                log.error(f"Response to Quackstack returned code {response.status}")
+                await ctx.send(embed=error_embed)
+                return
+            data = await response.json()
```
Bluenix2, line 56:
```suggestion
                return

            data = await response.json()
```

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,72 @@
...
+            file = data["file"]
+
+        embed = discord.Embed(
+            title=f"Quack! Here's a {ducktype} for you.",
+            description=f"A {ducktype} from [Quackstack]({API_URL}/docs).",
+            color=Colours.grass_green,
+        )
```
Bluenix2, line 63: Can we move it to the footer?
```suggestion
        embed = discord.Embed(
            title=f"Quack! Here's a {ducktype} for you.",
            color=Colours.grass_green,
        )
        
        embed.set_footer(text=f"A {ducktype} from [Quackstack]({API_URL}/docs).")
```

bot/exts/fun/quack.py
```diff
@@ -0,0 +1,75 @@
...
+
+            data = await response.json()
+            file = data["file"]
+
+        embed = discord.Embed(
+            title=f"Quack! Here's a {ducktype} for you.",
+            color=Colours.grass_green,
+        )
+
+        embed.set_footer(text=f"A {ducktype} from [Quackstack]({API_URL}/docs).")
```
Bluenix2, line 66: Apparently hyperlinks don't work in footers?
image
Not sure what to do here.

Kronifer: I can move it to the description again

Bluenix2: I feel like we can just use the `url=` kwarg that will lead to Quackstack (making the title "clickable").

Bluenix2: Bumping this @Kronifer as I would like to see this fixed before being merged.

Kronifer: I'll fix this very soon

Kronifer: resolved

bot/exts/fun/quack.py
D0rs4n: @Kronifer You might've forgotten about that one.

Bluenix2 (approved):
Thanks for this Kronifier!

Xithrius (approved):
Yes.

PR #855, opened 2021-09-06
CI: check license compatibility

Akarys42:
Port of https://github.com/python-discord/bot/pull/1779

---

Since our project is licensed under the MIT License, we can't be using any incompatible license-wise dependencies in our project, such as GPL.

This commit adds a step to the CI lint phase that will use pip-licenses to verify that all the installed packages are part of an ALLOWED_LICENSE variable. This variable is currently set to be every license we currently use.

We opted to use an allowlist instead of a denylist to make sure that new licenses are reviewed by hand and added to that variable.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -17,7 +17,7 @@ jobs:
     env:
       # List of licenses that are compatible with the MIT License and
       # can be used in our project
-      ALLOWED_LICENSE: Apache Software License;
+      ALLOWED_LICENSES: Apache Software License;
         BSD;  BSD License;
         GNU Library or Lesser General Public License (LGPL);
         Historical Permission Notice and Disclaimer (HPND);
@@ -79,7 +79,7 @@ jobs:
       # please make sure it is compatible with the license for this project,
       # and add it to the ALLOWED_LICENSE variable
       - name: Check Dependencies License
-        run: pip-licenses --allow-only="$ALLOWED_LICENSE"
+        run: pip-licenses --allow-only="$ALLOWED_LICENSES"
 
       # This step caches our pre-commit environment. To make sure we
       # do create a new environment when our pre-commit setup changes,
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,7 +15,19 @@ jobs:
     name: Run pre-commit & flake8
     runs-on: ubuntu-latest
     env:
-      # Configure pip to cache dependencies and do a user install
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -29,9 +29,10 @@ flake8-string-format = "~=0.3"
 flake8-tidy-imports = "~=4.1"
 flake8-todo = "~=0.7"
 pep8-naming = "~=0.11"
+pip-licenses = "~=3.5.2"
 pre-commit = "~=2.1"
-taskipy = "^1.6.0"
 python-dotenv = "^0.15.0"
+taskipy = "^1.6.0"
```
doublevcodes, line 35:
```suggestion
python-dotenv = "~=0.15"
taskipy = "~=1.6"
```
Shouldn't these be pinned like other deps for consistency, and I realise this is not related to your PR :P

doublevcodes (commented):
Looks good. 🚀

pyproject.toml
Akarys42: Perhaps, but that's is indeed out of scope :P

.github/workflows/lint.yaml
```diff
@@ -15,7 +15,19 @@ jobs:
     name: Run pre-commit & flake8
     runs-on: ubuntu-latest
     env:
-      # Configure pip to cache dependencies and do a user install
+      # List of licenses that are compatible with the MIT License and
+      # can be used in our project
+      ALLOWED_LICENSE: Apache Software License;
```
ChrisLovering, line 20:
```suggestion
      ALLOWED_LICENSES: Apache Software License;
```
😞

.github/workflows/lint.yaml
```diff
@@ -62,6 +74,13 @@ jobs:
           pip install poetry
           poetry install --no-interaction --no-ansi
 
+      # Check all the dependencies are compatible with the MIT license.
+      # If you added a new dependencies that is being rejected,
+      # please make sure it is compatible with the license for this project,
+      # and add it to the ALLOWED_LICENSE variable
+      - name: Check Dependencies License
+        run: pip-licenses --allow-only="$ALLOWED_LICENSE"
```
ChrisLovering, line 82:
```suggestion
        run: pip-licenses --allow-only="$ALLOWED_LICENSES"
```

ChrisLovering (changes requested):
Yes, I'm petty

ChrisLovering (approved):
🚀 Consider this approval whether you change the dependency format like mentioned or not

dawnofmidnight (approved):

PR #856, opened 2021-09-06
Allow `.bm` command to be used everywhere

### LINKED ISSUE #827: Allow bookmark command in every channel
## Description

As discussed in https://github.com/python-discord/sir-lancebot/pull/816 we should allow the bookmark command to be allowed in every channel.

## Reasoning

Bookmarking messages is very useful for people who want to remember things, and the command itself doesn't use up much screen real-estate, so it shouldn't cause an issue

## Proposed Implementation

Possibly add the everyone role to the `whitelist_override` but I'll leave the details up to the developer.

Objectivitix:
## Relevant Issues
Closes #827

## Description
I added a `every` parameter in `in_whitelist_check`, which when set to `True` means a command is in the whitelist. Consequently, this means setting this parameter to `True` in `whitelist_override` essentially represents that a command can be used everywhere by everyone. I did that to the `.bm` command.

This has been tested in my test server and my alt account without any roles can use the command in all channels.

## Did you:

### DIFF
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -7,7 +7,7 @@ import discord
 from discord.ext import commands
 
 from bot.bot import Bot
-from bot.constants import Categories, Colours, ERROR_REPLIES, Icons, WHITELISTED_CHANNELS
+from bot.constants import Colours, ERROR_REPLIES, Icons
 from bot.utils.converters import WrappedMessageConverter
 from bot.utils.decorators import whitelist_override
 
@@ -16,7 +16,6 @@ log = logging.getLogger(__name__)
 # Number of seconds to wait for other users to bookmark the same message
 TIMEOUT = 120
 BOOKMARK_EMOJI = "📌"
-WHITELISTED_CATEGORIES = (Categories.help_in_use,)
 
 
 class Bookmark(commands.Cog):
@@ -87,8 +86,8 @@ class Bookmark(commands.Cog):
         await message.add_reaction(BOOKMARK_EMOJI)
         return message
 
-    @whitelist_override(channels=WHITELISTED_CHANNELS, categories=WHITELISTED_CATEGORIES)
     @commands.command(name="bookmark", aliases=("bm", "pin"))
+    @whitelist_override(every=True)
     async def bookmark(
         self,
         ctx: commands.Context,
--- a/bot/utils/checks.py
+++ b/bot/utils/checks.py
@@ -40,6 +40,7 @@ def in_whitelist_check(
     channels: Container[int] = (),
     categories: Container[int] = (),
     roles: Container[int] = (),
+    every: bool = False,
     redirect: Optional[int] = constants.Channels.community_bot_commands,
     fail_silently: bool = False,
 ) -> bool:
@@ -51,6 +52,8 @@ def in_whitelist_check(
     - `channels`: a container with channel ids for whitelisted channels
     - `categories`: a container with category ids for whitelisted categories
     - `roles`: a container with with role ids for whitelisted roles
+    - `every`: a boolean value that, when set to True, allows the command
+               to be used everywhere by everyone
 
     If the command was invoked in a context that was not whitelisted, the member is either
     redirected to the `redirect` channel that was passed (default: #bot-commands) or simply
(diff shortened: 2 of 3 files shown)


Xithrius (changes requested):
The f-string prefix is not needed for lines that are just plain strings without any variables.

Lines 76, 83, 91, etc. of the `bot/utils/checks.py` file are examples of this.

Objectivitix:
@Xithrius That was my personal preference - I think string literal concatenation looks neater that way, and is easier and more organized to look at. If you insist, I could remove them.

Xithrius:
Looking through the rest of the file, to keep it consistent you'd have to remove the unnecessary f-string prefix.

TizzySaurus:
> I think string literal concatenation looks neater that way, and is easier and more organized to look at.

I could be missing something but is there even a need for the string to be split over two lines rather than just having one string? IMO splitting in this instance doesn't add anything to readability etc.

bot/utils/checks.py
```diff
@@ -67,24 +70,43 @@ def in_whitelist_check(
         # categories, it's probably not wise to rely on its category in any case.
         channels = tuple(channels) + (redirect,)
 
+    if every:
+        log.trace(
+            f"{ctx.author} may use the `{ctx.command.name}` command "
+            f"as it is completely whitelisted by every."
```
Bluenix2, line 76:
```suggestion
            f"as it is completely whitelisted by 'every'."
```
We want to make it clear that we're referring to the `every` keyword argument.

bot/utils/checks.py
```diff
@@ -67,24 +70,43 @@ def in_whitelist_check(
...
+        )
+        return True
+
     if channels and ctx.channel.id in channels:
-        log.trace(f"{ctx.author} may use the `{ctx.command.name}` command as they are in a whitelisted channel.")
+        log.trace(
+            f"{ctx.author} may use the `{ctx.command.name}` command "
+            f"as they are in a whitelisted channel."
+        )
```
Bluenix2, line 84: Why did you make these changes? The previous code wasn't longer than the line limit?

bot/exts/utilities/bookmark.py
```diff
@@ -87,8 +86,8 @@ async def send_reaction_embed(
         await message.add_reaction(BOOKMARK_EMOJI)
         return message
 
-    @whitelist_override(channels=WHITELISTED_CHANNELS, categories=WHITELISTED_CATEGORIES)
     @commands.command(name="bookmark", aliases=("bm", "pin"))
```
Kronifer, line 89: missing a trailing comma in aliases tuple

bot/utils/checks.py
```diff
@@ -67,24 +70,43 @@ def in_whitelist_check(
         # categories, it's probably not wise to rely on its category in any case.
         channels = tuple(channels) + (redirect,)
 
+    if every:
+        log.trace(
+            f"{ctx.author} may use the `{ctx.command.name}` command "
```
Kronifer, line 75: why not use a triple-quoted string?

bot/utils/checks.py
```diff
@@ -67,24 +70,43 @@ def in_whitelist_check(
...
+            f"as it is completely whitelisted by every."
+        )
+        return True
+
     if channels and ctx.channel.id in channels:
-        log.trace(f"{ctx.author} may use the `{ctx.command.name}` command as they are in a whitelisted channel.")
+        log.trace(
+            f"{ctx.author} may use the `{ctx.command.name}` command "
```
Kronifer, line 82: View above comment

bot/utils/checks.py
```diff
@@ -67,24 +70,43 @@ def in_whitelist_check(
...
+            f"as they are in a whitelisted channel."
+        )
         return True
 
     # Only check the category id if we have a category whitelist and the channel has a `category_id`
     if categories and hasattr(ctx.channel, "category_id") and ctx.channel.category_id in categories:
-        log.trace(f"{ctx.author} may use the `{ctx.command.name}` command as they are in a whitelisted category.")
+        log.trace(
```
Kronifer, line 89: Just keep this in mind for all `log.trace()` calls.

bot/utils/decorators.py
```diff
@@ -280,7 +280,10 @@ def predicate(ctx: Context) -> bool:
     return predicate
 
 
-def whitelist_override(bypass_defaults: bool = False, **kwargs: Container[int]) -> Callable:
+def whitelist_override(
```
Kronifer, line 283: Trailing comma?

Kronifer (changes requested):
Few nitpicks

bot/utils/checks.py
TizzySaurus: @Kronifer If you look at the above review by Bluenix and my comment you'll see that these should in fact be one-line, so recommending usage of triple-quote string isn't correct yet.

bot/exts/utilities/bookmark.py
TizzySaurus: This shouldn't have a trailing comma. You only use a trailing comma here when there's one alias (to make it a tuple).

bot/utils/decorators.py
TizzySaurus: Likewise shouldn't have a trailing comma here.

brad90four (approved):
LGTM

ChrisLovering (changes requested):
Rather than implementing all of this, why not use `@whitelist_override(roles=(Roles.everyone,))`?

You'd just need to add the everyone role to constants, the ID of the everyone role in pydis is 267624335836053506

Objectivitix:
First of all, sorry that I haven't been able to produce any updates lately. School just started for me two weeks ago, and I have some things to sort out, so it's going to take me some time to get my developing and studying schedule back on track. And also, @ChrisLovering, I just realized: the everyone role didn't work for me before because I placed the decorators in the wrong order - and I assumed that it was the role's fault. Now, looking at the logic, you're correct. I will update this PR soon to use the new role implementation.

Objectivitix:
But I do have a counter-argument: adding an `every` does seem more semantically precise - it's not immediately obvious what `roles=(Roles.everyone,)` should do, as I was confused as well when I started looking at the code.

Bluenix2:
I disagree, it is rather clear that you allow anyone with the everyone role to use the command. Which also overrides any channel permissions..

If that still is a worry, add a comment!

Bluenix2 (changes requested):
Please undo the changes you did to some of the logging in `checks.py`.

Objectivitix:
A new PR will soon be made implementing the change in an easier way (roles). This one will be closed.

bot/exts/utilities/bookmark.py
Objectivitix: Trailing commas are only for multiline.

PR #857, opened 2021-09-06
Add support to query AoC results in respect of days and stars

### LINKED ISSUE #321: AoC: Leaderboard per star
We currently show two types of statistics for the participants of Advent of Code:

- The leaderboard; a cumulative rundown of each players' progress
- The number and percentage of users who has completed each star

Some more granular statistics would be cool to see. I suggest an alternative leaderboard, which shows scores for a specific star only, and times instead of score (like the [daily breakdown of the global leaderboard](https://adventofcode.com/2019/leaderboard/day/1)). 

For example, `.aoc leaderboard 2-1` to get a top-10 (or optionally a different length) list for the 2nd day's 1st star:

> Top 5 solvers for star 1, day 2:
06:02 Ikusaba-san
07:44 WillDaSilva
08:46 smartspot2
09:50 Sedsarq
10:11 salt-die

The timestamps per user and star is stored in the [JSON](https://adventofcode.com/2019/leaderboard/private/view/631135.json) we grab from the site, but it looks like we're currently not using them.

I'll refrain from providing implementation details, being the rookie that I am - don't want to make bad suggestions!

D0rs4n:
## Relevant Issues
Closes #321 
## Description
From now on the AoC leaderboard command accepts a total of 2 optional arguments a day and star string (eg.: 1-2, for the second star of the first day) and a number of results they would like to see, with a total maximum of 15. (Default is 10)
(The reason I used csharp syntax highlighting is because it surprisingly fits the results quite well)
alt text
## Did you:

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -88,6 +88,7 @@ class AdventOfCode:
     ignored_days = environ.get("AOC_IGNORED_DAYS", "").split(",")
     leaderboard_displayed_members = 10
     leaderboard_cache_expiry_seconds = 1800
+    max_dayandstar_results = 15
     year = int(environ.get("AOC_YEAR", datetime.utcnow().year))
     role_id = int(environ.get("AOC_ROLE_ID", 518565788744024082))
 
--- a/bot/exts/events/advent_of_code/_cog.py
+++ b/bot/exts/events/advent_of_code/_cog.py
@@ -2,6 +2,7 @@ import json
 import logging
 from datetime import datetime, timedelta
 from pathlib import Path
+from typing import Optional
 
 import arrow
 import discord
@@ -150,7 +151,7 @@ class AdventOfCode(commands.Cog):
         else:
             try:
                 join_code = await _helpers.get_public_join_code(author)
-            except _helpers.FetchingLeaderboardFailed:
+            except _helpers.FetchingLeaderboardFailedError:
                 await ctx.send(":x: Failed to get join code! Notified maintainers.")
                 return
 
@@ -185,15 +186,47 @@ class AdventOfCode(commands.Cog):
         brief="Get a snapshot of the PyDis private AoC leaderboard",
     )
     @whitelist_override(channels=AOC_WHITELIST_RESTRICTED)
-    async def aoc_leaderboard(self, ctx: commands.Context) -> None:
-        """Get the current top scorers of the Python Discord Leaderboard."""
+    async def aoc_leaderboard(
+            self,
+            ctx: commands.Context,
+            dayandstar: Optional[str] = None,
+            maximum_scorers: Optional[int] = 10) -> None:
+        """
+        Get the current top scorers of the Python Discord Leaderboard.
+
+        Additionally, you can provide a day-star(eg.: 2-1, as second day first star) to filter your query more.
+        """
+        if maximum_scorers > AocConfig.max_dayandstar_results or maximum_scorers <= 0:
+            raise commands.BadArgument(
+                f"Unfortunately, \n the maximum number of results you can query is {AocConfig.max_dayandstar_results} "
+                "\n and the minimum number is 1"
+            )
         async with ctx.typing():
             try:
                 leaderboard = await _helpers.fetch_leaderboard()
-            except _helpers.FetchingLeaderboardFailed:
+            except _helpers.FetchingLeaderboardFailedError:
                 await ctx.send(":x: Unable to fetch leaderboard!")
                 return
-
+            if dayandstar:
+                # Fetches a dictionary that contains solvers in respect of day, and star.
+                # eg.: 1-1 will fetch the solvers of the first star of the first day and their completion time
+                per_dayandstar = json.loads(leaderboard['leaderboard_per_dayandstar'])
+                try:
+                    scorers = per_dayandstar[dayandstar][:maximum_scorers]
+                except KeyError:
+                    raise commands.BadArgument(
+                        "You provided an invalid \"day-star\" formula! "
+                        "Please use the `day-star` notation. eg.: \"1-2\". \n"
+                        "Also keep in mind that there are only two stars each day, and days are from 1 to 25"
+                    )
+                day, star = dayandstar.split("-")
+                codeblock = f"```csharp\nTop scorers for day {day}, star {star} in UTC timezone \n"
+                for scorer in scorers:
+                    time_data = datetime.fromtimestamp(scorer['completion_time']).strftime("%b.%d %I:%M %p")
+                    codeblock += f"[{scorer['member_name']}] {time_data}  \n"
+                codeblock += "```"
+                await ctx.send(codeblock)
+                return
             number_of_participants = leaderboard["number_of_participants"]
 
             top_count = min(AocConfig.leaderboard_displayed_members, number_of_participants)
@@ -231,7 +264,7 @@ class AdventOfCode(commands.Cog):
         """Send an embed with daily completion statistics for the Python Discord leaderboard."""
         try:
             leaderboard = await _helpers.fetch_leaderboard()
-        except _helpers.FetchingLeaderboardFailed:
+        except _helpers.FetchingLeaderboardFailedError:
             await ctx.send(":x: Can't fetch leaderboard for stats right now!")
             return
 
@@ -267,7 +300,7 @@ class AdventOfCode(commands.Cog):
         async with ctx.typing():
             try:
                 await _helpers.fetch_leaderboard(invalidate_cache=True)
-            except _helpers.FetchingLeaderboardFailed:
+            except _helpers.FetchingLeaderboardFailedError:
                 await ctx.send(":x: Something went wrong while trying to refresh the cache!")
             else:
                 await ctx.send("\N{OK Hand Sign} Refreshed leaderboard cache!")
--- a/bot/exts/events/advent_of_code/_helpers.py
+++ b/bot/exts/events/advent_of_code/_helpers.py
@@ -151,7 +151,22 @@ def _parse_raw_leaderboard_data(raw_leaderboard_data: dict) -> dict:
         # this data to JSON in order to cache it in Redis.
         daily_stats[day] = {"star_one": star_one, "star_two": star_two}
 
-    return {"daily_stats": daily_stats, "leaderboard": sorted_leaderboard}
+        perdaystar_stats = collections.defaultdict(list)
+        for member in raw_leaderboard_data.values():
+            username = member["name"] if member["name"] else f"Anonymous #{member['id']}"
+
+            for day_data, stars_data in member["completion_day_level"].items():
+                # Iterate over the complete stars for this day for this participant
+                for star, data in stars_data.items():
+                    # Record completion of this star for this individual
+                    completion_time = int(data["get_star_ts"])
+                    perdaystar_stats[f"{day_data}-{star}"].append(
+                        {'completion_time': completion_time, 'member_name': username}
+                    )
+        for key in perdaystar_stats.keys():
+            perdaystar_stats[key] = sorted(perdaystar_stats[key], key=operator.itemgetter('completion_time'))
+
+        return {"daily_stats": daily_stats, "leaderboard": sorted_leaderboard, 'per_dayandstar': perdaystar_stats}
 
 
 def _format_leaderboard(leaderboard: dict[str, dict]) -> str:
@@ -289,6 +304,7 @@ async def fetch_leaderboard(invalidate_cache: bool = False) -> dict:
             "leaderboard_fetched_at": leaderboard_fetched_at,
             "number_of_participants": number_of_participants,
             "daily_stats": json.dumps(parsed_leaderboard_data["daily_stats"]),
+            "leaderboard_per_dayandstar": json.dumps(parsed_leaderboard_data["per_dayandstar"])
         }
 
         # Store the new values in Redis


bot/constants.py
```diff
@@ -88,6 +88,7 @@ class AdventOfCode:
     ignored_days = environ.get("AOC_IGNORED_DAYS", "").split(",")
     leaderboard_displayed_members = 10
     leaderboard_cache_expiry_seconds = 1800
+    max_dayandstar_results = 15
```
Xithrius, line 91: Should we have variables that contain `dayandstar` be separated into `day_and_star`?

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
-        """Get the current top scorers of the Python Discord Leaderboard."""
+    async def aoc_leaderboard(
+            self,
+            ctx: commands.Context,
+            dayandstar: Optional[str] = None,
+            maximum_scorers: Optional[int] = 10) -> None:
+        """
+        Get the current top scorers of the Python Discord Leaderboard.
+
+        Additionally, you can provide a day-star(eg.: 2-1, as second day first star) to filter your query more.
```
Xithrius, line 197:
```suggestion
        Additionally, you can provide a day-star (e.g. 2-1, as second day first star) to filter your query more.
```

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+        """
+        if maximum_scorers > AocConfig.max_dayandstar_results or maximum_scorers <= 0:
+            raise commands.BadArgument(
+                f"Unfortunately, \n the maximum number of results you can query is {AocConfig.max_dayandstar_results} "
+                "\n and the minimum number is 1"
+            )
```
Xithrius, line 203: I don't think the newlines are necessary here, and the minimum intuitively makes sense in this context so it doesn't need to be stated.

```suggestion
            raise commands.BadArgument(
                f"Unfortunately, the maximum number of results you can query is {AocConfig.max_dayandstar_results}"
            )
```

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
             try:
                 leaderboard = await _helpers.fetch_leaderboard()
-            except _helpers.FetchingLeaderboardFailed:
+            except _helpers.FetchingLeaderboardFailedError:
                 await ctx.send(":x: Unable to fetch leaderboard!")
                 return
-
+            if dayandstar:
+                # Fetches a dictionary that contains solvers in respect of day, and star.
+                # eg.: 1-1 will fetch the solvers of the first star of the first day and their completion time
```
Xithrius, line 212:
```suggestion
                # e.g. 1-1 will fetch the solvers of the first star of the first day and their completion time
```

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+                per_dayandstar = json.loads(leaderboard['leaderboard_per_dayandstar'])
+                try:
+                    scorers = per_dayandstar[dayandstar][:maximum_scorers]
+                except KeyError:
+                    raise commands.BadArgument(
+                        "You provided an invalid \"day-star\" formula! "
+                        "Please use the `day-star` notation. eg.: \"1-2\". \n"
+                        "Also keep in mind that there are only two stars each day, and days are from 1 to 25"
+                    )
```
Xithrius, line 221:
```suggestion
                    raise commands.BadArgument(
                        "Day and star format has a notation of 'day-star' day being 1 to 25, and star is either 1 or 2 (e.g. 3-1)"
                    )
```

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+                day, star = dayandstar.split("-")
+                codeblock = f"```csharp\nTop scorers for day {day}, star {star} in UTC timezone \n"
+                for scorer in scorers:
+                    time_data = datetime.fromtimestamp(scorer['completion_time']).strftime("%b.%d %I:%M %p")
+                    codeblock += f"[{scorer['member_name']}] {time_data}  \n"
```
Xithrius, line 226:
```suggestion
                codeblock = f"```csharp\nTop scorers for day {day}, star {star} in UTC timezone\n"
                for scorer in scorers:
                    time_data = datetime.fromtimestamp(scorer['completion_time']).strftime("%b.%d %I:%M %p")
                    codeblock += f"[{scorer['member_name']}] {time_data}\n"
```

Xithrius (changes requested):
This is all I have for now.

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
         brief="Get a snapshot of the PyDis private AoC leaderboard",
     )
     @whitelist_override(channels=AOC_WHITELIST_RESTRICTED)
-    async def aoc_leaderboard(self, ctx: commands.Context) -> None:
-        """Get the current top scorers of the Python Discord Leaderboard."""
+    async def aoc_leaderboard(
+            self,
+            ctx: commands.Context,
+            dayandstar: Optional[str] = None,
+            maximum_scorers: Optional[int] = 10) -> None:
```
brad90four, line 193:
```suggestion
            maximum_scorers: Optional[int] = 10,
        ) -> None:
```
Not sure if I am mis-remembering the code style, but should the closing of the function be on its own line?

brad90four (approved):
Looks solid to me, just one minor thing I thought I should point out.

bot/constants.py
D0rs4n: I believe so, yes.

bot/exts/events/advent_of_code/_cog.py
D0rs4n: Yes, It would make sense I believe. Thanks for pointing out, I totally missed that one. 👍

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+    ) -> None:
+        """
+        Get the current top scorers of the Python Discord Leaderboard.
+
+        Additionally, you can provide a day-star(e.g. 2-1, as second day first star) to filter your query more.
+        """
+        if maximum_scorers > AocConfig.max_day_and_star_results or maximum_scorers <= 0:
+            raise commands.BadArgument(
+                "Unfortunately, you reached a limit. "
+                f"(The maximum number of results you can query is {AocConfig.max_day_and_star_results})"
```
Bluenix2, line 203:
```suggestion
                f"The maximum number of results you can query is {AocConfig.max_day_and_star_results}."
```

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
             try:
                 leaderboard = await _helpers.fetch_leaderboard()
-            except _helpers.FetchingLeaderboardFailed:
+            except _helpers.FetchingLeaderboardFailedError:
                 await ctx.send(":x: Unable to fetch leaderboard!")
                 return
-
+            if day_and_star:
+                # Fetches a dictionary that contains solvers in respect of day, and star.
+                # e.g. 1-1 will fetch the solvers of the first star of the first day and their completion time
```
Bluenix2, line 213:
```suggestion
                # This is a dictionary that contains solvers in respect of day, and star.
                # e.g. 1-1 will fetch the solvers of the first star of the first day and their completion time
```

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+                        "2 (e.g. 3-1) "
+                    )
+                day, star = day_and_star.split("-")
+                codeblock = f"```csharp\nTop scorers for day {day}, star {star} in UTC timezone\n"
+                for scorer in scorers:
+                    time_data = datetime.fromtimestamp(scorer['completion_time']).strftime("%b.%d %I:%M %p")
+                    codeblock += f"[{scorer['member_name']}] {time_data}\n"
+                codeblock += "```"
+                await ctx.send(codeblock)
+                return
```
Bluenix2, line 229: It's a pandemic and these lines of code are not 6 feet apart!
```suggestion
                    )

                day, star = day_and_star.split("-")
                codeblock = f"```csharp\nTop scorers for day {day}, star {star} in UTC timezone\n"

                for scorer in scorers:
                    time_data = datetime.fromtimestamp(scorer['completion_time']).strftime("%b.%d %I:%M %p")
                    codeblock += f"[{scorer['member_name']}] {time_data}\n"

                codeblock += "```"
                await ctx.send(codeblock)
                return
```

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,15 +186,47 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+            )
         async with ctx.typing():
             try:
                 leaderboard = await _helpers.fetch_leaderboard()
-            except _helpers.FetchingLeaderboardFailed:
+            except _helpers.FetchingLeaderboardFailedError:
                 await ctx.send(":x: Unable to fetch leaderboard!")
                 return
-
+            if day_and_star:
```
Bluenix2, line 211: Because of how `day_and_star` has more code than the "else", can we flip this? So `if not day_and_star` and move the code outside the if-statement into it and vice versa.

D0rs4n: Of course, I missed that one! 👍

D0rs4n:
When implementing this, I realised that this kind of day-and-star "formula" might not be the most convenient if someone wants to check multiple stars,days. So I went ahead an implemented a small panel, so that after invoking the command with something like: "1-1" you can step the day and star parameters and get new results accordingly. Also, I think it should be included in this PR that's what I proposed my idea here.
PReview

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,24 +186,61 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+            try:
+                scorers = per_day_and_star[day_and_star][:maximum_scorers]
+            except KeyError:
+                raise commands.BadArgument(
+                    "Day and star format has a notation of 'day-star' day being 1 to 25, and star is either 1 or "
+                    "2 (e.g. 3-1) "
+                )
+
+            day, star = day_and_star.split("-")
+            codeblock = f"```csharp\nTop scorers for day {day}, star {star} in UTC timezone\n"
```
Vthechamp22, line 235: What exactly does the `csharp` bit do here? What importance does the coloring for csharp have?

Vthechamp22 (changes requested):
Here's what I could find at a quick glance, but other than that it looks good!

bot/exts/events/advent_of_code/_cog.py
D0rs4n: It highlights the data in the codeblock, and I find it a perfect fit for this case. Take a look at the image sent below.

bot/exts/events/advent_of_code/_helpers.py
```diff
@@ -151,7 +151,22 @@ def _parse_raw_leaderboard_data(raw_leaderboard_data: dict) -> dict:
...
+            username = member["name"] if member["name"] else f"Anonymous #{member['id']}"
+
+            for day_data, stars_data in member["completion_day_level"].items():
+                # Iterate over the complete stars for this day for this participant
+                for star, data in stars_data.items():
+                    # Record completion of this star for this individual
+                    completion_time = int(data["get_star_ts"])
+                    per_day_star_stats[f"{day_data}-{star}"].append(
+                        {'completion_time': completion_time, 'member_name': username}
+                    )
```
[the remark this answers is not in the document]
D0rs4n: I believe it can be. Initially I wanted to modify _StarResults_ but then decided to go with that and forgot that It could be solved there. Good catch! Thank you!

bot/exts/events/advent_of_code/_cog.py
Vthechamp22: I see. all right lgtm then!

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,24 +186,61 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+
+            for scorer in scorers:
+                time_data = datetime.fromtimestamp(scorer['completion_time']).strftime("%b.%d %I:%M %p")
+                codeblock += f"[{scorer['member_name']}] {time_data}\n"
+
+            codeblock += "```"
+            await ctx.send(codeblock)
```
Bluenix2, line 242:
```suggestion
            await ctx.send(codeblock + "```")
```

bot/exts/events/advent_of_code/_helpers.py
```diff
@@ -122,6 +123,12 @@ def _parse_raw_leaderboard_data(raw_leaderboard_data: dict) -> dict:
                 star_results[(day, star)].append(
                     StarResult(member_id=member_id, completion_time=completion_time)
                 )
+                completion_time_raw = int(data["get_star_ts"])
+                per_day_star_stats[f"{day}-{star}"].append(
+                    {'completion_time': completion_time_raw, 'member_name': name}
+                )
+    for key in per_day_star_stats.keys():
```
Bluenix2, line 130:
```suggestion
    for key in per_day_star_stats:
```

bot/exts/events/advent_of_code/_helpers.py
```diff
@@ -151,7 +158,7 @@ def _parse_raw_leaderboard_data(raw_leaderboard_data: dict) -> dict:
         # this data to JSON in order to cache it in Redis.
         daily_stats[day] = {"star_one": star_one, "star_two": star_two}
 
-    return {"daily_stats": daily_stats, "leaderboard": sorted_leaderboard}
+        return {"daily_stats": daily_stats, "leaderboard": sorted_leaderboard, 'per_day_and_star': per_day_star_stats}
```
Bluenix2, line 161: Was this intended? This will now return after the first iteration.

Bluenix2 (commented):
Any tips on how I can test this locally?

bot/exts/events/advent_of_code/_helpers.py
D0rs4n: Absolutely not, the logic was moved around a bit and I did not test daily stats just the `per_day_and_star` stats, so I did not notice. Thank you! 👍

D0rs4n:
> Any tips on how I can test this locally?
I'm not sure. SInce the AoC of 2021 hasn't started just yet, I was provided access to last year's data.

janine9vn:
So! I haven't done my formal review yet but here's my feedback so far. Overall, thank you for doing work on this command! My comments will be mostly about how the output looks and to have it be consistent with existing styling.

Currently, the output looks like:
image

I'd prefer it if it matched the overall leaderboard look:
image

I think it would be neat if it displays it like this:
```
     | Name                      | ⭐                      | ⭐⭐
-----------------------------------------------------------------------------------
   1 | Anonymous #500370         |  <points> - <timestamp> | <points> - <timestamp>
   2 | Mushinako                 |  <points> - <timestamp> | <points> - <timestamp>
   3 | godlygeek                 |  <points> - <timestamp> | <points> - <timestamp>
   4 | SavagePastaMan            |  <points> - <timestamp> | <points> - <timestamp>
   5 | ks129                     |  <points> - <timestamp> | <points> - <timestamp>
   6 | salt-die                  |  <points> - <timestamp> | <points> - <timestamp>
   7 | Dennis Pham               |  <points> - <timestamp> | <points> - <timestamp>
   8 | Square789                 |  <points> - <timestamp> | <points> - <timestamp>
   9 | Bast                      |  <points> - <timestamp> | <points> - <timestamp>
  10 | Alex                      |  <points> - <timestamp> | <points> - <timestamp>
```

Then, whether you specify the `-1` or the `-2` would change which day it's sorting on, or just which day to display. Additionally, I think it would be neat to have just a `.aoc lb 3` show both days, with it sorting on the total points earned for that day.

These are just ideas, so I'm fine if you choose not to implement it. But I would like the styling to be a bit more consistent with the existing leaderboard. It may make sense to omit score if the interest is just showing submittal time.

### Other Ideas
So since the timestamp can be rather large with the full date, maybe it would be neat to show `minutes:seconds` since midnight for the leaderboard instead of the full datetime.

D0rs4n:
> So! I haven't done my formal review yet but here's my feedback so far. Overall, thank you for doing work on this command! My comments will be mostly about how the output looks and to have it be consistent with existing styling.
> 
> Currently, the output looks like:
> image
> 
> I'd prefer it if it matched the overall leaderboard look:
> image
> 
> I think it would be neat if it displays it like this:
> 
> ```
>      | Name                      | ⭐                      | ⭐⭐
> -----------------------------------------------------------------------------------
>    1 | Anonymous #500370         |  <points> - <timestamp> | <points> - <timestamp>
>    2 | Mushinako                 |  <points> - <timestamp> | <points> - <timestamp>
>    3 | godlygeek                 |  <points> - <timestamp> | <points> - <timestamp>
>    4 | SavagePastaMan            |  <points> - <timestamp> | <points> - <timestamp>
>    5 | ks129                     |  <points> - <timestamp> | <points> - <timestamp>
>    6 | salt-die                  |  <points> - <timestamp> | <points> - <timestamp>
>    7 | Dennis Pham               |  <points> - <timestamp> | <points> - <timestamp>
>    8 | Square789                 |  <points> - <timestamp> | <points> - <timestamp>
>    9 | Bast                      |  <points> - <timestamp> | <points> - <timestamp>
>   10 | Alex                      |  <points> - <timestamp> | <points> - <timestamp>
> ```
> 
> Then, whether you specify the `-1` or the `-2` would change which day it's sorting on, or just which day to display. Additionally, I think it would be neat to have just a `.aoc lb 3` show both days, with it sorting on the total points earned for that day.
> 
> These are just ideas, so I'm fine if you choose not to implement it. But I would like the styling to be a bit more consistent with the existing leaderboard. It may make sense to omit score if the interest is just showing submittal time.
> ### Other Ideas
> 
> So since the timestamp can be rather large with the full date, maybe it would be neat to show `minutes:seconds` since midnight for the leaderboard instead of the full datetime.

Okay, I see. I think I can make it more consistent. However, I'm not sure if we should parameterize this command too much.
As a golden mean, I believe I could implement this whole thing using buttons and pagination. 
It would look like this(except I would modify the codeblock to be consitent with the other leaderboards) : hellodevlogs
I know, it would lose the sorting part, but it would however provide a fast and clean interface to switch between stars and days.
(that's purely a suggestion and depends on you, I can implement this sorting as well, I just find this a bit more convenient for users)

janine9vn:
@D0rs4n So my concern there is that if you want to view drastically different days (like day 2 and day 23) that's going to be a lot of clicking. I also wonder if a dropdown to select from the current available days wouldn't be better than clicking continually.

I'm also curious if this would only react to the person who invoked it or multiple people, but that's a minor implementation detail.

I think people would also like to view both days side by side, as it would be neat to see the different in submission times for part 1 and part 2.

D0rs4n:
> @D0rs4n So my concern there is that if you want to view drastically different days (like day 2 and day 23) that's going to be a lot of clicking. I also wonder if a dropdown to select from the current available days wouldn't be better than clicking continually.
> 
> I'm also curious if this would only react to the person who invoked it or multiple people, but that's a minor implementation detail.
> 
> I think people would also like to view both days side by side, as it would be neat to see the different in submission times for part 1 and part 2.

As for the dropdown I see how that could work. (Also, for the views there's a global check that gets run on every interaction so it's a minor thing to implement, I personally prefer to add a check so that only the one who invoked command can perform further interactions with the view.) I'll add this with the next change. (After the finalizing this discussion of course.)

As for showing days side by side.. I'm not sure, I could imagine it as a distinct command. Like.: `.aoc days 1 3` and that would show the stats for day 1, and 3, and both of the stars for each of those days, with a dropdown similar to the one mentioned could send codeblocks separated from each other (for example in this case the first codeblock would show stats for day 1, and the second one would show the stats for day 3) for the sake of consistency and edit them as the user chooses a different option in the dropdown.

janine9vn:
> As for showing days side by side.. I'm not sure, I could imagine it as a distinct command. Like.: `.aoc days 1 3` and that would show the stats for day 1, and 3, and both of the stars for each of those days, with a dropdown similar to the one mentioned. It could send codeblocks separated from each other (for example in this case the first codeblock would show stats for day 1, and the second one would show the stats for day 3) for the sake of consistency and edit them as the user chooses a different option in the dropdown.

So it's not days side-by-side, but specifically the part 1 and part 2 of a single day. So I'd prefer it to be kept in one embed.
For instance, it might take 10 minutes to complete the first part, but then 30 seconds to complete the second. Having that kind of comparison available easily would be fun for the event.

D0rs4n:
> 
> 
> > As for showing days side by side.. I'm not sure, I could imagine it as a distinct command. Like.: `.aoc days 1 3` and that would show the stats for day 1, and 3, and both of the stars for each of those days, with a dropdown similar to the one mentioned. It could send codeblocks separated from each other (for example in this case the first codeblock would show stats for day 1, and the second one would show the stats for day 3) for the sake of consistency and edit them as the user chooses a different option in the dropdown.
> 
> 
> 
> So it's not days side-by-side, but specifically the part 1 and part 2 of a single day. So I'd prefer it to be kept in one embed.
> 
> For instance, it might take 10 minutes to complete the first part, but then 30 seconds to complete the second. Having that kind of comparison available easily would be fun for the event.

Okay, I see now. However, I still believe that we should implement a distinct command like "aoc day".
(I can implement this command in that PR as well, if you agree with this idea)

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,68 @@
...
+        self.element_map = {element.custom_id: element for element in self.children}
+        self.maximum_scorers = maximum_scorers
+        self.original_author = original_author
+
+    def generate_output(self) -> str:
+        """Generates a formatted codeblock with AoC statistics based on the currently selected day and star."""
+        header = AOC_DAY_ANDSTAR_TEMPLATE.format(rank="Rank", name="Name",
+                                                 completion_time="Completion time (UTC)")
+        header = f"{header}\n{'-' * (len(header) + 2)}"
+        lines = [header]
```
Bluenix2, line 25:
```suggestion
        lines = [f"{header}\n{'-' * (len(header) + 2)}"]
```

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,68 @@
...
+
+        for rank, scorer in enumerate(self.data[f"{self.day}-{self.star}"][:self.maximum_scorers]):
+            time_data = datetime.fromtimestamp(scorer['completion_time']).strftime("%I:%M:%S %p")
+            lines.append(AOC_DAY_ANDSTAR_TEMPLATE.format(datastamp="", rank=rank + 1, name=scorer['member_name'],
+                                                         completion_time=time_data))
+        joined_lines = "\n".join(lines)
+        return f"Statistics for Day: {self.day}, Star: {self.star}.\n ```\n{joined_lines}\n```"
+
+    async def interaction_check(self, interaction: discord.Interaction) -> bool:
+        """A global check to ensure that the interacting user is the user who invoked the command originally."""
```
Bluenix2, line 35:
```suggestion
        """Global check to ensure that the interacting user is the user who invoked the command originally."""
```

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,68 @@
...
+        return interaction.user == self.original_author
+
+    @discord.ui.select(
+        placeholder="Day",
+        options=[discord.SelectOption(label=str(i)) for i in range(1, 26)],
+        custom_id="day_select"
+    )
+    async def day_select(self, select: discord.ui.Select, interaction: discord.Interaction) -> None:
+        """Represents a Dropdown to choose a Day of the AoC."""
```
Bluenix2, line 44:
```suggestion
        """Dropdown to choose a Day of the AoC."""
```

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,68 @@
...
+        self.day = select.values[0]
+        return
+
+    @discord.ui.select(
+        placeholder="Star",
+        options=[discord.SelectOption(label=str(i)) for i in range(1, 3)],
+        custom_id="star_select"
+    )
+    async def star_select(self, select: discord.ui.Select, interaction: discord.Interaction) -> None:
+        """Represents a Dropdown to choose either the first or the second star."""
```
Bluenix2, line 54:
```suggestion
        """Dropdown to choose either the first or the second star."""
```

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,68 @@
...
+        self.star = select.values[0]
+        return
+
+    @discord.ui.button(label="Fetch", style=discord.ButtonStyle.blurple)
+    async def fetch(self, button: discord.ui.Button, interaction: discord.Interaction) -> None:
+        """Represents a button that fetches the statistics based on the dropdown values."""
```
Bluenix2, line 60:
```suggestion
        """Button that fetches the statistics based on the dropdown values."""
```

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,68 @@
...
+
+class AoCDropdownView(discord.ui.View):
+    """Represents a Dropdown to filter AoC stats by Day and Star."""
+
+    def __init__(self, original_author: discord.Member, day_and_star_data: dict[str: dict], maximum_scorers: int):
+        super().__init__()
+        self.day = 1
+        self.star = 1
+        self.data = day_and_star_data
+        self.element_map = {element.custom_id: element for element in self.children}
```
Bluenix2, line 16: Why is this used now?

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    async def star_select(self, select: discord.ui.Select, interaction: discord.Interaction) -> None:
+        """Represents a Dropdown to choose either the first or the second star."""
+        self.star = select.values[0]
+        return
+
+    @discord.ui.button(label="Fetch", style=discord.ButtonStyle.blurple)
+    async def fetch(self, button: discord.ui.Button, interaction: discord.Interaction) -> None:
+        """Represents a button that fetches the statistics based on the dropdown values."""
+        if not self.element_map['star_select'].values or not self.element_map['day_select'].values:
```
Bluenix2, line 61: Can't we just use `self.star` and `self.day`?

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,68 @@
+from datetime import datetime
+
+import discord
+
+AOC_DAY_ANDSTAR_TEMPLATE = "{rank: >4} | {name:25.25} | {completion_time: >10}"
+
+
+class AoCDropdownView(discord.ui.View):
+    """Represents a Dropdown to filter AoC stats by Day and Star."""
```
Bluenix2, line 9:
```suggestion
    """Interactive view to filter AoC stats by Day and Star."""
```

bot/exts/events/advent_of_code/_helpers.py
```diff
@@ -122,6 +123,12 @@ def _parse_raw_leaderboard_data(raw_leaderboard_data: dict) -> dict:
                 star_results[(day, star)].append(
                     StarResult(member_id=member_id, completion_time=completion_time)
                 )
+                completion_time_raw = int(data["get_star_ts"])
+                per_day_star_stats[f"{day}-{star}"].append(
+                    {'completion_time': completion_time_raw, 'member_name': name}
+                )
```
Bluenix2, line 129:
```suggestion
                per_day_star_stats[f"{day}-{star}"].append(
                    {'completion_time': int(data["get_star_ts"]), 'member_name': name}
                )
```

bot/exts/events/advent_of_code/views/dayandstarview.py
D0rs4n: I don't think so, no. This check is meant to check whether in the current interaction in the very moment when this button is pushed there is one value chosen from both of the dropdowns, since when the user pushes the fetch button, the interaction will revert itself to its original form. (This way we ensure that the placeholder will be shown every time they want to interact with the view). If I were to use those variables in this check,  the view would just fetch the same view that was fetched previously without any warning.

bot/exts/events/advent_of_code/views/dayandstarview.py
D0rs4n: You mean the `element_map`? If so, I came up with this, because I couldn't find an alternative to the built in `children` attribute, I mean that as in using them with indexes. If there's a better alternative I would be more than interested to implement it.

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -185,24 +187,52 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+    async def aoc_leaderboard(
+            self,
+            ctx: commands.Context,
+            day_and_star: Optional[bool] = False,
+            maximum_scorers: Optional[int] = 10
+    ) -> None:
+        """
+        Get the current top scorers of the Python Discord Leaderboard.
+
+        Additionally, you can provide a day-star(e.g. 2-1, as second day first star) to filter your query more.
```
Bluenix2, line 199: This docstring needs updating, `day_and_star` is now a boolean.

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -0,0 +1,67 @@
...
+        lines = [f"{header}\n{'-' * (len(header) + 2)}"]
+
+        for rank, scorer in enumerate(self.data[f"{self.day}-{self.star}"][:self.maximum_scorers]):
+            time_data = datetime.fromtimestamp(scorer['completion_time']).strftime("%I:%M:%S %p")
+            lines.append(AOC_DAY_ANDSTAR_TEMPLATE.format(datastamp="", rank=rank + 1, name=scorer['member_name'],
+                                                         completion_time=time_data))
+        joined_lines = "\n".join(lines)
+        return f"Statistics for Day: {self.day}, Star: {self.star}.\n ```\n{joined_lines}\n```"
+
+    async def interaction_check(self, interaction: discord.Interaction) -> bool:
```
Bluenix2, line 33: Please add an [`on_timeout`](https://discordpy.readthedocs.io/en/master/api.html#discord.ui.View.on_timeout) that will edit the message to have disabled components.

bot/exts/events/advent_of_code/views/dayandstarview.py
Bluenix2: This doesn't even work, I can select one of the dropdowns and fetch. I think it is unneccessary.

bot/exts/events/advent_of_code/views/dayandstarview.py
Bluenix2: But you have `self.star_select` no?

D0rs4n:
> But you have `self.star_select` no?

I do, I did not realize the obvious.. thank you for pointing out. I'll patch my code.

> This doesn't even work, I can select one of the dropdowns and fetch. I think it is unneccessary.

Actually, it does on the first interaction, however I did not take it into account, that the dropdown values will never be reverted to zero. I've found a workaround with the `day` and `star` variables.

bot/exts/events/advent_of_code/views/dayandstarview.py
D0rs4n: I'm not sure how that could work. As far as I'm concerned the on_timeout callback won't receive any kind of interaction, hence even if I disable the elements, or just clear the View, it won't take an effect.
However, after the timeout the View cannot be used.

Bluenix2: Yeah that's exactly what I would've hoped to not get.

I understand why `on_timeout()` doesn't give us an interaction but it does complicate things. There's ways to work around it but I don't think it is worth the bother then.

Bluenix2: Reviving this conversation, see https://github.com/python-discord/sir-lancebot/pull/860#discussion_r716115195.

D0rs4n: Oh, right. It makes sense, I was _still thinking inside the box (view)_  😄 
Will push a patch, later today.

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -178,21 +180,35 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
         else:
             await ctx.message.add_reaction(Emojis.envelope)
 
-    @in_month(Month.DECEMBER)
```
Bluenix2, line 181: Why was this removed?

Bluenix2 (changes requested):

bot/exts/events/advent_of_code/_cog.py
D0rs4n: Oops, yes. I wanted to test it out quickly before pushing a small change previously and I forgot to put it back.

Bluenix2 (approved):
Looks good now, thanks D0rs4n.

PR #858, opened 2021-09-06
Add coloredlogs and selective trace loggers

### LINKED ISSUE #706: Adapt selective trace logging from bot
## Description
Implement selective trace logging from the python-discord/bot#1529 PR, enabling whitelisting or blacklisting certain loggers to emit trace logs instead of always emitting all trace logs with the debug mode.

## Reasoning
Discussed in python-discord/bot#1502

## Proposed Implementation

A new entry for the trace loggers will need to be added to the `Client` config, grabbing its value from the `BOT_TRACE_LOGGERS` 
env var.
The `_set_trace_loggers` function can be copied over from the bot code and called in lancebot's init (possibly with some considerations to differences between the logging setup of the two bots). The default debug levels will have to be changed from trace to debug.

lancebot also sets some levels on the individual handlers, I don't think the difference between them and the root logger is useful as it only affects non debug environments and can be removed, but this may need to be discussed

### LINKED ISSUE #854: Add coloredlogs
We have been using coloredlogs on bot for a while now which makes the logs more readable. There initially were some issues on windows but those seem to have been taken care of now.
The same configuration as the bot uses can be used for consistency.

I can take care of this along with #706 if @ToxicKidz is not working on that issue

Numerlor:
## Relevant Issues
closes #854, closes #706

## Description
Adds coloredlogs to the bot and standardizes the logging format with bot. I've also moved the log files to a logs directory in the project root and renamed the file name from `hackbot.log` to `sir-lancebot.log`. 
Also adds selective trace loggers from python-discord/bot#1529

The trace loggers env var should be documented in the contributing guidelines to be reachable from lancebot's guide


bot/log.py
```diff
@@ -0,0 +1,96 @@
...
+
+    root_logger = logging.getLogger()
+    root_logger.addHandler(file_handler)
+
+    if "COLOREDLOGS_LEVEL_STYLES" not in os.environ:
+        coloredlogs.DEFAULT_LEVEL_STYLES = {
+            **coloredlogs.DEFAULT_LEVEL_STYLES,
+            "trace": {"color": 246},
+            "critical": {"background": "red"},
+            "debug": coloredlogs.DEFAULT_LEVEL_STYLES["info"]
```
brad90four, line 40: ``

brad90four (approved):
Most of this looks like black magic to me, but I don't see anything obviously bad. The one thing I did see is "missing" a trailing comma.

bot/log.py
Numerlor: resolved through a fixup on 6d7f46fabe530e4cafdffa25d7fced730965e481

bot/log.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    level_filter = Client.trace_loggers
+    if level_filter:
+        if level_filter.startswith("*"):
+            logging.getLogger().setLevel(logging.TRACE)
+
+        elif level_filter.startswith("!"):
+            logging.getLogger().setLevel(logging.TRACE)
+            for logger_name in level_filter.strip("!,").split(","):
+                logging.getLogger(logger_name).setLevel(logging.DEBUG)
```
Bluenix2, line 92: Shouldn't there be another loop that sets all of them to TRACE?

Numerlor: The root logger is set to TRACE so loggers we don't set to DEBUG explicitly inherit that

Bluenix2:
Is there anything I need to do to test this? I have both rebuilt the container with `docker-compose build` and removed it using `docker-compose rm`.

I added this logging to make sure changes were applied:
```python
def setup() -> None:
    ...  # Setup of all the loggers

    coloredlogs.install(level=logging.TRACE, logger=root_logger, stream=sys.stdout)
    root_logger.info('Installed coloredlogs')
```

Yet still there's no colours like the other bot has, you can see that Docker adds colors so they should display in my terminal no?

image

I am on Windows 10, using what I assume is Powershell through Visual Studio Code.

Numerlor:
> Is there anything I need to do to test this? I have both rebuilt the container with `docker-compose build` and removed it using `docker-compose rm`.
> 
> I added this logging to make sure changes were applied:
> 
> ```python
> def setup() -> None:
>     ...  # Setup of all the loggers
> 
>     coloredlogs.install(level=logging.TRACE, logger=root_logger, stream=sys.stdout)
>     root_logger.info('Installed coloredlogs')
> ```
> 
> Yet still there's no colours like the other bot has, you can see that Docker adds colors so they should display in my terminal no?

Ah, I forgot to test Docker; it looks like the colours aren't applied because it's not tty when running through Docker.
Do we want to do enable it through `tty: true` in docker-compose like bot does or by simply forcing coloredlogs to colour output with `isattay=True` in the install call?

Bluenix2:
> > Is there anything I need to do to test this? I have both rebuilt the container with `docker-compose build` and removed it using `docker-compose rm`.
> > I added this logging to make sure changes were applied:
> > ```python
> > def setup() -> None:
> >     ...  # Setup of all the loggers
> > 
> >     coloredlogs.install(level=logging.TRACE, logger=root_logger, stream=sys.stdout)
> >     root_logger.info('Installed coloredlogs')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Yet still there's no colours like the other bot has, you can see that Docker adds colors so they should display in my terminal no?
> 
> Ah, I forgot to test Docker; it looks like the colours aren't applied because it's not tty when running through Docker. Do we want to do enable it through `tty: true` in docker-compose like bot does or by simply forcing coloredlogs to colour output with `isattay=True` in the install call?

I think we should just follow what bot does

Numerlor:
> I think we should just follow what bot does

Should work now

jchristgit (approved):
Looks good to me. From my understanding, this is only enabled if a tty is detected, so no weird formatting errors should crop up.

Xithrius (approved):
I see good stuff

PR #802, opened 2021-08-07
Fix type annotations

decorator-factory:
This PR aims to improve type annotations (type hints) in the project:
- change old-style annotations (like `typing.List[int]`) to new-style annotations (like `list[int]`)
- fix annotations that are just wrong

PR #859, opened 2021-09-07
Add WTF Python Command - Continuation of #605

brad90four:
## Relevant Issues

Closes PR #605 
Closes Issue #602 

## Description

Gets the fuzzy match of a search query, returns None if the ratio is below 90 else returns the fuzzy match. If the return is None, then we send our custom Error Message else it sends the link to the subsection of the WTF Python README.md.

## Reasoning
Approved by Staff

## Screenshots from original PR
image

## Did you:

### DIFF
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -60,8 +60,8 @@ class WTFPython(commands.Cog):
                 log.trace("Fetching the latest WTF Python README.md")
 
                 if resp.status == 200:
-                    self.raw = await resp.text()
-                    await self.parse_readme(self.raw)
+                    raw = await resp.text()
+                    self.parse_readme(raw)
                     log.debug("Successfully fetched the latest WTF Python README.md, breaking out of retry loop")
                     break
 
@@ -86,7 +86,7 @@ class WTFPython(commands.Cog):
         else:
             log.debug(f"Extension {verb}ed: `{ext}`.")
 
-    async def parse_readme(self, data: str) -> None:
+    def parse_readme(self, data: str) -> None:
         """
         Parses the README.md into a dict.
 
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -14,7 +14,7 @@ from bot.bot import Bot
 
 log = logging.getLogger(__name__)
 
-WTF_PYTHON_RAW_URL = "http://raw.githubusercontent.com/satwikkansal/wtfpython/master/"
+WTF_PYTHON_RAW_URL = "http://raw.githubusercontent.com/satwikkansal/wtfpython/mster/"
 BASE_URL = "https://github.com/satwikkansal/wtfpython"
 FETCH_TRIES = 3
 
@@ -72,12 +72,13 @@ class WTFPython(commands.Cog):
                     )
 
         if failed_tries == 3:
+            log.error("Couldn't fetch WTF Python README.md after 3 tries, unloading extension.")
             action = Action.UNLOAD
         else:
             action = Action.LOAD
 
         verb = action.name.lower()
-        ext = self.__class__.__name__
+        ext = "bot.exts.evergreen.wtf_python"
 
         try:
             action.value(self.bot, ext)
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -53,7 +53,9 @@ class WTFPython(commands.Cog):
         """
         Parses the README.md into a dict.
 
-        Key is the name of the section and link to it.
+        It parses the readme into the `self.headers` dict,
+        where the key is the heading and the value is the
+        link to the heading.
         """
         table_of_contents = re.findall(
             r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
@@ -71,10 +73,7 @@ class WTFPython(commands.Cog):
 
     def fuzzy_match_header(self, query: str) -> Optional[str]:
         """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
-        match, certainty = process.extractOne(
-            query,
-            list(self.headers.keys())
-        )
+        match, certainty = process.extractOne(query, self.headers.keys())
         return match if certainty > 90 else None
 
     @staticmethod
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -37,12 +37,11 @@ class WTFPython(commands.Cog):
         self.bot = bot
         self.get_wtf_python_readme.start()
 
-        self.data: dict = {}
         self.headers: Dict[str] = dict()
 
     @tasks.loop(hours=1)
     async def get_wtf_python_readme(self) -> None:
-        """Gets content of README.md of WTF Python Repository."""
+        """"Gets the content of README.md from the WTF Python Repository."""
         async with self.bot.http_session.get(WTF_PYTHON_RAW_URL) as resp:
             if resp.status == 200:
                 self.raw = await resp.text()
@@ -71,13 +70,9 @@ class WTFPython(commands.Cog):
                 )
 
     def fuzzy_match_header(self, query: str) -> Optional[str]:
-        """Returns the fizzy match of a query if its ratio is above 90 else returns None."""
-        keys = list(self.headers.keys())
-        match = process.extractOne(query, keys)
-        print(match)
-        if match[1] < 90:
-            return
-        return match[0]
+         """Returns the fuzzy match of a query if its ratio is above 90 else returns None."""
+        match, certainty = process.extractOne(query, list(self.headers.keys()))
+        return match if certainty > 90 else None
 
     @staticmethod
     async def make_embed(query: str, link: str) -> Embed:
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -1,0 +1,117 @@
+import logging
+import random
+import re
+from typing import Dict, Optional
+
+from discord import Embed
+from discord.ext import commands, tasks
+from fuzzywuzzy import process
+
+from bot import constants
+from bot.bot import Bot
+
+log = logging.getLogger(__name__)
+
+WTF_PYTHON_RAW_URL = ("http://raw.githubusercontent.com/satwikkansal/"
+                      "wtfpython/master/README.md")
+BASE_URL = "https://github.com/satwikkansal/wtfpython"
+THUMBNAIL = "https://raw.githubusercontent.com/satwikkansal/wtfpython/master/images/logo.png"
+
+ERROR_MESSAGE = f"""
+Unknown WTF Python Query. Please try to reformulate your query.
+
+**Examples**:
+```md
+{constants.Client.prefix}wtf wild imports
+{constants.Client.prefix}wtf subclass
+{constants.Client.prefix}wtf del
+```
+If the problem persists send a message in <#{constants.Channels.dev_contrib}>
+"""
+
+
+class WTFPython(commands.Cog):
+    """Cog that allows users to retrieve issues from GitHub."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.get_wtf_python_readme.start()
+
+        self.data: dict = {}
+        self.headers: Dict[str] = dict()
+
+    @tasks.loop(hours=1)
+    async def get_wtf_python_readme(self) -> None:
+        """Gets content of README.md of WTF Python Repository."""
+        async with self.bot.http_session.get(WTF_PYTHON_RAW_URL) as resp:
+            if resp.status == 200:
+                self.raw = await resp.text()
+                await self.parse_readme(self.raw)
+            else:
+                log.debug(f"Failed to get latest WTF Python README.md. Status code {resp.status}")
+
+    async def parse_readme(self, data: str) -> None:
+        """
+        Parses the README.md into a dict.
+
+        Key is the name of the section and link to it.
+        """
+        table_of_contents = re.findall(
+            r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
+        )[0].split("\n")
+        table_of_contents = list(map(str.strip, table_of_contents))
+
+        for header in table_of_contents:
+            match = re.findall(r"\[▶ (.*)\]\((.*)\)", header)
+            if match:
+                self.headers.update(
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fizzy match of a query if its ratio is above 90 else returns None."""
+        keys = list(self.headers.keys())
+        match = process.extractOne(query, keys)
+        print(match)
+        if match[1] < 90:
+            return
+        return match[0]
+
+    @staticmethod
+    async def make_embed(query: str, link: str) -> Embed:
+        """Generates a embed for a search."""
+        embed = Embed(
+            title=f"WTF Python Search Result For {query}",
+            colour=constants.Colours.dark_green,
+            description=f"[Go to Repository Section]({link})"
+        )
+        embed.set_thumbnail(url=THUMBNAIL)
+        return embed
+
+    @commands.command(aliases=("wtf",))
+    async def wtf_python(self, ctx: commands.Context, *query: str) -> None:
+        """
+        Search wtf python.
+
+        Gets the link of the fuzzy matched query from https://github.com/satwikkansal/wtfpython.
+        Usage:
+            --> .wtf wild imports
+        """
+        query = " ".join(query)
+        match = self.fuzzy_match_header(query)
+        if not match:
+            embed = Embed(
+                title=random.choice(constants.ERROR_REPLIES),
+                description=ERROR_MESSAGE,
+                colour=constants.Colours.soft_red
+            )
+        else:
+            embed = await self.make_embed(query, self.headers[match])
+        await ctx.send(embed=embed)
+
+
+def setup(bot: Bot) -> None:
+    """Load WTFPython Cog."""
+    bot.add_cog(WTFPython(bot))
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -7,7 +7,7 @@ from typing import Dict, Optional
 
 from discord import Embed
 from discord.ext import commands, tasks
-from fuzzywuzzy import process
+from rapidfuzz import process
 
 from bot import constants
 from bot.bot import Bot
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -1,6 +1,8 @@
+import functools
 import logging
 import random
 import re
+from enum import Enum
 from typing import Dict, Optional
 
 from discord import Embed
@@ -12,10 +14,9 @@ from bot.bot import Bot
 
 log = logging.getLogger(__name__)
 
-WTF_PYTHON_RAW_URL = ("http://raw.githubusercontent.com/satwikkansal/"
-                      "wtfpython/master/README.md")
+WTF_PYTHON_RAW_URL = "http://raw.githubusercontent.com/satwikkansal/wtfpython/master/"
 BASE_URL = "https://github.com/satwikkansal/wtfpython"
-THUMBNAIL = "https://raw.githubusercontent.com/satwikkansal/wtfpython/master/images/logo.png"
+FETCH_TRIES = 3
 
 ERROR_MESSAGE = f"""
 Unknown WTF Python Query. Please try to reformulate your query.
@@ -32,6 +33,14 @@ If the problem persists send a message in <#{constants.Channels.dev_contrib}>
 MINIMUM_CERTAINTY = 75
 
 
+class Action(Enum):
+    """Represents an action to perform on an extension."""
+
+    # Need to be partial otherwise they are considered to be function definitions.
+    LOAD = functools.partial(Bot.load_extension)
+    UNLOAD = functools.partial(Bot.unload_extension)
+
+
 class WTFPython(commands.Cog):
     """Cog that allows users to retrieve issues from GitHub."""
 
@@ -44,12 +53,42 @@ class WTFPython(commands.Cog):
     @tasks.loop(hours=1)
     async def get_wtf_python_readme(self) -> None:
         """Gets the content of README.md from the WTF Python Repository."""
-        async with self.bot.http_session.get(WTF_PYTHON_RAW_URL) as resp:
-            if resp.status == 200:
-                self.raw = await resp.text()
-                await self.parse_readme(self.raw)
-            else:
-                log.debug(f"Failed to get latest WTF Python README.md. Status code {resp.status}")
+        failed_tries = 0
+
+        for x in range(FETCH_TRIES):
+            async with self.bot.http_session.get(f"{WTF_PYTHON_RAW_URL}README.md") as resp:
+                log.trace(f"Fetching the latest WTF Python README.md")
+
+                if resp.status == 200:
+                    self.raw = await resp.text()
+                    await self.parse_readme(self.raw)
+                    log.debug(f"Successfully fetched the latest WTF Python README.md, breaking out of retry loop")
+                    break
+
+                else:
+                    failed_tries += 1
+                    log.debug(
+                        f"Failed to get latest WTF Python README.md on try {x}/{FETCH_TRIES}. Status code {resp.status}"
+                    )
+
+        if failed_tries == 3:
+            action = Action.UNLOAD
+        else:
+            action = Action.LOAD
+
+        verb = action.name.lower()
+        ext = self.__class__.__name__
+
+        try:
+            action.value(self.bot, ext)
+        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded):
+            log.debug(
+                f"Reached maximum failed `FETCH_TRIES`.\n\nExtension `{ext}` is already  is already {verb}ed.."
+            )
+        else:
+            log.debug(
+                f"Reached maximum failed `FETCH_TRIES`.\n\nExtension {verb}ed: `{ext}`."
+            )
 
     async def parse_readme(self, data: str) -> None:
         """
@@ -86,7 +125,7 @@ class WTFPython(commands.Cog):
             colour=constants.Colours.dark_green,
             description=f"[Go to Repository Section]({link})"
         )
-        embed.set_thumbnail(url=THUMBNAIL)
+        embed.set_thumbnail(url=f"{WTF_PYTHON_RAW_URL}images/logo.png")
         return embed
 
     @commands.command(aliases=("wtf",))
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -14,7 +14,7 @@ from bot.bot import Bot
 
 log = logging.getLogger(__name__)
 
-WTF_PYTHON_RAW_URL = "http://raw.githubusercontent.com/satwikkansal/wtfpython/mster/"
+WTF_PYTHON_RAW_URL = "http://raw.githubusercontent.com/satwikkansal/wtfpython/master/"
 BASE_URL = "https://github.com/satwikkansal/wtfpython"
 FETCH_TRIES = 3
 
--- a/bot/exts/evergreen/wtf_python.py
+++ b/bot/exts/evergreen/wtf_python.py
@@ -81,14 +81,10 @@ class WTFPython(commands.Cog):
 
         try:
             action.value(self.bot, ext)
-        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded):
-            log.debug(
-                f"Reached maximum failed `FETCH_TRIES`.\n\nExtension `{ext}` is already  is already {verb}ed.."
-            )
+        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded) as e:
+            log.debug(f"Extension `{ext}` is already  is already {verb}ed.")
         else:
-            log.debug(
-                f"Reached maximum failed `FETCH_TRIES`.\n\nExtension {verb}ed: `{ext}`."
-            )
+            log.debug(f"Extension {verb}ed: `{ext}`.")
 
     async def parse_readme(self, data: str) -> None:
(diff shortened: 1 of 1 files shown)


bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+        Gets the link of the fuzzy matched query from https://github.com/satwikkansal/wtfpython.
+        Usage:
+            --> .wtf wild imports
+        """
+        match = self.fuzzy_match_header(query)
+        if not match:
+            embed = Embed(
+                title=random.choice(constants.ERROR_REPLIES),
+                description=ERROR_MESSAGE,
+                colour=constants.Colours.soft_red
```
Shom770, line 131: Trailing comma,
whenever there are parentheses broken down, at the end there should always be a trailing comma as it makes the git blame more accurate, and adheres to PEP8
Like for this line, it would be:
```py
embed = Embed(
           title=random.choice(constants.ERROR_REPLIES),
           description=ERROR_MESSAGE,
           colour=constants.Colours.soft_red,
            )
```

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+        else:
+            embed = Embed(
+                title=f"WTF Python Search Result For {query}",
+                colour=constants.Colours.dark_green,
+                description=f"[Go to Repository Section]({self.headers[match]})"
```
Shom770, line 137: See comment above regarding trailing comma

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+            embed.set_thumbnail(url=f"{WTF_PYTHON_RAW_URL}images/logo.png")
+        await ctx.send(embed=embed)
+
+
+def setup(bot: Bot) -> None:
+    """Load WTFPython Cog."""
```
Shom770, line 144: Grammatical nitpick - ``"""Load the WTFPython Cog"""``

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+            match = re.findall(r"\[▶ (.*)\]\((.*)\)", header)
+            if match:
+                self.headers.update(
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fuzzy match of a query if its ratio is above "MINIMUM_CERTAINTY" else returns None."""
```
Shom770, line 113: Could you give a brief explanation of what the ratio means in this context?
Possibly adding it to the docstring so if anyone ever needs to edit this in the future they know what it means?

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,145 @@
...
+    LOAD = functools.partial(Bot.load_extension)
+    UNLOAD = functools.partial(Bot.unload_extension)
+
+
+class WTFPython(commands.Cog):
+    """Cog that allows getting WTF Python entries from the WTF Python repository."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.get_WTF_python_readme.start()
```
Shom770, line 49: There is a typo - this should be ``self.get_wtf_python_readme.start()``

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,149 @@
...
+                    )
+
+        if failed_tries == 3:
+            log.error("Couldn't fetch WTF Python README.md after 3 tries, unloading extension.")
+            action = Action.UNLOAD
+        else:
+            action = Action.LOAD
+
+        verb = action.name.lower()
+        ext = "bot.exts.evergreen.wtf_python"
```
Shom770, line 85: This won't work as the filestructure got updated, so this should be
```suggestion
        ext = "bot.exts.utilities.wtf_python"
```

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,149 @@
...
+            if match:
+                self.headers.update(
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
+
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """Returns the fuzzy match of a query if its ratio is above "MINIMUM_CERTAINTY" else returns None."""
+        match, certainty = process.extractOne(query, self.headers.keys())
```
Shom770, line 118: This will return a ValueError, as rapidfuzz.process got updated for extractOne where it returns three values, the match itself, the certainty, and the index the string was found.
It doesn't seem like you are using the index so you could just replace it with a wildcard.
This should be updated to:
```suggestion
        match, certainty, _ = process.extractOne(query, self.headers.keys())
```

Shom770:
image
@brad90four Would you know as to why this is happening?

brad90four:
> 
> 
> image
> @brad90four Would you know as to why this is happening?

Nope, I went ahead and marked as draft if it is not working. I will need to dive deeper into debugging, I am currently away from my testing setup.

Shom770:
> > image
> > @brad90four Would you know as to why this is happening?
> 
> Nope, I went ahead and marked as draft if it is not working. I will need to dive deeper into debugging, I am currently away from my testing setup.

I found the issue -- the minimum certainty is too high.

Right now it's at 75, and when you do ``.wtf del``, it finds the string, but the certainty is only 60%

brad90four:
Hello all,

Just clarifying something for this command and the implementation of it. Is the goal to send all of the text from the Markdown or just a link to that section? 

The way the command is now, it is only the link. In `#dev-contrib` I had a short [conversation](https://discord.com/channels/267624335836053506/635950537262759947/884815062454374501) with Jason Terror that the intention was to parse the `README.md` file and send the text of the section back to the user. 

With the size of the text that could be sent I just wanted to make sure that this is the intention. 

# Current Status:
 
image

# Future Status (mockup):

`.wtf del`
▶ Stubborn del operation

```py
class SomeClass:
    def __del__(self):
        print("Deleted!")
```
Output: 1.
```py
\>>> x = SomeClass()
\>>> y = x
\>>> del x # this should print "Deleted!"
\>>> del y
Deleted!
```
Phew, deleted at last. You might have guessed what saved __del__ from being called in our first attempt to delete x. Let's add more twists to the example.

2.
```py
\>>> x = SomeClass()
\>>> y = x
\>>> del x
\>>> y # check if y exists
<__main__.SomeClass instance at 0x7f98a1a67fc8>
\>>> del y # Like previously, this should print "Deleted!"
\>>> globals() # oh, it didn't. Let's check all our global variables and confirm
Deleted!
{'__builtins__': <module '__builtin__' (built-in)>, 'SomeClass': <class __main__.SomeClass at 0x7f98a1a5f668>, '__package__': None, '__name__': '__main__', '__doc__': None}
```
Okay, now it's deleted 😕
💡 Explanation:

    del x doesn’t directly call x.__del__().
    When del x is encountered, Python deletes the name x from current scope and decrements by 1 the reference count of the object x referenced. __del__() is called only when the object's reference count reaches zero.
    In the second output snippet, __del__() was not called because the previous statement (>>> y) in the interactive interpreter created another reference to the same object (specifically, the _ magic variable which references the result value of the last non None expression on the REPL), thus preventing the reference count from reaching zero when del y was encountered.
    Calling globals (or really, executing anything that will have a non None result) caused _ to reference the new result, dropping the existing reference. Now the reference count reached 0 and we can see "Deleted!" being printed (finally!).

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,159 @@
...
+
+        verb = action.name.lower()
+        ext = "bot.exts.utilities.wtf_python"
+
+        try:
+            action.value(self.bot, ext)
+        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded):
+            log.debug(f"Extension `{ext}` is already {verb}ed.")
+        else:
+            log.debug(f"Extension {verb}ed: `{ext}`.")
```
Bluenix2, line 92: Instead of this, can't we just have the bot save a time of when it was last fetched, then when you use the command and that is older than an hour it will attempt to fetch a new copy?

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,159 @@
...
+        Parses the README.md into a dict.
+
+        It parses the readme into the `self.headers` dict,
+        where the key is the heading and the value is the
+        link to the heading.
+        """
+        table_of_contents = re.findall(
+            r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
+        )[0].split("\n")
+        table_of_contents = list(map(str.strip, table_of_contents))
```
Bluenix2, line 105: In general it's better to do your best to stay away from overriding variables unless that is it's purpose. Even then you only want to do so when they represent the same thing.

In this case, we can just move the variable into the for loop itself. The benefit of that as well is `map()` being an iterator so we skip exhausting the list.

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,159 @@
...
+    def parse_readme(self, data: str) -> None:
+        """
+        Parses the README.md into a dict.
+
+        It parses the readme into the `self.headers` dict,
+        where the key is the heading and the value is the
+        link to the heading.
+        """
+        table_of_contents = re.findall(
+            r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
```
Bluenix2, line 103: It is not awfully apparent to me what these regexes match, are you able to add a comment explaining what they match so that I can see at a glance?

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,159 @@
...
+        table_of_contents = list(map(str.strip, table_of_contents))
+
+        for header in table_of_contents:
+            match = re.findall(r"\[▶ (.*)\]\((.*)\)", header)
+            if match:
+                self.headers.update(
+                    {
+                        match[0][0]: f"{BASE_URL}{match[0][1]}"
+                    }
+                )
```
Bluenix2, line 114:
```suggestion
                self.headers[match[0][0]] = f"{BASE_URL}{match[0][1]}"})
```

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,159 @@
...
+        where the key is the heading and the value is the
+        link to the heading.
+        """
+        table_of_contents = re.findall(
+            r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
+        )[0].split("\n")
+        table_of_contents = list(map(str.strip, table_of_contents))
+
+        for header in table_of_contents:
+            match = re.findall(r"\[▶ (.*)\]\((.*)\)", header)
```
Bluenix2, line 108: Can't you use [re.search](https://docs.python.org/3/library/re.html#re.search) instead? Considering you only use the first match.

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,159 @@
...
+                colour=constants.Colours.soft_red,
+            )
+        else:
+            embed = Embed(
+                title=f"WTF Python Search Result For {query}",
+                colour=constants.Colours.dark_green,
+                description=f"[Go to Repository Section]({self.headers[match]})",
+            )
+            embed.set_thumbnail(url=f"{WTF_PYTHON_RAW_URL}images/logo.png")
+        await ctx.send(embed=embed)
```
Bluenix2, line 154:
```suggestion

        await ctx.send(embed=embed)
```

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class WTFPython(commands.Cog):
+    """Cog that allows getting WTF Python entries from the WTF Python repository."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.get_wtf_python_readme.start()
+
+        self.headers: Dict[str] = dict()
```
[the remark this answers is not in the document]
brad90four: It should be the header of the README.md that matches the user's query, followed by the link to that header. 
This is the README that is being searched against:
https://github.com/satwikkansal/wtfpython/blob/master/README.md

An example header: 
### ▶ Evaluation time discrepancy

And the generated link: 
https://github.com/satwikkansal/wtfpython/blob/master/README.md#-evaluation-time-discrepancy

bot/exts/utilities/wtf_python.py
brad90four: Yeah the README is pretty wild with these in the text: https://github.com/satwikkansal/wtfpython/blob/master/README.md#-examples

bot/exts/utilities/wtf_python.py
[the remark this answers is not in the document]
Bluenix2: So `str`?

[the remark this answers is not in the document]
brad90four: Yessir

bot/exts/utilities/wtf_python.py
Bluenix2: Right so it matches a markdown header and then what's `<!-- tocstop -->`?

brad90four: To be honest, I am not sure. I inherited this PR btw.

But in the raw .md file, the table of contents has that string at the end of the table of contents:

```
# Table of Contents

<!-- Generated using "markdown-toc -i README.md --maxdepth 3"-->

<!-- toc -->

- [Structure of the Examples](#structure-of-the-examples)
    + [▶ Some fancy Title](#-some-fancy-title)
- [Usage](#usage)
- [👀 Examples](#-examples)
  * [Section: Strain your brain!](#section-strain-your-brain)
    + [▶ First things first! *](#-first-things-first-)
    + [▶ Strings can be tricky sometimes](#-strings-can-be-tricky-sometimes)
    + [▶ Be careful with chained operations](#-be-careful-with-chained-operations)
...
...
...
    + [▶ Slowing down `dict` lookups *](#-slowing-down-dict-lookups-)
    + [▶ Bloating instance `dict`s *](#-bloating-instance-dicts-)
    + [▶ Minor Ones *](#-minor-ones-)
- [Contributing](#contributing)
- [Acknowledgements](#acknowledgements)
- [🎓 License](#-license)
  * [Surprise your friends as well!](#surprise-your-friends-as-well)
  * [More content like this?](#more-content-like-this)

<!-- tocstop -->
```

bot/exts/utilities/wtf_python.py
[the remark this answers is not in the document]
Bluenix2: Additionally, after #802 the repository now fully uses `dict` over `typing.Dict`.
```suggestion
        self.headers: dict[str, str] = dict()
```

bot/exts/utilities/wtf_python.py
Bluenix2: Oh I get it now, `<!-- XYZ -->` is of course a comment (HTML-style). Then it stands for "Table Of Contents Stop".

What about a comment like this?
```suggestion
            # Match the start of examples, until the end of the table of contents (toc)
            r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
```

brad90four: Added that above the `re.findall` line. I'll need to test before I change that to `re.search`

bot/exts/utilities/wtf_python.py
brad90four: The way the program is currently written, the `parse_readme` function creates a dictionary of the Table of Content's headers, as well as the hyperlink to them. This gets called wtih `fetch_readme`, which is called in the `__init__`.

Once this dictionary is created, the `fuzzy_match_header` function uses the `rapidfuzz.process.extractOne` method with the user search term and this created dictionary.

If I change that to re.search, then I probably wouldn't be leveraging the fuzzy matching properly with `rapidfuzz`. Is it worth the re-write for that?

bot/exts/utilities/wtf_python.py
brad90four: To check that I'm following you here:
From this:
```py
        table_of_contents = re.findall(
            r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
        )[0].split("\n")
        table_of_contents = list(map(str.strip, table_of_contents))
        log.debug(f"{table_of_contents = }")

        for header in table_of_contents:
```

to this: 
```py
        table_of_contents = re.findall(
            r"\[👀 Examples\]\(#-examples\)\n([\w\W]*)<!-- tocstop -->", data
        )[0].split("\n")
        log.debug(f"{table_of_contents = }")

        for header in list(map(str.strip, table_of_contents)):
```

bot/exts/utilities/wtf_python.py
Bluenix2: I don't think I understand, I simply suggested changing this `re.findall` to `re.search` because you only grab the first result from the "findall".

bot/exts/utilities/wtf_python.py
Bluenix2: Yes

bot/exts/utilities/wtf_python.py
brad90four: Yeah, I tested it out and it still worked. Fixed it!

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,170 @@
...
+        return match if certainty > MINIMUM_CERTAINTY else None
+
+    def readme_file(self, header: str) -> File:
+        """Create a markdown file from the readme using the matched header."""
+        toc_end = self.raw.find("<!-- tocstop -->")
+        header_start = self.raw.find(header, toc_end)
+        header_stop = self.raw.find("\n---\n", header_start)
+        content = self.raw[header_start:header_stop]
+        file_name = "WTF_content.md"
+        with open(file_name, "w") as f:
```
D0rs4n, line 132: I'm not an expert of async programming, but I believe this might be a blocking call. Normally, I would consider using something like aiofiles, however using something like `io.StringIO` might fit your needs better. You can use it with `discord.File` since it's a file-like object.

brad90four: Resolved with 5dcc43f and 917de69

brad90four:
Hey all, 
I think I will just stick with sending the link to the relevant section in the WTF readme file. I don't see a way forward with the file preview of the .md or .txt file.
If that is ok, I will remove the `readme_file` command and the `embed_file` variable.

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,154 @@
...
+        with 100 being a perfect match.
+        """
+        match, certainty, _ = process.extractOne(query, self.headers.keys())
+        log.debug(f"{match = }, {certainty = }")
+        return match if certainty > MINIMUM_CERTAINTY else None
+
+    @commands.command(aliases=("wtf", "WTF"))
+    async def wtf_python(self, ctx: commands.Context, *, query: str) -> None:
+        """
+        Search WTF python.
```
Bluenix2, line 127:
```suggestion
        Search WTF Python repository.
```

bot/exts/utilities/wtf_python.py
brad90four: Whoa I definitely thought that was updated.

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,154 @@
...
+
+        verb = action.name.lower()
+        ext = "bot.exts.utilities.wtf_python"
+
+        try:
+            action.value(self.bot, ext)
+        except (commands.ExtensionAlreadyLoaded, commands.ExtensionNotLoaded):
+            log.debug(f"Extension `{ext}` is already {verb}ed.")
+        else:
+            log.debug(f"Extension {verb}ed: `{ext}`.")
```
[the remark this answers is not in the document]
brad90four: Yeah I'm getting stuck on this one. I haven't used `asyncio` yet so maybe that is what is tripping me up. 
Also, how can I get the "last" time the cog was loaded instead of just calling it immediately?

Here is what I have so far:

```py
class WTFPython(commands.Cog):
    """Cog that allows getting WTF Python entries from the WTF Python repository."""

    def __init__(self, bot: Bot):
        self.bot = bot
        self.headers: dict[str, str] = dict()
        self.last_fetched = datetime.datetime.now()  # this is just the current time, how to access last load?
        log.debug(f"{self.last_fetched = }")

        if self.last_fetched + datetime.timedelta(seconds=10) < datetime.datetime.now():  # this will never be true currently
            log.debug("Last loading time was within x amount of time.")
            asyncio.create_task(self.fetch_readme)  # is this correct?

    # cut out most everything from the function, this is what is left
    async def fetch_readme(self) -> None:
        """Gets the content of README.md from the WTF Python Repository."""
        async with self.bot.http_session.get(f"{WTF_PYTHON_RAW_URL}README.md") as resp:
            log.trace("Fetching the latest WTF Python README.md")
            if resp.status == 200:
                raw = await resp.text()
                self.parse_readme(raw)
```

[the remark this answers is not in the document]
Bluenix2: I am thinking something more like this:
```python

README_REFRESH = 60  # Amount of minutes

class WTFPython(commands.Cog):
    """Cog that allows getting WTF Python entries from the WTF Python repository."""

    def __init__(self, bot: Bot):
        self.bot = bot
        self.headers: dict[str, str] = dict()
        self.last_fetched = datetime.datetime.now()
        log.debug(f"{self.last_fetched = }")

        asyncio.create_task(self.fetch_readme)  # Start a first task

    async def fetch_readme(self) -> None:
        """Gets the content of README.md from the WTF Python Repository."""
        refresh = self.last_fetched + datetime.timedelta(minutes=README_REFRESH)

        if refresh > datetime.datetime.now() and self.headers:  # We also want to fetch if we haven't previously
            return  # Our cache should be up-to-date

        async with self.bot.http_session.get(f"{WTF_PYTHON_RAW_URL}README.md") as resp:
            log.trace("Fetching the latest WTF Python README.md")
            if resp.status == 200:
                raw = await resp.text()
                self.parse_readme(raw)
```
You then call `asyncio.create_task(self.fetch_readme)` every time someone runs the command.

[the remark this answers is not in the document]
brad90four: Whoa that is awesome! Thanks for the super clear example.

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,121 @@
...
+MINIMUM_CERTAINTY = 50
+README_REFRESH = 60  # minutes between fetch calls
+
+
+class WTFPython(commands.Cog):
+    """Cog that allows getting WTF Python entries from the WTF Python repository."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.headers: dict[str, str] = dict()
```
Bluenix2, line 42:
```suggestion
        self.headers: dict[str, str] = {}
```

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,121 @@
...
+        self.last_fetched = datetime.datetime.now()
+        log.debug(f"{self.last_fetched = }")
```
Bluenix2, line 44:
```suggestion
```
Please make sure to remove the debug calls that shouldn't make it into main.

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,121 @@
...
+    def fuzzy_match_header(self, query: str) -> Optional[str]:
+        """
+        Returns the fuzzy match of a query if its ratio is above "MINIMUM_CERTAINTY" else returns None.
+
+        "MINIMUM_CERTAINTY" is the lowest score at which the fuzzy match will return a result.
+        The certainty returned by rapidfuzz.process.extractOne is a score between 0 and 100,
+        with 100 being a perfect match.
+        """
+        match, certainty, _ = rapidfuzz.process.extractOne(query, self.headers.keys())
+        log.debug(f"{match = }, {certainty = }")
```
Bluenix2, line 88:
```suggestion
```

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,122 @@
...
+            --> .wtf wild imports
+        """
+        match = self.fuzzy_match_header(query)
+        if match:
+            embed = Embed(
+                title=f"WTF Python Search Result For {query}",
+                colour=constants.Colours.dark_green,
+                description=f"[Go to Repository Section]({self.headers[match]})",
+            )
+            embed.set_thumbnail(url=f"{WTF_PYTHON_RAW_URL}images/logo.png")
```
Bluenix2, line 104: This looks poor on dark mode:

image

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,122 @@
...
+        Gets the link of the fuzzy matched query from https://github.com/satwikkansal/wtfpython.
+        Usage:
+            --> .wtf wild imports
+        """
+        match = self.fuzzy_match_header(query)
+        if match:
+            embed = Embed(
+                title=f"WTF Python Search Result For {query}",
+                colour=constants.Colours.dark_green,
+                description=f"[Go to Repository Section]({self.headers[match]})",
```
Bluenix2, line 102: The hyperlink doesn't work for me:
image

Additionally, the URL leads to a 404: https://github.com/satwikkansal/wtfpythonWild

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,122 @@
...
+README_REFRESH = 60  # minutes between fetch calls
+
+
+class WTFPython(commands.Cog):
+    """Cog that allows getting WTF Python entries from the WTF Python repository."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.headers: dict[str, str] = {}
+        self.last_fetched = datetime.datetime.now()
```
Bluenix2, line 42: This is no longer used since you decided to use discord.py's tasks.

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,122 @@
...
+        self.fetch_readme.start()
+
+    @tasks.loop(minutes=README_REFRESH)
+    async def fetch_readme(self) -> None:
+        """Gets the content of README.md from the WTF Python Repository."""
+        refresh = self.last_fetched + datetime.timedelta(minutes=README_REFRESH)
+
+        if refresh > datetime.datetime.now() and self.headers:
+            return  # Cache should be up-to-date
```
Bluenix2, line 51: Dead code

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,122 @@
...
+```md
+{constants.Client.prefix}wtf wild imports
+{constants.Client.prefix}wtf subclass
+{constants.Client.prefix}wtf del
+```
+If the problem persists send a message in <#{constants.Channels.dev_contrib}>
+"""
+
+MINIMUM_CERTAINTY = 50
+README_REFRESH = 60  # minutes between fetch calls
```
Bluenix2, line 33: Unneccessary, discord.py now takes care of this

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,122 @@
...
+        Search WTF Python repository.
+
+        Gets the link of the fuzzy matched query from https://github.com/satwikkansal/wtfpython.
+        Usage:
+            --> .wtf wild imports
+        """
+        match = self.fuzzy_match_header(query)
+        if match:
+            embed = Embed(
+                title=f"WTF Python Search Result For {query}",
```
Bluenix2, line 100:
```suggestion
                title=f"WTF Python search result for '{query}'",
```
Additionally, isn't this better added in the description and the title can be something more general?

Bluenix2 (changes requested):

bot/exts/utilities/wtf_python.py
brad90four: Since the image is a .png, how can I change the embed background but only for dark mode?

bot/exts/utilities/wtf_python.py
brad90four: How does this look? 

image

bot/exts/utilities/wtf_python.py
Bluenix2: You'll have to find a picture that works better on both platforms, there's no way to differentiate it like that.

bot/exts/utilities/wtf_python.py
Bluenix2: I like that, but the capitalization is a bit over-the-top (just do `Search results for 'del'`)

bot/exts/utilities/wtf_python.py
brad90four: I could save it as a .jpg and add it to the resources folder?

Bluenix2: Because of the license ([see here](https://github.com/satwikkansal/wtfpython/blob/master/LICENSE)) we actually could do that afaik.

For legal reasons I am not impliying anything and this is not legal advice ;)

brad90four: How about no image at all?

brad90four: Doesn't look to bad to me:

image

bot/exts/utilities/wtf_python.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        match = self.fuzzy_match_header(query)
+        if match:
+            log.debug(f"{self.headers[match]}")
+            embed = Embed(
+                title="WTF Python?!",
+                colour=constants.Colours.dark_green,
+                description=f"""Search result for '{query}'
+                [Go to Repository Section]({self.headers[match]})""",
+            )
```
Bluenix2, line 98: Can this embed clarify that an option was found and what it is?

brad90four: More than the quoted user query? Maybe the `self.headers[match]` to show exactly what the link is taking the user to?

Bluenix2: Yeah something like that. So that it shows that it found for example "wtf imports" or similar.

brad90four: This is what it looks like after 2197fc6:
image

Bluenix2: I like that, thank you Brad.

Bluenix2 (commented):
Can we have the command have some minimum threshold?

I tried it with the start of the Bee Movie script, first few seconds of Nicki Minaj's superbass, first page of the Bible, the first sentence of the FitnessGram PACER Test, a paragraph about the MRSA bacteria, the MIT license and your name....

but it wasn't until I brought out the morse code of "Hello World" and Never Gonna Give You Up music video link that I got a failure.

PR #860, opened 2021-09-07
Challenges

### LINKED ISSUE #848: Adding a challenge command
## Description

The `.challenge` command would pull a random kata (challenge) from codewars.com, and the user can filter with either the query or the kyu (difficulty), 1 being the hardest, 8 being the easiest, along with the programming language the kata is in.

## Reasoning

`codewars.com` is a popular website for coding challenges, which many people use. A lot of people including myself want to be challenged sometimes but aren't sure what to do, and sometimes people also hit 'code blocks'. This command would provide a random challenge from codewars.com, or kata as codewars calls it, so said person can be inspired to try the challenge provided. This command can be considered analogous to the `stackoverflow` and `realpython` commands, though different in the sense that it focuses on giving ideas rather than either sending the user to a helpful question that solves their problem or articles that help increase the understanding of the topic the user wants to learn.

## Proposed Implementation

**.challenge without any parameters, where the language parameter will default to python, and a random difficulty is chosen**
image
**.challenge with a difficulty parameter, looking for a random kata in that difficulty**
image
**.challenge with a query parameter**
image
**.challenge with a query AND difficulty parameter, splitted by the dilimiter ','**
image
**.challenge with an invalid query**
image

Shom770:
## Relevant Issues

Closes #848

## Description

This pull request adds a `.challenge` command, which uses bs4 to filter through the queries given for the challenge command, get the kata ID, then uses the codewars API to get information about the kata.

Here is an example of it running:
Say I used ``.challenge python binary tree``, it essentially will search for the query 'binary tree' for katas within the scope of the python katas. It'll format the link to `https://www.codewars.com/kata/search/python?q=binary%20tree&beta=false`, and pull a random kata from all the katas using bs4 to filter through the website.

Once it has the kata ID, it will use codewars' API with the link `https://codewars.com/api/v1/code-challenges/{kata-id}`, where kata-id is the kata we just pulled.

Lastly, it will format the embed using the JSON provided and display like this:
image
Do note that I don't have the emojis uploaded, though if added to Sir Lancebot, they will display properly.

## Did you:


bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,242 @@
...
+
+logger = logging.getLogger(__name__)
+API_ROOT = "https://www.codewars.com/api/v1/code-challenges/{kata_id}"
+
+
+class Challenges(commands.Cog):
+    """
+    Cog for a challenge command.
+
+    Pulls a random kata from codewars.com, and can be filtered through.
```
Bluenix2, line 19: I feel like there's room for improvement over this.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,242 @@
...
+                original_message=original_message,
+                main_embed=kata_embed,
+                language_embed=language_embed,
+                tags_embed=tags_embed,
+                other_info_embed=miscellaneous_embed,
+            )
+            await original_message.edit(
+                embed=kata_embed,
+                view=KataView(f'https://codewars.com/kata/{first_kata_id}', dropdown)
+            )
```
Bluenix2, line 166: This is a HUUUUGEE command, you have a whole cog to use! Can you split this into multiple methods and keep the command itself as small as possible?

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+# creates the information dropdown
```
Bluenix2, line 169:
```suggestion
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,242 @@
...
+class KataView(ui.View):
+    """The view for the InformationDropdown and the link to the Kata."""
+
+    def __init__(self, link: str, dropdown: InformationDropdown):
+        super().__init__()
+
+        # adding the dropdown
+        self.add_item(dropdown)
+        # manually creating the button as it is a link button and has no decorator
+        self.add_item(ui.Button(label="View the Kata", url=link))
```
Bluenix2, line 237: We don't use the functionality of subclassing particularly much, we can just create an instance directly.
```py
view = ui.View()
view.add_item(dropdown)
view.add_item(ui.Button(label="View the kata", url=link))
```
Just place this in the command body itself

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def setup(bot: Bot) -> None:
+    """Sets up Challenges cog."""
```
Bluenix2, line 241:
```suggestion
    """Load the Challenges cog."""
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,242 @@
...
+        super().__init__(
+            placeholder='See more information regarding this kata',
+            min_values=1,
+            max_values=1,
+            options=options,
+        )
+
+    async def callback(self, interaction: Interaction) -> None:
+        """Callback for when someone clicks on a dropdown."""
+        # option chosen will retrieve the embed for that specific option then the message will be edited
```
Bluenix2, line 222:
```suggestion
        # Edit the message to the embed selected in the option
```

bot/exts/utilities/challenges.py
Shom770: Yep, I'll split them into helper functions, though the functionality should still stay as one command so it makes sense

bot/exts/utilities/challenges.py
Shom770: Sounds good

bot/exts/utilities/challenges.py
Shom770: How does this sound
```
"""
Cog for a challenge command.

The challenge command pulls a random kata from codewars.com. A kata is the name for a challenge, specific to `codewars.com`

The challenge command also has filters to customize the kata that is given. You can specify the language the kata should be from, the difficulty of the kata (1 is the hardest, 8 is the easiest, on a scale from 1-8), and lastly you can customize the topic you want the kata to be about
"""
```

Bluenix2: Sounds good to me.

bot/exts/utilities/challenges.py
Shom770: @Bluenix2 Alright, I made the command much smaller!
Let me know what you think when you're free!

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+            4: (60, 126, 187), 3: (60, 126, 187), 2: (134, 108, 199), 1: (134, 108, 199)
+        }
+
+        kata_description = kata_information["description"]
+
+        # ensuring it isn't over the length 1024
+        if len(kata_description) > 1024:
+            kata_description = "\n".join(kata_description[:1000].split("\n")[:-1])
+            kata_id = kata_information["id"]
+            kata_description += f"\n[Read more]({f'https://codewars.com/kata/{kata_id}'})"
```
Bluenix2, line 148: Is this [f-string inception](https://open.spotify.com/track/5xKVYMxOHB2XRLCUafFrz6?si=fdd0a0924b5c4d7c)?

```suggestion
            kata_url = f"https://codewars.com/kata/{kata_information['id']}"
            kata_description += f"\n[Read more](kata_url)"
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+
+        # creating the main kata embed
```
Bluenix2, line 150:
```suggestion
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+        kata_embed = Embed(
+            title=kata_information['name'],
+            description=kata_description,
+            color=Color.from_rgb(
+                *mapping_of_kyu[int(kata_information['rank']['name'].replace(' kyu', ''))]
+            ),
+        )
```
Bluenix2, line 157: Can we add a URL to this embed so that the title is clickable as well?

Even though you do get the link with a button.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+        kata_embed.add_field(name="Difficulty", value=kata_information['rank']['name'], inline=False)
+        return kata_embed
+
+    @staticmethod
+    def language_embed(kata_information: dict) -> Embed:
+        """
+        Creates the 'language embed' which displays all languages the kata supports.
+
+        Takes in the kata information dictionary as an argument from the codewars.com API result (JSON)
```
Bluenix2, line 166:
```suggestion
        Takes in the kata information dictionary as an argument from the codewars.com API result.
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+            color=Colours.python_blue,
+        )
+        return language_embed
+
+    @staticmethod
+    def tags_embed(kata_information: dict) -> Embed:
+        """
+        Creates the 'tags embed' which displays all the tags of the Kata.
+
+        Tags are what the Kata is about, and is another word for 'categories'
```
Bluenix2, line 181:
```suggestion
        Tags explain what the Kata is about, this is what codewars.com calls categories.
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+
+        Takes in the kata information dictionary as an argument from the codewars.com API result (JSON)
```
Bluenix2, line 183: We understand this from the function head (`def tags_embed(kata_information: dict) -> Embed:`).

Please remove this from all methods.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+    def miscellaneous_embed(kata_information: dict) -> Embed:
+        """
+        Creates the 'other information embed' which displays miscellaneous information about the kata.
+
+        This embed shows statistics like the total number of people who completed the kata,
+        the total number of stars of the Kata, etc.
+
+        Takes in the kata information dictionary as an argument from the codewars.com API result (JSON)
+        """
+        miscellaneous_embed = Embed(
```
Bluenix2, line 203: Could just be called `embed` ¯\\_(ツ)_/¯

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+            name="`Total Attempts`",
+            value=f"**{kata_information['totalAttempts']}**",
+            inline=False,
+        )
+        return miscellaneous_embed
+
+    @staticmethod
+    def create_view(dropdown: InformationDropdown, link: str) -> ui.View:
+        """
+        Creates the 'view', which is essentially a name for the discord UI.
```
Bluenix2, line 232: "Discord UI" is also another made up name and does not make more sense in the grand-scheme of the Discord API.

```suggestion
        Creates the discord.py View for the Discord message components (dropdowns and buttons).
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+
+        The different ways to use this command are:
+        `.challenge <language>` - Pulls a random challenge within that language's scope.
+        `.challenge <language> <difficulty>` - The difficulty can be from 1-8,
+        1 being the hardest, 8 being the easiest. This pulls a random challenge within that difficulty & language.
+        `.challenge <language> <query>` - Pulls a random challenge with the query provided under the language
+        `.challenge <language> <query>, <difficulty>` - Pulls a random challenge with the query provided,
+        under that difficulty within the language's scope.
+        """
+        get_kata_link = f"https://codewars.com/kata/search/{language}"
```
Bluenix2, line 257: This means we are accepting arbitrary user-input. Any better option? Is it documented all the languages Codewars.com has?

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+
+        if language and not query:
+            level = f"-{choice(['1', '2', '3', '4', '5', '6', '7', '8'])}"
+        elif "," in query or ", " in query:
```
Bluenix2, line 261: `1, 2` and `1,2` will both match `',' in query`.
```suggestion
        elif "," in query:
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+            main_embed=kata_embed,
+            language_embed=language_embed,
+            tags_embed=tags_embed,
+            other_info_embed=miscellaneous_embed,
+        )
+        kata_view = self.create_view(dropdown, f'https://codewars.com/kata/{first_kata_id}')
+        await original_message.edit(
+            embed=kata_embed,
+            view=kata_view
+        )
```
Bluenix2, line 304: Any reason we can't just create the message with the view, and assign the attribute later on? API calls are expensive.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+        `.challenge <language> <difficulty>` - The difficulty can be from 1-8,
+        1 being the hardest, 8 being the easiest. This pulls a random challenge within that difficulty & language.
+        `.challenge <language> <query>` - Pulls a random challenge with the query provided under the language
+        `.challenge <language> <query>, <difficulty>` - Pulls a random challenge with the query provided,
+        under that difficulty within the language's scope.
+        """
+        get_kata_link = f"https://codewars.com/kata/search/{language}"
+
+        if language and not query:
+            level = f"-{choice(['1', '2', '3', '4', '5', '6', '7', '8'])}"
```
Bluenix2, line 260: We can save some space by just having these as integers, when put into the f-string they will become formatted into strings.
```suggestion
            level = f"-{choice([1, 2, 3, 4, 5, 6, 7, 8])}"
```
That said, you can take a look at [`random.randint()`](https://docs.python.org/3/library/random.html#random.randint).

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+                )
+
+            query, level = query_splitted
+            level = f"-{level}"
+        elif query.isnumeric():
+            level, query = f"-{query}", None
+        else:
+            level = None
+
+        params = {name: value for name, value in zip(["q", "r[]"], [query, level]) if value}
```
Bluenix2, line 276: You could create the `params` dictionary first, and then instead of setting variables just do `params["q"] = f"-{query}"` in the if-statement.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+
+        kata_information = await self.kata_information(first_kata_id)
+
+        kata_embed = self.main_embed(kata_information)
+
+        language_embed = self.language_embed(kata_information)
+
+        tags_embed = self.tags_embed(kata_information)
+
+        miscellaneous_embed = self.miscellaneous_embed(kata_information)
```
Bluenix2, line 289: I appreciate the social distancing on this one, but we still need to build up a herd-immunity.
```suggestion
        first_kata_id = await self.kata_id(get_kata_link, params)
        kata_information = await self.kata_information(first_kata_id)

        kata_embed = self.main_embed(kata_information)
        language_embed = self.language_embed(kata_information)
        tags_embed = self.tags_embed(kata_information)
        miscellaneous_embed = self.miscellaneous_embed(kata_information)
```

Bluenix2 (changes requested):
PS: Damm you for hooking me into Hans Zimmer film music again! Now I won't be able to listen to anything else but Inception and Interstellar for another 2 weeks

bot/exts/utilities/challenges.py
Shom770: > Any reason we can't just create the message with the view, and assign the attribute later on? API calls are expensive.

I suppose I could assign the attribute later

bot/exts/utilities/challenges.py
Shom770: Haha, yeah I'll be fixing all of these when I get back to my device

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+        This will webscrape the search page with `search_link` and then get the ID of a kata for the
+        codewars.com API to use.
+        """
+        async with self.bot.http_session.get(search_link, params=params) as response:
+            if response.status != 200:
+                logger.error(
+                    f"Unexpected status code {response.status} from codewars.com"
+                )
+            soup = BeautifulSoup(await response.text(), features="lxml")
+            first_kata_div = soup.find_all("div", class_="item-title px-0")
```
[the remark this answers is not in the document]
Shom770: The Python discord bot implements bs4 not in another thread iirc, and it's been working fine, but this is a valid point

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+    You can specify the language the kata should be from, the difficulty of the kata.
+    Lastly, you can customize the topic you want the kata to be about!
+    """
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    async def kata_id(self, search_link: str, params: dict) -> str:
+        """
+        Using bs4, and with a formatted link with the filters provided by the user.
```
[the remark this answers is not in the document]
Shom770: Oops.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,309 @@
...
+    def main_embed(kata_information: dict) -> Embed:
+        """
+        Creates the main embed which displays the description of the kata and the difficulty, along with the name.
+
+        Takes in the kata information dictionary as an argument from the codewars.com API result (JSON)
+        """
+        # maps difficulty to color for embed
+        mapping_of_kyu = {
+            8: (221, 219, 218), 7: (221, 219, 218), 6: (236, 182, 19), 5: (236, 182, 19),
+            4: (60, 126, 187), 3: (60, 126, 187), 2: (134, 108, 199), 1: (134, 108, 199)
```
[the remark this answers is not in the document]
Shom770: > Is there any reason this doesn't direclty use the `discord.Color` class?

Yes, the colors are actually the colors of the kyu levels on codewars itself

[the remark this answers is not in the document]
Shom770: I will note that down!

bot/exts/utilities/challenges.py
Shom770: That's a bit too generic for my liking, haha

bot/exts/utilities/challenges.py
Shom770: Sure!

Shom770:
@Bluenix2 Added most of your comments, besides the ones I wanted elaboration on!

bot/exts/utilities/challenges.py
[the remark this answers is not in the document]
Bluenix2: I'd rather see these as hex values, it's a much more common way of sharing colors and makes it much easier for someone to reuse.

bot/exts/utilities/challenges.py
Bluenix2: But we are inside of the `miscellaneous_embed()` method.

Bluenix2:
> @Bluenix2 Added most of your comments, besides the ones I wanted elaboration on!

I don't understand, I can't any question or similar for me to answer.

Bluenix2 (changes requested):
It appears that the Hangman game from your other pull request has leaked over into this one. Are you not using two different branches?

Shom770:
> It appears that the Hangman game from your other pull request has leaked over into this one. Are you not using two different branches?

I accidentally pulled the hangman branch instead of the challenges branch which is what this is on, and I thought the changes were gone, I'll remove the hangman from it

Shom770:
> > @Bluenix2 Added most of your comments, besides the ones I wanted elaboration on!
> 
> I don't understand, I can't any question or similar for me to answer.

Nevermind, I think I got confused earlier whether you were asking a question or not

Bluenix2:
What happened to `bs4` and `lxml`? It was removed from the lock file.

Shom770:
> What happened to `bs4` and `lxml`? It was removed from the lock file.

I am not thinking today lol, I removed it because I didn't realize that it was part of this command

My bad, I'll add it back

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,323 @@
...
+            min_values=1,
+            max_values=1,
+            options=options,
+        )
+
+    async def callback(self, interaction: Interaction) -> None:
+        """Callback for when someone clicks on a dropdown."""
+        # Edit the message to the embed selected in the option
+        result_embed = self.mapping_of_embeds[self.values[0]]
+        await self.original_message.edit(embed=result_embed)
```
Bluenix2, line 88: Should probably comment that `original_message` will be set after being initialized.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,323 @@
...
+from bot.constants import Colours, Emojis, NEGATIVE_REPLIES
+
+logger = logging.getLogger(__name__)
+API_ROOT = "https://www.codewars.com/api/v1/code-challenges/{kata_id}"
+# Map difficulty for the kata to color we want to display in the embed.
+# These colors are representative of the colors that each 'kyu' level represents on codewars.com
+MAPPING_OF_KYU = {
+    8: (221, 219, 218), 7: (221, 219, 218), 6: (236, 182, 19), 5: (236, 182, 19),
+    4: (60, 126, 187), 3: (60, 126, 187), 2: (134, 108, 199), 1: (134, 108, 199),
+}
```
Bluenix2, line 19: Technically this could just be a list where we do the lookup by index. That said, then it will start at 0 not 1.

Not sure what's best

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,323 @@
...
+    def tags_embed(kata_information: dict) -> Embed:
+        """
+        Creates the 'tags embed' which displays all the tags of the Kata.
+
+        Tags explain what the Kata is about, this is what codewars.com calls categories.
+        """
+        tags = '\n'.join(kata_information['tags'])
+        tags_embed = Embed(
+            title="Tags",
+            description=f"```nim\n{tags}```",
```
Bluenix2, line 198: Does this not need to be followed by a newline as well?

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,323 @@
...
+        elif query.isnumeric():
+            level, query = f"-{query}", None
+        else:
+            level = None
+
+        params = {}
+        if query:
+            params["q"] = query
+        if level:
+            params["r[]"] = level
```
Bluenix2, line 289: Again, you can even just move this into the if-statements themselves.

So instead of doing `level, query = f"-{query}", None` you can just do `params["q"] = f"-{query}`.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,323 @@
...
+
+        params["beta"] = "true" if language in SUPPORTED_LANGUAGES["beta"] else "false"
```
Bluenix2, line 291: I feel like aiohttp or httpx should handle this, did you get issues with doing `params["beta"] = language in SUPPORTED_LANGUAGES["beta"]`?

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,323 @@
...
+
+        first_kata_id = await self.kata_id(get_kata_link, params)
+        if isinstance(first_kata_id, Embed):
+            # ran into an error with retrieving the website link
```
Bluenix2, line 295:
```suggestion
            # We ran into an error when retrieving the website link
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,323 @@
...
+            await ctx.send(embed=first_kata_id)
+            return
+
+        kata_information = await self.kata_information(first_kata_id)
+        if isinstance(kata_information, Embed):
+            # ran into an error with the codewars api, getting the kata information
```
Bluenix2, line 301:
```suggestion
            # Something went wrong when trying to fetch the kata information
```
Please make it a habit to capitalize your comments

bot/exts/utilities/challenges.py
Shom770: Can you elaborate?

Bluenix2: Well, after `{tags}` do we not need a new line so that it becomes (demonstrated with single-quotes)
```
'''nim
{tags}
'''
```
Instead of what this now becomes
```
'''nim
{tags}'''
```
But that may not matter?

bot/exts/utilities/challenges.py
Shom770: Unfortunately, it doesn't handle it.

bot/exts/utilities/challenges.py
Shom770: Hm, I honestly think that the dictionary is more fitting since it is a mapping

bot/exts/utilities/challenges.py
Shom770: Will do

bot/exts/utilities/challenges.py
Kronifer: yes, [here](https://docs.codewars.com/languages/)

Kronifer (approved):
lgtm :shipit:

bot/exts/utilities/challenges.py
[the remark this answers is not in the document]
tooboredtocode: There is the option to run the calculation in a process pool (relevant docs [here](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) and [here](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor)) if this does take up a lot of processing, this would allow the main event loop be freed to do other things.

bot/exts/utilities/challenges.py
Shom770: Hm, true

bot/exts/utilities/challenges.py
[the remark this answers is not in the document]
Shom770: Thank you so much! So essentially, asyncio can run blocking code in a ThreadPoolExecutor (threading) and the main event loop can do other things?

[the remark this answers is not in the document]
tooboredtocode: I'd rather use a the process pool executer (the second here) since this uses the multiprocessing feature allowing us to bypass the GIL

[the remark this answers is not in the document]
Numerlor: What would be the use of bypassing the GIL here? The threads can still switch and I don't think we need to utilize more of the CPU.

From some tests I've done for the docs bs4 took quite some time for longer HTML so if it blocks for more than a ms or so I'd say it's definitely worth it, spawning threads is cheap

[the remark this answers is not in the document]
tooboredtocode: Using threading could still potentially block the event loop, which is why I suggested using Multiprocessing, but yeah it is a bit unnecessary.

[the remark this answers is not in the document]
Bluenix2: No it won't potentially do so. Threading will not block the event loop.

The thing with threading is that it won't go faster, because it is not run in parallel and even when it is ran concurrently Python has the GIL which means that even concurrently only one thread can run at a time.. so half concurrently!

But that doesn't matter and we don't care, this is not a high priority calculation that we need to speed up with parallelism. The objective is to just not make it not block the event loop.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,324 @@
...
+                    description="We ran into an error with getting the kata information, try again later.",
+                    color=Colours.soft_red,
+                )
+                return error_embed
+
+            return await response.json()
+
+    @staticmethod
+    def main_embed(kata_information: dict) -> Embed:
+        """Creates the main embed which displays the description of the kata and the difficulty, along with the name."""
```
TizzySaurus, line 159: "... which displays the name, difficulty and description of the kata." seems to read better.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+        kata_embed.add_field(name="Difficulty", value=kata_information['rank']['name'], inline=False)
+        return kata_embed
+
+    @staticmethod
+    def language_embed(kata_information: dict) -> Embed:
+        """Creates the 'language embed' which displays all languages the kata supports."""
+        languages = '\n'.join(map(str.title, kata_information['languages']))
+        language_embed = Embed(
+            title="Languages Supported",
```
TizzySaurus, line 184: Nitpicky but think this reads better as "Supported Languages" instead of "Languages Supported".

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,324 @@
...
+            color=Colours.python_blue,
+        )
+        return language_embed
+
+    @staticmethod
+    def tags_embed(kata_information: dict) -> Embed:
+        """
+        Creates the 'tags embed' which displays all the tags of the Kata.
+
+        Tags explain what the Kata is about, this is what codewars.com calls categories.
```
TizzySaurus, line 195: Think this reads better as "Tags explain what the kata is about, **and** is what codewars.com calls categories.". There's also the fact that `Kata` here is capitalised whereas in most places it's lowercase (should probably make all occurences upper/lower for consistency -- L211 also has kata capitalised).

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,324 @@
...
+            color=Colours.grass_green,
+        )
+        return tags_embed
+
+    @staticmethod
+    def miscellaneous_embed(kata_information: dict) -> Embed:
+        """
+        Creates the 'other information embed' which displays miscellaneous information about the kata.
+
+        This embed shows statistics like the total number of people who completed the kata,
```
TizzySaurus, line 210: I think "such as" in place of "like" reads better here due to the "etc."

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,324 @@
...
+
+            # there are numerous divs before arriving at the id of the kata, which can be used for the link.
+            first_kata_id = first_kata_div.a["href"].split("/")[-1]
+            return first_kata_id
+
+    async def kata_information(self, kata_id: str) -> Union[dict, Embed]:
+        """
+        Returns the information about the Kata.
+
+        Uses the codewars.com API to get information about the kata using the kata's ID.
```
TizzySaurus, line 144: Perhaps change "using the kata's ID" to "using \`kata_id\`" to indicate use of the parameter.

TizzySaurus (changes requested):
A lot of these are nitpicks and up to you whether you implement, but I do feel the following should be addressed:

- L101-L102
- trailing commas
- the split `__init__` function
- the inconsistency of "kata" vs "Kata"
- the inconsistency of "codewars.com" vs "\`codewards.com\`"

wookie184 (dismissed):
I am currently getting an error when trying to run this. It seems to be here:
```python
 first_kata_div = await to_thread(soup.find_all("div", class_="item-title px-0"))
```
`to_thread` takes a function as the argument for what to run, this is instead running `soup.find_all` and passing the result to `to_thread`, which causes an error. Looking at the docs, https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread,
 it seems something like
 ```python
 first_kata_div = await to_thread(soup.find_all, "div", class_="item-title px-0")
 ```
 would fix that, although I haven't tried it yet myself.

Shom770:
> I am currently getting an error when trying to run this. It seems to be here:
> 
> ```python
>  first_kata_div = await to_thread(soup.find_all("div", class_="item-title px-0"))
> ```
> 
> `to_thread` takes a function as the argument for what to run, this is instead running `soup.find_all` and passing the result to `to_thread`, which causes an error. Looking at the docs, https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread,
> it seems something like
> 
> ```python
> first_kata_div = await to_thread(soup.find_all, "div", class_="item-title px-0")
> ```
> 
> would fix that, although I haven't tried it yet myself.

Fixed! I can’t test it out since I’m on my phone but let me know if any other errors come up.

Bluenix2 (changes requested):
I get this error from running the `.challenges` command.
```python
 File "/usr/local/lib/python3.9/site-packages/discord/ext/commands/core.py", line 167, in wrapped
  File "/bot/bot/exts/utilities/challenges.py", line 298, in challenge
    kata_embed = self.main_embed(kata_information)
  File "/bot/bot/exts/utilities/challenges.py", line 167, in main_embed
    color=int(MAPPING_OF_KYU[int(kata_information['rank']['name'].replace(' kyu', ''))], 16),
TypeError: int() can't convert non-string with explicit base
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,319 @@
...
+
+        # Ensuring it isn't over the length 1024
+        if len(kata_description) > 1024:
+            kata_description = "\n".join(kata_description[:1000].split("\n")[:-1]) + "..."
+            kata_description += f"\n[Read more...]({kata_url})"
+
+        kata_embed = Embed(
+            title=kata_information['name'],
+            description=kata_description,
+            color=int(MAPPING_OF_KYU[int(kata_information['rank']['name'].replace(' kyu', ''))], 16),
```
Bluenix2, line 167: After https://github.com/python-discord/sir-lancebot/pull/860#discussion_r711644770 was commited this line of code wasn't changed.
```suggestion
            color=MAPPING_OF_KYU[int(kata_information['rank']['name'].replace(' kyu', ''))],
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+        kata_embed.add_field(name="Difficulty", value=kata_information['rank']['name'], inline=False)
+        return kata_embed
+
+    @staticmethod
+    def language_embed(kata_information: dict) -> Embed:
+        """Creates the 'language embed' which displays all languages the kata supports."""
+        languages = '\n'.join(map(str.title, kata_information['languages']))
+        language_embed = Embed(
+            title="Supported Languages",
```
Bluenix2, line 178: Can the title for these include the title for the Kata the same way the main embed does?

Maybe in the style of "Main Information | Name on Billboard" (or the other way around) or in some other style if you prefer that. I got a bit lost after selecting other information.

*Applies to the other embeds*

Bluenix2 (commented):
This is like.. *really cool*. Love the way it uses the interactions. Just got a little bit lost when using the menu, see my comment.

bot/exts/utilities/challenges.py
Shom770: How does this look?
image

Bluenix2: I think it is good there in the select menu, I am talking about the embed title.

The issue is that the main information has the title, but for example "Programming Languages" has exactly that as its title (not the name of the Kata included somewhere).

Shom770: Ah, alright

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,319 @@
...
+    @staticmethod
+    def main_embed(kata_information: dict) -> Embed:
+        """Creates the main embed which displays the name, difficulty and description of the kata."""
+        kata_description = kata_information["description"]
+        kata_url = f"https://codewars.com/kata/{kata_information['id']}"
+
+        # Ensuring it isn't over the length 1024
+        if len(kata_description) > 1024:
+            kata_description = "\n".join(kata_description[:1000].split("\n")[:-1]) + "..."
+            kata_description += f"\n[Read more...]({kata_url})"
```
Bluenix2, line 162:
```suggestion
            kata_description += f" [continue reading]({kata_url})"
```
This way it mimics the mailing list's behaviour which I quite like.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,319 @@
...
+                logger.error(f"Unexpected response from codewars.com/api/v1, status code: {response.status}")
+                return error_embed
+
+            return await response.json()
+
+    @staticmethod
+    def main_embed(kata_information: dict) -> Embed:
+        """Creates the main embed which displays the name, difficulty and description of the kata."""
+        kata_description = kata_information["description"]
+        kata_url = f"https://codewars.com/kata/{kata_information['id']}"
```
Bluenix2, line 157: Can all other embeds also get this URL attribute?

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+        kata_embed.add_field(name="Difficulty", value=kata_information['rank']['name'], inline=False)
+        return kata_embed
+
+    @staticmethod
+    def language_embed(kata_information: dict) -> Embed:
+        """Creates the 'language embed' which displays all languages the kata supports."""
+        languages = '\n'.join(map(str.title, kata_information['languages']))
+        language_embed = Embed(
+            title=f"Supported Languages for {kata_information['name']}",
```
Bluenix2, line 178: What about keeping the title the same but moving the `Supported Languages:` part into the description?

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,319 @@
...
+        """
+        Creates the discord.py View for the Discord message components (dropdowns and buttons).
+
+        The discord UI is implemented onto the embed, where the user can choose what information about the kata they
+        want, along with a link button to the kata itself.
+        """
+        view = ui.View()
+        view.add_item(dropdown)
+        view.add_item(ui.Button(label="View the Kata", url=link))
+        return view
```
Bluenix2, line 244: Should we perhaps add an [`on_timeout`](https://discordpy.readthedocs.io/en/master/api.html#discord.ui.View.on_timeout) that will edit the message to have disabled components?

Compared to the bot's usual trash-can emoji it would be removed, we should signal to the user somehow that the bot is no longer accepting input.

Bluenix2: UPDATE: This is more complicated than I anticipated so I think we can skip this.

bot/exts/utilities/challenges.py
Shom770: For all of these, or just the "Supported Languages" embed

bot/exts/utilities/challenges.py
onerandomusername: This is actually not that complicated. At the end of the command, we want to the wait for the instance of the view to stop.

This is easily achieved with this method: https://discordpy.readthedocs.io/en/master/api.html?highlight=view#discord.ui.View.wait

After this, we can edit the message to remove the view, by running `await message.edit(view=None)`

bot/exts/utilities/challenges.py
Bluenix2: I think it'll work for all other embeds yeah? Try it out, if you prefer it like it already is then we can keep it.

I am just coming with different ideas because I am not sure if it feels quite right.

onerandomusername:
copying this from discord so its easier to reference:
  
> It's highly worth removing them. Internally buttons and their names are referred to with a custom ID, which fwiw, should be manually set instead of setting the library decide--this will be useful in the end and I can explain why later.
  
> When the bot receives an interaction over the gateway, it only receives the message and the button ID. this means that we'd be getting a lot of old requests if a user used an old button. It's not worth the possible extra requests by leaving them active.
  
> For the custom ids, these determine which button is which. I personally recommend giving every button or select its own ID, since the api itself requires them. If they aren't provided discord py does a very rudimentary random generate which can easily cause future problems if we have enough selects. It's best to give all a custom ID. IMO, this should be name spaced, eg "challenges_select"

source: https://canary.discord.com/channels/267624335836053506/635950537262759947/891520803143823401

Bluenix2:
Can we change this to disable the items instead? That way they appear greyed out, [see this attribute](https://discordpy.readthedocs.io/en/master/api.html#discord.ui.Button.disabled).

Instead of removing the view, just disable all of its children (the buttons and select menu) and edit it to that view then return.

This works as well, so if you prefer the current behaviour feel free to keep it.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+from bs4 import BeautifulSoup
+from discord import Embed, Interaction, SelectOption, ui
+from discord.ext import commands
+
+from bot.bot import Bot
+from bot.constants import Colours, Emojis, NEGATIVE_REPLIES
+
+logger = logging.getLogger(__name__)
+API_ROOT = "https://www.codewars.com/api/v1/code-challenges/{kata_id}"
+# Map difficulty for the kata to color we want to display in the embed.
```
Bluenix2, line 15:
```suggestion
API_ROOT = "https://www.codewars.com/api/v1/code-challenges/{kata_id}"

# Map difficulty for the kata to color we want to display in the embed.
```

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+            ),
+            SelectOption(
+                label='Tags',
+                description='See what categories this kata falls under!',
+                emoji=Emojis.stackoverflow_tag
+            ),
+            SelectOption(
+                label='Other Information',
+                description='See how other people performed on this kata and more!',
+                emoji='ℹ'
```
Bluenix2, line 60: Look over the quotes here, Python Discord uses double-quotes.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+        # We map the option label to the embed instance so that it can be easily looked up later in O(1)
+        self.mapping_of_embeds = {
+            "Main Information": main_embed,
+            "Languages": language_embed,
+            "Tags": tags_embed,
+            "Other Information": other_info_embed,
+        }
+
+        super().__init__(
+            placeholder='See more information regarding this kata',
```
Bluenix2, line 73: This one should be changed as well.

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+                return error_embed
+
+            soup = BeautifulSoup(await response.text(), features="lxml")
+            first_kata_div = await to_thread(soup.find_all, "div", class_="item-title px-0")
+
+            if not first_kata_div:
+                raise commands.BadArgument("No katas could be found with the filters provided.")
+            elif len(first_kata_div) >= 3:
+                first_kata_div = choice(first_kata_div[:3])
+            elif 'q=' not in search_link:
```
Bluenix2, line 126: Here as well

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+        kata_description = kata_information["description"]
+        kata_url = f"https://codewars.com/kata/{kata_information['id']}"
+
+        # Ensuring it isn't over the length 1024
+        if len(kata_description) > 1024:
+            kata_description = "\n".join(kata_description[:1000].split("\n")[:-1]) + "..."
+            kata_description += f" [continue reading]({kata_url})"
+
+        kata_embed = Embed(
+            title=kata_information['name'],
```
Bluenix2, line 165: See my comment about quotes

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+            description=kata_description,
+            color=MAPPING_OF_KYU[int(kata_information['rank']['name'].replace(' kyu', ''))],
```
Bluenix2, line 167: See my comment about quotes

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+            url=kata_url
+        )
+        kata_embed.add_field(name="Difficulty", value=kata_information['rank']['name'], inline=False)
```
Bluenix2, line 170: See my comment about quotes

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+        return kata_embed
+
+    @staticmethod
+    def language_embed(kata_information: dict) -> Embed:
+        """Creates the 'language embed' which displays all languages the kata supports."""
+        kata_url = f"https://codewars.com/kata/{kata_information['id']}"
+
+        languages = '\n'.join(map(str.title, kata_information['languages']))
```
Bluenix2, line 178: See my comment about quotes

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+        language_embed = Embed(
+            title=kata_information['name'],
```
Bluenix2, line 180: See my comment about quotes

bot/exts/utilities/challenges.py
```diff
@@ -0,0 +1,335 @@
...
+    @staticmethod
+    def tags_embed(kata_information: dict) -> Embed:
+        """
+        Creates the 'tags embed' which displays all the tags of the Kata.
+
+        Tags explain what the kata is about, this is what codewars.com calls categories.
+        """
+        kata_url = f"https://codewars.com/kata/{kata_information['id']}"
+
+        tags = '\n'.join(kata_information['tags'])
```
Bluenix2, line 196: You get the point, I won't spam you with all references.

Bluenix2 (changes requested):
```python
sir-lancebot  | In components.0.components.1: The specified component exceeds the maximum width
sir-lancebot  | Traceback (most recent call last):
sir-lancebot  |   File "/usr/local/lib/python3.9/site-packages/discord/ext/commands/core.py", line 167, in wrapped
sir-lancebot  |     ret = await coro(*args, **kwargs)
sir-lancebot  |   File "/bot/bot/exts/utilities/challenges.py", line 331, in challenge
sir-lancebot  |     await original_message.edit(embed=kata_embed, view=kata_view)
sir-lancebot  |   File "/usr/local/lib/python3.9/site-packages/discord/message.py", line 1289, in edit
sir-lancebot  |     data = await self._state.http.edit_message(self.channel.id, self.id, **payload)
sir-lancebot  |   File "/usr/local/lib/python3.9/site-packages/discord/http.py", line 335, in request
sir-lancebot  |     raise HTTPException(response, data)
sir-lancebot  | discord.errors.HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body
sir-lancebot  | In components.0.components.1: The specified component exceeds the maximum width
```

Bluenix2 (changes requested):
Whenever I invoke the `.challenge` command then try to pick a selection option I get `This interaction failed`. Not sure why

Shom770:
@Bluenix2 I'm not sure why you're getting an interaction failed message, nothing is going wrong on my end? Did you spam it?

Bluenix2:
Yes I did, I opened a lot of views.

That said I realize now what's wrong and I'll test your PR again today.

brad90four (approved):
I was able to test it out this morning. Looks like it works as intended, a pretty cool command! I like the dropdowns and the views.

Bluenix2 (approved):
Played around with it, I am satisfied with it now haha. Good job.

Thanks Shom.

Shom770:
Thank you so much @brad90four and @Bluenix2!

PR #877, opened 2021-09-20
Make the discord.py datetimes offset-naive

Akarys42:
You know the drill, due to discord.py 2.0a0 datetimes are now offset-aware, breaking some code.

Closes python-discord/sir-lancebot#875

### DIFF
--- a/bot/exts/events/hacktoberfest/hacktober-issue-finder.py
+++ b/bot/exts/events/hacktoberfest/hacktober-issue-finder.py
@@ -52,10 +52,10 @@ class HacktoberIssues(commands.Cog):
     async def get_issues(self, ctx: commands.Context, option: str) -> Optional[dict]:
         """Get a list of the python issues with the label 'hacktoberfest' from the Github api."""
         if option == "beginner":
-            if (ctx.message.created_at - self.cache_timer_beginner).seconds <= 60:
+            if (ctx.message.created_at.replace(tzinfo=None) - self.cache_timer_beginner).seconds <= 60:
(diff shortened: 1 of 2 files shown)


ChrisLovering (approved):

wookie184 (changes requested):
I think we'd have the same problem here, correct me if i'm wrong
 https://github.com/python-discord/sir-lancebot/blob/8a410f3abd39a1b48c514d32651a50d4bdced492/bot/exts/events/hacktoberfest/hacktober-issue-finder.py#L55-L58

Akarys42:
> I think we'd have the same problem here, correct me if i'm wrong
> https://github.com/python-discord/sir-lancebot/blob/8a410f3abd39a1b48c514d32651a50d4bdced492/bot/exts/events/hacktoberfest/hacktober-issue-finder.py#L55-L58

Indeed, it is a bit of a kaizen change but who cares. Changed in 24e099e.

wookie184 (changes requested):
Just noticed it for the hacktoberissues bit it also needs to be changed here, line 89
```python
            if option == "beginner":
                self.cache_beginner = data
                self.cache_timer_beginner = ctx.message.created_at
            else:
                self.cache_normal = data
                self.cache_timer_normal = ctx.message.created_at
  ```
  Tested after that change and seems to work so that should be all. Alternatively I guess the original datetime set in the `__init__` could have been changed so it is timezone aware, idk what is good or bad though and haven't tried it.

wookie184 (approved):

PR #880, opened 2021-09-24
Add a 5 minute cooldown to the topic command

### LINKED ISSUE #868: Add a per-channel cooldown for .topic
## Description & Reasoning
The `.topic` command provides conversation starters for our channels. It's very useful for getting things back on topic or to bring up a new thing to talk about. However, I've noticed that some people will just use this command as a rapid-fire answer mechanism, so they end up using `.topic` several times in as many minutes. 

That's not really the intent of this command and is quite spammy, especially in already fast moving channels like python-general. I propose adding some sort of cooldown to the command to allow people to actually discuss the topic rather than fire one-word answers and get the next topic.

## Proposed Implementation
Add the appropriate cooldown decorator, making it per-channel.

ChrisLovering:
## Relevant Issues

Closes #868

## Description

Using the command while it's on cooldown will hit the error handler, which sends an error message showing how long is left on the cooldown, which is deleted after 7.5 seconds.
image

## Did you:

### DIFF
--- a/bot/exts/utilities/conversationstarters.py
+++ b/bot/exts/utilities/conversationstarters.py
@@ -36,6 +36,7 @@ class ConvoStarters(commands.Cog):
     """General conversation topics."""
 
     @commands.command()
+    @commands.cooldown(1, 300, commands.BucketType.channel)
     @whitelist_override(channels=ALL_ALLOWED_CHANNELS)
     async def topic(self, ctx: commands.Context) -> None:
         """


Akarys42:
I'm not a fan of this solution. I think there are relevant reasons to use two topic commands in a five minutes span. I'd propose one of the following:
- Reduce to one minute
- Add a re-roll reaction that can be only used by the command runner

bot/exts/utilities/conversationstarters.py
```diff
@@ -36,6 +36,7 @@ class ConvoStarters(commands.Cog):
     """General conversation topics."""
 
     @commands.command()
+    @commands.cooldown(1, 300, commands.BucketType.channel)
```
brad90four, line 39: What if the cool down was only per user?
```suggestion
    @commands.cooldown(1, 300, commands.BucketType.user)
```

ChrisLovering: The reason for a channel cooldown was so that multiple users can't spam the topic command in a single channel

bot/exts/utilities/conversationstarters.py
```diff
@@ -35,35 +38,75 @@
...
+    @whitelist_override(channels=ALL_ALLOWED_CHANNELS)
+    async def topic(self, ctx: commands.Context) -> None:
+        """
+        Responds with a random topic to start a conversation.
+
+        Allows the refresh of a topic by pressing an emoji.
+        """
+        message = await ctx.send(embed=self._build_topic_embed(ctx.channel.id))
+        self.bot.loop.create_task(self._listen_for_refresh(message))
```
Bluenix2, line 107: Isn't the command itself a task?

bot/exts/utilities/conversationstarters.py
```diff
@@ -35,35 +38,75 @@
...
+
+    def _predicate(self, message: discord.Message, reaction: discord.Reaction, user: discord.User) -> bool:
+        right_reaction = (
+            user != self.bot.user
+            and reaction.message.id == message.id
+            and str(reaction.emoji) == "🔄"
+        )
+        if not right_reaction:
+            return False
+        is_moderator = any(role.id in MODERATION_ROLES for role in getattr(user, "roles", []))
```
Bluenix2, line 75:
```suggestion
            return False

        is_moderator = any(role.id in MODERATION_ROLES for role in getattr(user, "roles", []))
```
Mostly a styling issue

Bluenix2 (commented):
I don't really have much else to comment, looks good otherwise.

bot/exts/utilities/conversationstarters.py
ChrisLovering: My reasoning for adding this as a task onto the loop, rather than awaiting it directly was so that the refresh loop isn't considered as part of the command invocation directly, this mostly matters for things like timing how long a command takes to complete,

bot/exts/utilities/conversationstarters.py
ChrisLovering: Force pushed to the PR

PR #885, opened 2021-09-26
Allow everyone to use the `.bm` command everywhere

### LINKED ISSUE #827: Allow bookmark command in every channel

Objectivitix:
## Relevant Issues
Closes #827

## Description
Allowed the `.bm` command to be used everywhere by using our whitelist overrider.

## Did you:

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -280,12 +280,12 @@ if Client.month_override is not None:
 
 
 class Roles(NamedTuple):
+    owner = 267627879762755584
     admin = int(environ.get("BOT_ADMIN_ROLE_ID", 267628507062992896))
     moderator = 267629731250176001
-    owner = 267627879762755584
     helpers = int(environ.get("ROLE_HELPERS", 267630620367257601))
     core_developers = 587606783669829632
-    everyone = 267624335836053506
+    everyone = int(environ.get("BOT_GUILD", 267624335836053506)) 
 
 
 class Tokens(NamedTuple):
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -285,6 +285,7 @@ class Roles(NamedTuple):
     owner = 267627879762755584
     helpers = int(environ.get("ROLE_HELPERS", 267630620367257601))
     core_developers = 587606783669829632
+    everyone = 267624335836053506
 
 
 class Tokens(NamedTuple):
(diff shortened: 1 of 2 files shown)


HassanAbouelela (approved):
Very nice, thanks!

bot/constants.py
```diff
@@ -285,6 +285,7 @@ class Roles(NamedTuple):
     owner = 267627879762755584
     helpers = int(environ.get("ROLE_HELPERS", 267630620367257601))
     core_developers = 587606783669829632
+    everyone = 267624335836053506
```
onerandomusername, line 288: this should probably copy from https://github.com/python-discord/sir-lancebot/blob/522bea4fe05e74dc3d6a02f97585221a9da7cd2c/bot/constants.py#L133

bot/exts/utilities/bookmark.py
```diff
@@ -87,8 +86,8 @@ async def send_reaction_embed(
         await message.add_reaction(BOOKMARK_EMOJI)
         return message
 
-    @whitelist_override(channels=WHITELISTED_CHANNELS, categories=WHITELISTED_CATEGORIES)
     @commands.command(name="bookmark", aliases=("bm", "pin"))
+    @whitelist_override(roles=(Roles.everyone,))
```
onerandomusername, line 90: This should go above commands.command to be consistent with how it was before; also keeps git diff to a minimum.

onerandomusername (changes requested):
Looks functional, two code suggestions which seem easy enough to fix.

bot/exts/utilities/bookmark.py
Objectivitix: It was above, and that's why it didn't work.

HassanAbouelela: This actually needs to go below the command decorator, otherwise it has no effect. I believe it did not work previously.

Objectivitix: (yes, no one tested the old PR besides mina)

bot/constants.py
Objectivitix: I agree.

bot/exts/utilities/bookmark.py
onerandomusername: Ah, okay.

onerandomusername (approved):

ChrisLovering (approved):

PR #891, opened 2021-10-03
Ignore all bot messages for spooky react

gustavwilliam:
Spooky reactions are now limited to users. The example below was previously possible, but won't happen anymore:
Screenshot 2021-10-03 at 15 35 28

The bot previously only ignored its own messages, but now ignores messages from all bots. Even though we love our bots, they simply aren't supposed to be a part of this Halloween tradition. Sorry Xith!


Akarys42 (approved):
Excellent mister gurkan

D0rs4n (approved):

PR #892, opened 2021-10-04
Monkey patch http.send_typing to catch 403s

ChrisLovering:
## Description

Sometimes discord turns off typing events by throwing 403's, so we should catch those

## Did you:


bot/monkey_patches.py
```diff
@@ -0,0 +1,78 @@
...
+            raise TypeError("Root aliases of a group must be a list or a tuple of strings.")
+
+
+def patch_typing() -> None:
+    """
+    Sometimes discord turns off typing events by throwing 403's.
+
+    Handle those issues by patching the trigger_typing method so it ignores 403's in general.
+    """
+    log.info("Patching send_typing, which should fix things breaking when discord disables typing events. Stay safe!")
```
jchristgit, line 61: IMO this should be a `log.debug` or `log.trace`, no? This is rather uninteresting, I think.

jchristgit (approved):
Yes, very good. Just a small nitpick. Out of curiosity: Why `datetime.now()` instead of `utcnow()`?

bot/monkey_patches.py
ChrisLovering: this is true, we should probably update it in the other repos too

ChrisLovering:
> Yes, very good. Just a small nitpick. Out of curiosity: Why `datetime.now()` instead of `utcnow()`?

no real reason for using it over utcnow(), is utcnow() preferable?

jchristgit:
> no real reason for using it over utcnow(), is utcnow() preferable?

I think in case the time zone changes we might get weird log lines here. That happens like twice a year though, and timezones are insane anyways, just like unicode. So personally I prefer it because UTC does not change.

jchristgit:
Great work, Chris. Thanks!

Xithrius (approved):

PR #901, opened 2021-10-12
Madlibs

### LINKED ISSUE #847: Add Madlibs game
## Description

I would like to create a Madlibs game command for Sir Lancebot.

## Reasoning

Most often than not, the output of the game (the story with the blanks filled in) will sound really funny and/or silly, so I'm sure many people will want to play it!

## Proposed Implementation

The command for this game will be `.madlibs`

When the command, `.madlibs`, is run, the bot will ask the user to enter a word that fits a random part of speech (e.g. noun, adjective, verb, plural noun, etc.). The user will simply send a message with any word that fits the given part of speech.

In the API, the computer (bot in this case) chooses a random number of user inputs to use for the game and a random story. This is what I would like my program to do as well.

Once the user has finished entering all the input prompts, the bot should send a message with the completed story using the words that the user entered.

As part of the implementation, I will be using this Madlibs API: https://madlibz.herokuapp.com/api

DMFriends:
## Relevant Issues
Closes #847 

## Description

This PR is designed to add a fun Madlibs game to Sir Lancebot.

When the command, `.madlibs`, is run, the bot will ask the user to enter a word that fits a random part of speech (e.g. noun, adjective, verb, plural noun, etc.). The user will simply send a message with any word that fits the given part of speech.

The bot chooses a random number of user inputs to use for the game and a random story.

Once the user has finished entering all the input prompts, the bot should send a message with the completed story using the words that the user entered.

## Did you:


Xithrius:
@DMFriends @Bluenix2 What's the progress of this PR so far?

DMFriends:
> @DMFriends @Bluenix2 What's the progress of this PR so far?

I think I'm almost done. I'll probably need another day or 2 to get this done.

Xithrius:
@DMFriends very well. Thanks for the reply!

DMFriends:
Np!

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,116 @@
...
+        Madlibs is a game where the player is asked to enter a word that
+        fits a random part of speech (e.g. noun, adjective, verb, plural noun, etc.)
+        a random amount of times, depending on the story chosen by the bot at the beginning.
+        """
+        random_template = choice(self.templates)
+
+        def author_check(message: discord.Message) -> bool:
+            return message.channel.id == ctx.channel.id and message.author.id == ctx.author.id
+
+        original_message = await ctx.send("Loading your Madlibs game...")
```
Shom770, line 72: I think the transition because it'll edit so fast it'll look a bit 'wonky' if that makes sense. Going from pure text to an embed might look that way.

You could set up an embed for this:
```suggestion
        
        loading_embed = discord.Embed(title="Madlibs", color=Colours.python_blue)
        original_message = await ctx.send(embed=loading_embed)
```

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,116 @@
...
+        with open(madlibs_stories) as file:
+            return json.load(file)["templates"]
+
+    @staticmethod
+    def madlibs_embed(part_of_speech: str, number_of_inputs: int) -> discord.Embed:
+        """Method to have the bot send an embed with the game information."""
+        madlibs_embed = discord.Embed(
+            title="Madlibs",
+            color=Colours.python_blue,
+        )
```
Shom770, line 44: This will not exceed the line limit, so I don't think it's necessary to split this into different lines.
```suggestion
        madlibs_embed = discord.Embed(title="Madlibs", color=Colours.python_blue)
```

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,116 @@
...
+
+        madlibs_embed.set_footer(text=f"Inputs remaining: {number_of_inputs}")
+
+        return madlibs_embed
+
+    @commands.command()
+    async def madlibs(
+        self,
+        ctx: commands.Context,
+    ) -> None:
```
Shom770, line 59: This should also fit in one line, so the separating isn't necessary.
```suggestion
    async def madlibs(self, ctx: commands.Context) -> None:
```

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,116 @@
...
+from asyncio import TimeoutError
+from pathlib import Path
+from random import choice
+from typing import TypedDict
+
+import discord
+from discord.ext import commands
+
+from bot.bot import Bot
+from bot.constants import Colours
```
Shom770, line 11: See the comment below on line 86. If you want to accept that suggestion, you'd also want to accept this suggestion for it to work, as it requires an import from bot.constants.
```suggestion
from bot.constants import Colours, NEGATIVE_REPLIES
```

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,116 @@
...
+            inputs_left = len(random_template["blanks"]) - i
+
+            madlibs_embed = self.madlibs_embed(part_of_speech, inputs_left)
+            await original_message.edit(embed=madlibs_embed)
+
+            try:
+                message = await self.bot.wait_for(event="message", check=author_check, timeout=TIMEOUT)
+            except TimeoutError:
+                timeout_embed = discord.Embed(
+                    title="Timeout!",
```
Shom770, line 86: This is more of a consideration, up to you, but usually, it would be a random one of the negative messages to be in uniform with the other 'error' embeds
```suggestion
                    title=choice(NEGATIVE_REPLIES),
```

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,116 @@
...
+                    description="Uh oh! Looks like the bot timed out! Please try again later.",
+                    color=Colours.soft_red
+                )
+
+                await ctx.send(embed=timeout_embed)
+                return
+
+            word = message.content
+            submitted_words.append(" " + f"__{word}__")
```
Shom770, line 95: Hmm, this would also work and is a bit cleaner (instead of concatenating, you add a space before the __ in the f string) but this is a nitpick and is also just a consideration, not anything necessary.
```suggestion
            submitted_words.append(f" __{word}__")
```

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,116 @@
...
+
+        story = []
+        for value, blank in zip(random_template["value"], submitted_words):
+            story.append(f'{value}{blank}')
```
Shom770, line 99: The guide would prefer that you use `"` instead of `'` when possible, also another nitpick -
```suggestion
            story.append(f"{value}{blank}")
```

Shom770 (changes requested):
Good job on this PR! Here are a few comments I had:

bot/exts/fun/madlibs.py
Bluenix2: @DMFriends this was marked resolved but I don't see the changes applied in the "Changed Files" tab.

Have you pushed your changes?

DMFriends: Just pushed @Bluenix2

bot/exts/fun/madlibs.py
```diff
@@ -63,7 +64,11 @@ async def madlibs(self, ctx: commands.Context) -> None:
         def author_check(message: discord.Message) -> bool:
             return message.channel.id == ctx.channel.id and message.author.id == ctx.author.id
 
-        original_message = await ctx.send("Loading your Madlibs game...")
+        loading_embed = discord.Embed(
+            title="Madlibs", description="Loading your Madlibs game...", color=Colours.python_blue
+        )
+        original_message = await ctx.send(embed=loading_embed)
+        time.sleep(1)
```
Bluenix2, line 71: This will block the rest of the bot, besides, why sleep?

Bluenix2 (changes requested):

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,110 @@
...
+
+            madlibs_embed = self.madlibs_embed(part_of_speech, inputs_left)
+            await original_message.edit(embed=madlibs_embed)
+
+            try:
+                message = await self.bot.wait_for(event="message", check=author_check, timeout=TIMEOUT)
+            except TimeoutError:
+                timeout_embed = discord.Embed(
+                    title="Timeout!",
+                    description="Uh oh! Looks like the bot timed out! Please try again later.",
```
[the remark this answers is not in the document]
DMFriends: @Bluenix2 Well, what I meant this to mean is "the bot timed out _because_ the user took too long to respond." Does that make sense?

bot/exts/fun/madlibs.py
DMFriends: Why will this block the bot?

DMFriends: I thought `sleep` would make it more realistic that it's "loading the game", in a sense.

bot/exts/fun/madlibs.py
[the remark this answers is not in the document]
Bluenix2: Yeah technically that's true but I don't think that's clear to the user.

This sounds like something broke in the bot, such as some form of connection somewhere.

bot/exts/fun/madlibs.py
Bluenix2: > I thought `sleep` would make it more realistic that it's "loading the game", in a sense.

No we don't want it to be realistic, the faster the better. It is i just for getting a message to start with.

DMFriends: Alright, I will remove it then.

bot/exts/fun/madlibs.py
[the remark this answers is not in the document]
DMFriends: Ah I see, I can fix the message to make it more clear, no problem.

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+        madlibs_embed.set_footer(text=f"Inputs remaining: {number_of_inputs}")
+
+        return madlibs_embed
+
+    @commands.Cog.listener()
+    async def on_message_edit(self, _: discord.Message, after: discord.Message) -> None:
+        """A listener that checks for edits to messages from the user."""
+        self.edited_content[after.id] = after.content
```
Bluenix2, line 56: We need some form of if-statement so that not all messages get added here.

I have an idea, in discord.py there is a way to say how many commands can get used at once. If we use this we can limit the amount of commands to one at a time, then completely reset this dictionary each time the command ends.

Maybe with a before and after hook we can setup so that some form of attribute gets set that says that a command is ongoing? That way it'll run even if the command errors out.

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,124 @@
...
+                    description="Uh oh! You took too long to respond! Please try again later.",
+                    color=Colours.soft_red
+                )
+
+                await ctx.send(embed=timeout_embed)
+                return
+
+            word = message.content
+
+            submitted_words[message.id] = word
```
Bluenix2, line 99: You can just simplify it like this:

```suggestion
            submitted_words[message.id] = message.content
```

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,124 @@
...
+
+        blanks = []
+        for msg_id in submitted_words:
+            blanks.append(self.edited_content.pop(msg_id, submitted_words[msg_id]))
```
Bluenix2, line 103: We can make this a list comprehension, I am not sure how much experience you have this those. It would look something like this:

```suggestion
        blanks = [self.edited_content.pop(msg_id, submitted_words[msg_id] for msg_id in submitted_words]
```

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,124 @@
...
+
+        story = []
+        for value, blank in zip(random_template["value"], blanks):
+            story.append(f"{value}__{blank}__")
```
Bluenix2, line 107: This could also be a list comprehension - but I think it would become to long so it is probably fine.

bot/exts/fun/madlibs.py
DMFriends: I am a bit familiar with list comprehensions. I will accept this suggestion.

bot/exts/fun/madlibs.py
DMFriends: Can't I just use the `author_check` you added to my main program? @Bluenix2

bot/exts/fun/madlibs.py
DMFriends: Yeah, I'm afraid it will go over the PEP 8 line limit if I do. I'll resolve this, if you don't mind. @Bluenix2

ChrisLovering: List comprehensions don't need to be on one line. ```
story = [
    f"{value}__{blank}__"
    value, blank in zip(random_template["value"], blanks)
]
```
I'm not saying whether you should or shouldn't do this, I just noticed this thread so thought I'd comment.

Comment (DMFriends) [reply bot/exts/fun/madlibs.py:103]: Well yes it could, I just think it looks nicer the way I have it now. @ChrisLovering

File: bot/exts/fun/madlibs.py (hunk 10 above)
Comment (Bluenix2) [reply bot/exts/fun/madlibs.py:56]: This listener is not aware of that author check.

Comment (DMFriends) [reply bot/exts/fun/madlibs.py:56]: Well right now it isn't of course, but could I add it here? Would it work? @Bluenix2

Comment (Bluenix2) [reply bot/exts/fun/madlibs.py:56]: Sadly no, because there's no `ctx`.

Comment (DMFriends) [reply bot/exts/fun/madlibs.py:56]: Ahhh, ok. Well could I add it as a parameter to the function? Would that work? @Bluenix2

Comment (Bluenix2) [reply bot/exts/fun/madlibs.py:56]: No that wouldn't work either because discord.py creates the `ctx` for each command. So we need to somehow tell the listener each command running - kind of like how/what I suggested!

Comment (DMFriends) [reply bot/exts/fun/madlibs.py:56]: Ok, I see. I'll think about what you suggested then and if I have any questions, I will ask. Thanks!

File: bot/exts/fun/madlibs.py [hunk 14]
```diff
@@ -0,0 +1,120 @@
...
+    @staticmethod
+    def _load_templates() -> list[MadlibsTemplate]:
+        madlibs_stories = Path("bot/resources/fun/madlibs_templates.json")
+
+        with open(madlibs_stories) as file:
+            return json.load(file)
+
+    @staticmethod
+    def madlibs_embed(part_of_speech: str, number_of_inputs: int) -> discord.Embed:
+        """Method to have the bot send an embed with the game information."""
```
Bluenix2, line 41: This is no longer true, all this does is generate the embed.

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,120 @@
...
+
+            submitted_words[message.id] = message.content
+
+        blanks = [self.edited_content.pop(msg_id, submitted_words[msg_id]) for msg_id in submitted_words]
+
+        story = []
+        for value, blank in zip(random_template["value"], blanks):
+            story.append(f"{value}__{blank}__")
+
+        story.append(random_template["value"][-1])
```
Bluenix2, line 105: Since you'll be changing other code can you also take the to add a comment about why we do this?

I know why as we've been talking about it, but I don't know how clear it is to others.

Not the biggest issues though but shouldn't take too much effort if you're changing other things.

DMFriends: Sure, I can do that.

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,132 @@
...
+                message = await self.bot.wait_for(event="message", check=author_check, timeout=TIMEOUT)
+            except TimeoutError:
+                timeout_embed = discord.Embed(
+                    title=choice(NEGATIVE_REPLIES),
+                    description="Uh oh! You took too long to respond! Please try again later.",
+                    color=Colours.soft_red
+                )
+
+                await ctx.send(embed=timeout_embed)
+                return
```
Bluenix2, line 104: Dont forget to cleanup here (remove the check and remove messages from `edited_content` that has been added as a result of the check)!

We don't want to be leaking these and creating a huge dictionary that keeps all messages alive.

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,132 @@
...
+        blanks = [self.edited_content.pop(msg_id, submitted_words[msg_id]) for msg_id in submitted_words]
+
+        story = []
+        for value, blank in zip(random_template["value"], blanks):
+            story.append(f"{value}__{blank}__")
+
+        story.append(random_template["value"][-1])
+        # In each story template, there is always one more "value"
+        # (fragment from the story) than there are blanks (words that the player enters)
+        # so we need to compensate by appending the last line of the story again.
```
Bluenix2, line 117: Aah, usually you have the comment before the code it is referring to 😅

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,132 @@
...
+
+        story_embed = discord.Embed(
+            title=random_template["title"],
+            description="".join(story),
+            color=Colours.bright_green
+        )
+
+        story_embed.set_footer(text=f"Generated for {ctx.author}", icon_url=ctx.author.display_avatar.url)
+
+        await ctx.send(embed=story_embed)
```
Bluenix2, line 127: I don't see you removing the check in the end, can you do that?

Shom770 (dismissed):
The game itself works pretty well! Two suggestions from me however -
1 ) Change the __text__ to `text` when the embed with all the words used and formatted into the prompt
2 ) We could use validation for checking whether something is a verb, noun, or adjective? Not sure about plural nouns or places, but we could use `nltk.corpus.wordnet` for validation. This is something that is more of a consideration as it depends on what you want to do.

DMFriends:
> The game itself works pretty well! Two suggestions from me however - 1 ) Change the **text** to `text` when the embed with all the words used and formatted into the prompt 2 ) We could use validation for checking whether something is a verb, noun, or adjective? Not sure about plural nouns or places, but we could use `nltk.corpus.wordnet` for validation. This is something that is more of a consideration as it depends on what you want to do.

This comment no longer applies.

bot/exts/fun/madlibs.py
```diff
@@ -0,0 +1,140 @@
...
+
+        self.checks.remove(author_check)
```
Bluenix2, line 135: This needs to happen before sending the message (preferrably directly after defining `blanks`) to avoid a race condition.

DMFriends: Sounds good, will do.

bot/resources/fun/madlibs_templates.json
```diff
@@ -0,0 +1,135 @@
...
+        },
+        {
+             "title": "Star Wars",
+             "blanks": ["adjective", "noun", "adjective", "noun; place", "adjective", "adjective", "adjective", "adjective", "plural noun", "adjective", "plural noun", "plural noun", "adjective",
+                        "noun", "verb", "adjective", "verb", "plural noun; type of job", "adjective", "verb", "adjective"],
+             "value": ["Star Wars is a ", " ", " of ", " versus evil in a ", " far far away. There are ", " battles between ", " ships in ", " space and ", " duels with ", " called ",
+                       " sabers. ", " called \"droids\" are helpers and ", " to the heroes. A ", " power called The ", " ", "s people to do ", " things, like ", " ", " use The Force for the ",
+                       " side and the Sith ", " it for the ", " side."]
+        }
+]
```
Bluenix2, line 135: Can you look over these again, the result are a bit unsatisfactory at times. For example, there's inconsistencies with where "a" is used:
image

*It says "pizzeria" but "a mouse"*

bot/resources/fun/madlibs_templates.json
```diff
@@ -0,0 +1,135 @@
...
+                      " down the house. Then the wolf knocked on the third ", " pig's door. \"Let me in or I'll blow your house down!\" The little pig didn't so the wolf ",
+                      " and ", ". He could not blow the house down. All the pigs went to live in the ", " house and they all ", " happily ever after."]
+        },
+        {
+            "title": "Talk Like a Pirate",
+            "blanks": ["noun", "adjective", "verb", "adverb", "noun", "adjective", "plural noun", "plural noun", "plural noun", "part of the body", "noun",
+                        "noun", "noun", "noun", "part of the body"],
+            "value": ["Ye can always pretend to be a bloodthirsty ", ", threatening everyone by waving yer ", " sword in the air, but until ye learn to ",
+                      " like a pirate, ye'll never be ", " accepted as an authentic ", ". So here's what ye do: Cleverly work into yer daily conversations ",
+                      " pirate phrases such as \"Ahoy there, ", ", \"\"Avast, ye ", ",\" and \"Shiver me ", ".\" Remember to drop all yer gs when ye say such words as sailin', spittin', and fightin'. This will give ye a/an ",
```
Bluenix2, line 32: > "Ahoy there, ____, "

Shouldn't it be:

> "Ahoy there, ___",

bot/resources/fun/madlibs_templates.json
```diff
@@ -0,0 +1,135 @@
...
+                      " hour(s). We started off on a(n) ", " adventure, but we kept losing the trail. Night began to fall, and when we heard the howls of a ",
+                      ", we began to panic. It was getting darker and our flashlights were running on ", ". I'm sure glad my pet ", ", ", ", was with us. He is one gifted creature, because he was able to guide us back by ",
+                      " the ", " s'mores by the campfire. This year, before setting off on an ", " journey, I'll be sure to have working flashlights - and of course, my gifted pet!"]
+        },
+        {
+            "title": "Weird News",
+            "blanks": ["noun", "place", "verb ending in ing", "noun", "name", "verb", "noun", "verb", "noun", "part of body", "type of liquid", "place", " past tense verb ", "foreign country", "verb", "noun", "past tense verb", "adjective", "verb", "noun", "plural noun"],
+            "value": ["A ", " in ", " was arrested this morning after he was caught ", " in front of ", ". ", " had a history of ", ", but no one-not even his ", "-ever imagined he'd ", " with a ", " stuck in his ", ". After drinking a ", ", cops followed him to a ",
```
Bluenix2, line 67:
```suggestion
            "value": ["A ", " in ", " was arrested this morning after he was caught ", " in front of ", ". ", " had a history of ", ", but no one-not even his ", "ever imagined he'd ", " with a ", " stuck in his ", ". After drinking a ", ", cops followed him to a ",
```

bot/resources/fun/madlibs_templates.json
```diff
@@ -0,0 +1,135 @@
...
+                      " where he reportedly  ", "in the fry machine. Later, a woman from ", " was charged with a similar crime. But rather than ", "with a ", ", she ", "with a ", " dog. Either way, we imagine that after witnessing him ", " with a ", " there are probably a whole lot of ",
```
Bluenix2, line 68:
```suggestion
                      " where he reportedly ", "in the fry machine. Later, a woman from ", " was charged with a similar crime. But rather than ", "with a ", ", she ", "with a ", " dog. Either way, we imagine that after witnessing him ", " with a ", " there are probably a whole lot of ",
```

Bluenix2, line 68:
```suggestion
                      " where he reportedly  ", " in the fry machine. Later, a woman from ", " was charged with a similar crime. But rather than ", "with a ", ", she ", "with a ", " dog. Either way, we imagine that after witnessing him ", " with a ", " there are probably a whole lot of ",
```

Bluenix2, line 68:
```suggestion
                      " where he reportedly  ", "in the fry machine. Later, a woman from ", " was charged with a similar crime. But rather than ", " with a ", ", she ", " with a ", " dog. Either way, we imagine that after witnessing him ", " with a ", " there are probably a whole lot of ",
```

bot/resources/fun/madlibs_templates.json
```diff
@@ -0,0 +1,135 @@
...
+                      " that are going to need some therapy!"]
+        },
+        {
+            "title": "All About Vampires",
+            "blanks": ["adjective", "adjective", "body part", "verb ending in -ing", "verb", "verb", "verb", "noun", "verb", "verb", "body part", "verb", "verb", "verb", "noun", "body part", "adjective"],
+            "value": ["Vampires are ", "! They have ", " ", " for ", " blood. The sun can ", " vampires, so they only ", " at night and ", " during the day. Vampires also don't like ", " so ", " it or ", " it around your ",
+                      " to keep them away. If a vampire ", "s a person and ", "s their blood, they become a vampire, too. The only way to ", " a vampire is with a ", " through the ", ", but ",
```
Bluenix2, line 75:
```suggestion
                      " to keep them away. If a vampire ", " a person and ", "s their blood, they become a vampire, too. The only way to ", " a vampire is with a ", " through the ", ", but ",
```
It's better to say in the blank like "ending with 's'"

bot/resources/fun/madlibs_templates.json
```diff
@@ -0,0 +1,135 @@
...
+                       " and ", " instead of video games. Also, before cars were invented, people actually rode ", "! People read ", " instead of computers and tablets, and sent messages via ",
+                       " that took ", " days to arrive. I wonder how kids will view my life in ", " year(s); maybe they will ride flying cars to school and play with ", " and ", " ", "!"]
+        },
+        {
+             "title": "Star Wars",
+             "blanks": ["adjective", "noun", "adjective", "noun; place", "adjective", "adjective", "adjective", "adjective", "plural noun", "adjective", "plural noun", "plural noun", "adjective",
+                        "noun", "verb", "adjective", "verb", "plural noun; type of job", "adjective", "verb", "adjective"],
+             "value": ["Star Wars is a ", " ", " of ", " versus evil in a ", " far far away. There are ", " battles between ", " ships in ", " space and ", " duels with ", " called ",
+                       " sabers. ", " called \"droids\" are helpers and ", " to the heroes. A ", " power called The ", " ", "s people to do ", " things, like ", " ", " use The Force for the ",
```
Bluenix2, line 132: Same here, instead of that `s` move it to the blank for the uset to input.

Bluenix2 (commented):
It seems like most issues just originated from one specific story, I do have some comments on the other ones!

bot/resources/fun/madlibs_templates.json
DMFriends: Yeah, I agree.

bot/resources/fun/madlibs_templates.json
DMFriends: The dash is supposed to be there because:

"but no one - not even his _____ - ever imagined..."

Does that make sense? @Bluenix2

Bluenix2: Ah, well look how you added spaces around the dash.

DMFriends: Yes, I did notice. Will fix.

Shom770 (approved):

Bluenix2 (approved):
Awesome work the past two months! This is great, thanks DMFriends!

DMFriends:
> Awesome work the past two months! This is great, thanks DMFriends!

Thank you for all your help @Bluenix2 and @Shom770! This is my first time contributing and I really enjoyed working on this. You guys are the best!

Shom770:
> > Awesome work the past two months! This is great, thanks DMFriends!
> 
> 
> 
> Thank you for all your help @Bluenix2 and @Shom770! This is my first time contributing and I really enjoyed working on this. You guys are the best!

No problem!

Akarys42 (approved):
works as intended

onerandomusername (changes requested):
Should only allow a user to invoke the command once at a time. 

Easy fix, decorate the command with this: https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.max_concurrency

```py
@commands.max_concurrency(1, per=commands.BucketType.user)
```

onerandomusername:
It turns out that the bots error handler doesn't actually catch the error raised by a concurrency blocker.

To solve this, we can use local error handlers ([method](https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.Command.error)) ([guide](https://discordpy.readthedocs.io/en/master/ext/commands/commands.html#error-handling))

I've put an implementation together below, since some of this involves digging into our error handler implementation.
```py
    @madlibs.error
    async def handle_madlibs_error(self, ctx: commands.Context, error: commands.CommandError) -> None:
        if isinstance(error, commands.MaxConcurrencyReached):
            await ctx.send("You are already playing madlibs!")
            error.handled = True

```

The reason we set error.handled to True is because in our error handler we use getattr(error, "handler", False) to see if we have already handled the error.

DMFriends:
> It turns out that the bots error handler doesn't actually catch the error raised by a concurrency blocker.
> 
> To solve this, we can use local error handlers ([method](https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.Command.error)) ([guide](https://discordpy.readthedocs.io/en/master/ext/commands/commands.html#error-handling))
> 
> I've put an implementation together below, since some of this involves digging into our error handler implementation.
> ```py
>     @madlibs.error
>     async def handle_madlibs_error(self, ctx: commands.Context, error: commands.CommandError) -> None:
>         if isinstance(error, commands.MaxConcurrencyReached):
>             await ctx.send("You are already playing madlibs!")
>             error.handled = True
> 
> ```
> 
> The reason we set error.handled to True is because in our error handler we use getattr(error, "handler", False) to see if we have already handled the error.

So this would go in the error handler file in the repo, correct?

DMFriends:
> Should only allow a user to invoke the command once at a time. 
> 
> Easy fix, decorate the command with this: https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.max_concurrency
> 
> ```py
> @commands.max_concurrency(1, per=commands.BucketType.user)
> ```

Would this go right after line 65?

onerandomusername:
> > It turns out that the bots error handler doesn't actually catch the error raised by a concurrency blocker.
> > To solve this, we can use local error handlers ([method](https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.Command.error)) ([guide](https://discordpy.readthedocs.io/en/master/ext/commands/commands.html#error-handling))
> > I've put an implementation together below, since some of this involves digging into our error handler implementation.
> > ```python
> >     @madlibs.error
> >     async def handle_madlibs_error(self, ctx: commands.Context, error: commands.CommandError) -> None:
> >         if isinstance(error, commands.MaxConcurrencyReached):
> >             await ctx.send("You are already playing madlibs!")
> >             error.handled = True
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The reason we set error.handled to True is because in our error handler we use getattr(error, "handler", False) to see if we have already handled the error.
> 
> So this would go in the error handler file in the repo, correct?

No, this would go in the madlibs file, since its local to the command.

onerandomusername:
> > Should only allow a user to invoke the command once at a time.
> > Easy fix, decorate the command with this: https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.max_concurrency
> > ```python
> > @commands.max_concurrency(1, per=commands.BucketType.user)
> > ```
> 
> Would this go right after line 65? 

Yeah.

DMFriends:
> > > It turns out that the bots error handler doesn't actually catch the error raised by a concurrency blocker.
> > > To solve this, we can use local error handlers ([method](https://discordpy.readthedocs.io/en/master/ext/commands/api.html#discord.ext.commands.Command.error)) ([guide](https://discordpy.readthedocs.io/en/master/ext/commands/commands.html#error-handling))
> > > I've put an implementation together below, since some of this involves digging into our error handler implementation.
> > > ```python
> > >     @madlibs.error
> > >     async def handle_madlibs_error(self, ctx: commands.Context, error: commands.CommandError) -> None:
> > >         if isinstance(error, commands.MaxConcurrencyReached):
> > >             await ctx.send("You are already playing madlibs!")
> > >             error.handled = True
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The reason we set error.handled to True is because in our error handler we use getattr(error, "handler", False) to see if we have already handled the error.
> > 
> > So this would go in the error handler file in the repo, correct?
> 
> No, this would go in the madlibs file, since its local to the command.

Ah, gotcha. So where exactly does this go? Right after the `async def madlibs`, correct?

Shivansh-007:
> Ah, gotcha. So where exactly does this go? Right after the `async def madlibs`, correct?
>

Yeah, it can go anywhere in the cog, it's just a special decorator like the `on_command_error` but for a cog, which would be called whoever there is an error raised in the cog, also [`error.handled`](https://github.com/python-discord/sir-lancebot/blob/9202b432ccef1c13e476afd892fbd718397afa0c/bot/exts/core/error_handler.py#L51) is set so the error handler doesn't handle the error again.

DMFriends:
> > Ah, gotcha. So where exactly does this go? Right after the `async def madlibs`, correct?
> >
> 
> Yeah, it can go anywhere in the cog, it's just a special decorator like the `on_command_error` method but for a cog, which would be called whenever there is an error raised in the cog, also [`error.handled`](https://github.com/python-discord/sir-lancebot/blob/9202b432ccef1c13e476afd892fbd718397afa0c/bot/exts/core/error_handler.py#L51) is set so the error handler doesn't handle the error again.

Gotcha, thanks.

onerandomusername (approved):
lgtm :D

onerandomusername:
Not changing my approval, but there is this error message on timeout which I don't fully think makes sense.
image

Additonally, I'm not sure about the random titles, as its somewhat unprompted. 

Finally, I suggest mentioning the user with this error as its a timeout.

DMFriends:
> Not changing my approval, but there is this error message on timeout which I don't fully think makes sense. image
> 
> Additonally, I'm not sure about the random titles, as its somewhat unprompted.
> 
> Finally, I suggest mentioning the user with this error as its a timeout.

As you mentioned on [Discord](https://discord.com/channels/267624335836053506/635950537262759947/922305476857823282), it does say "Please try again later" and yes I agree that that the user could try again. To be honest, I could probably change what the embed says as a whole. I'm going to think about this more. Also I think mentioning the user would be good, will add that.

About the random titles, that was something that Shom770 pointed out in a [review](https://github.com/python-discord/sir-lancebot/pull/901#discussion_r748789096) so I added that. It sounds like this is a standard convention for similar projects so I decided to commit that suggestion. If you think I should not have it, I think it's fine either way.

onerandomusername:
> About the random titles, that was something that Shom770 pointed out in a [review](https://github.com/python-discord/sir-lancebot/pull/901#discussion_r748789096) so I added that. It sounds like this is a standard convention for similar projects so I decided to commit that suggestion. If you think I should not have it, I think it's fine either way.

Well, we should keep them then, but I think that message should mention the user at minimum, and remove the please try again later :D

DMFriends:
> > About the random titles, that was something that Shom770 pointed out in a [review](https://github.com/python-discord/sir-lancebot/pull/901#discussion_r748789096) so I added that. It sounds like this is a standard convention for similar projects so I decided to commit that suggestion. If you think I should not have it, I think it's fine either way.
> 
> Well, we should keep them then, but I think that message should mention the user at minimum, and remove the please try again later :D

Yep, I already fixed that locally, will commit sometime soon :)
Everything else besides the "please try again later" is good, right? @onerandomusername

onerandomusername:
Yep!

onerandomusername (approved):

Xithrius:
To whoever eventually hits the merge button: squash.

ChrisLovering (changes requested):

ChrisLovering (approved):

PR #902, opened 2021-10-13
feat: added cowsay

### LINKED ISSUE #667: Add Cowsay Command
## Description

It'll just emulate the cowsay program.

## Reasoning

This would be useful because it would be funny.

## Proposed Implementation

It would have the flags of the original cowsay, such as tux, except maybe the head in ass one.

Kronifer:
## Relevant Issues

Closes #667 

## Description

Added a cowsay command that emulates the classic CLI tool.

## Did you:

### DIFF
--- a/bot/exts/fun/cowsay.py
+++ b/bot/exts/fun/cowsay.py
@@ -1,0 +1,56 @@
+from cowpy import cow
+from discord.ext import commands
+
+from bot.bot import Bot
+
+SUPPORTED_COWS = []
+
+for element in cow.get_cowacters():
+    # The cows in the list are too big for Discord or innapropriate
+    if element[0] not in ["turtle", "stegosaurus", "ghostbusters", "turkey", "mutilated", "kiss"]:
+        SUPPORTED_COWS.append(element[0] if element[0] != "default" else "cow")
+
+
+class Cowsay(commands.Cog):
+    """Cog for the Cowsay command."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @commands.command(
+        help=f"""
+        Generates a cowsay string and sends it. The available characters are:
+
+        {', '.join(SUPPORTED_COWS)}"""
+    )
+    async def cowsay(self, ctx: commands.Context, cowtype: str = "cow", message: str = None) -> None:
+        """Builds a cowsay string and sends it to discord."""
+        if message is None:
+            message = f"I'm a {cowtype}"
+
+        if cowtype == "cow":
+            cowtype = "default"
+
+        if cowtype not in SUPPORTED_COWS:
+            raise commands.BadArgument("The cow submitted is invalid! Please use a valid cow.")
+        rawcow = cow.get_cow(cowtype)
+        message = message.replace("`", "`\u200b")
+
+        # Checks if message is too long
+        if len(message) >= 150 and cowtype != "beavis":
+            raise commands.BadArgument("The given text is too long! Please submit something under 150 characters.")
+
+        elif len(message) >= 100 and cowtype == "beavis":
+            raise commands.BadArgument("The given text is too long! Please submit something under 100 characters.")
+
+        # Tries to get the cow class.
+        try:
+            truecow = rawcow()
+        except TypeError:
+            raise commands.BadArgument("The cow submitted is invalid! Please use a valid cow.")
+        await ctx.send(f"```{truecow.milk(message)}```")
+
+
+def setup(bot: Bot) -> None:
+    """Loads the cowsay cog."""
+    bot.add_cog(Cowsay(bot))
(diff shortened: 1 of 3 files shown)


ChrisLovering:
It has two differing licences in the repo, so until the project decides what license it should have, we don't know what we need to comply with.

onerandomusername:
Both licenses are compatible with our project, so I don't see the issue

Shivansh-007:
How about using cows to send the error texts 😄 

Secondly, regarding the cowpy license issue, it has both [MIT](https://github.com/jeffbuttars/cowpy/blob/master/LICENSE.md) and [Apache](https://github.com/jeffbuttars/cowpy/blob/master/LICENSE) licenses but they are present only on the repo and haven't been uploaded on PyPI.

bot/exts/fun/cowsay.py
```diff
@@ -0,0 +1,56 @@
...
+
+        {', '.join(SUPPORTED_COWS)}"""
+    )
+    async def cowsay(self, ctx: commands.Context, cowtype: str = "cow", message: str = None) -> None:
+        """Builds a cowsay string and sends it to discord."""
+        if message is None:
+            message = f"I'm a {cowtype}"
+
+        if cowtype == "cow":
+            cowtype = "default"
```
brad90four, line 32: Right now this isn't working if I don't include the cowtype after calling the command, since it is trying to pass the first word in the message as the cowtype. You might need to change the function definition to capture the rest of the message and modify the logic to handle the cowtype if none is passed.

brad90four (changes requested):
In the current form this does not function the way I think it should. Two things that I see are:
1. Default cowsay mode cannot be called - suggestion to fix that worked during testing is to start the `SUPPORTED_COWS` list with "default".
2. Message content cannot be passed without defining a cowtype. If the user tries to do `.cowsay <message>` the first word of the message is interpreted as the cowtype, and raises the BadArgument error. I think this can be handled by changing the function definition to:
```py
async def cowsay(self, ctx: commands.Context, cowtype: str = "cow", *, message: str = None) -> None:
```
And then we need to make a decision to try to pass the message to the default cow, or try to determine if the user was attempting to pass an unknown cowtype instead.

bot/exts/fun/cowsay.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @commands.command(
+        help=f"""
+        Generates a cowsay string and sends it. The available characters are:
+
+        {', '.join(SUPPORTED_COWS)}"""
+    )
+    async def cowsay(self, ctx: commands.Context, cowtype: str = "cow", message: str = None) -> None:
```
Bluenix2, line 26: Use `Optional[Literal[*SUPPORTEDCOWS]]`, this means that if the user doesn't pass a supported cow it will become of the message following it.

bot/exts/fun/cowsay.py
```diff
@@ -0,0 +1,56 @@
+from cowpy import cow
+from discord.ext import commands
+
+from bot.bot import Bot
+
+SUPPORTED_COWS = []
+
+for element in cow.get_cowacters():
+    # The cows in the list are too big for Discord or innapropriate
+    if element[0] not in ["turtle", "stegosaurus", "ghostbusters", "turkey", "mutilated", "kiss"]:
```
Bluenix2, line 10:
```suggestion
    if element[0] not in {"turtle", "stegosaurus", "ghostbusters", "turkey", "mutilated", "kiss"}:
```

bot/exts/fun/cowsay.py
```diff
@@ -0,0 +1,56 @@
...
+    async def cowsay(self, ctx: commands.Context, cowtype: str = "cow", message: str = None) -> None:
+        """Builds a cowsay string and sends it to discord."""
+        if message is None:
+            message = f"I'm a {cowtype}"
+
+        if cowtype == "cow":
+            cowtype = "default"
+
+        if cowtype not in SUPPORTED_COWS:
+            raise commands.BadArgument("The cow submitted is invalid! Please use a valid cow.")
```
Bluenix2, line 35: This would be handled by the Literal.

bot/exts/fun/cowsay.py
```diff
@@ -0,0 +1,56 @@
...
+        rawcow = cow.get_cow(cowtype)
+        message = message.replace("`", "`\u200b")
```
Bluenix2, line 37: What character is that?

Xithrius:
@Kronifer Greetings. Will you be able to finish off this PR with the file conflicts and reviews?

Kronifer:
@Xithrius I will not be continuing this, cowpy has an incompatible license

Xithrius:
@Kronifer should we close this PR then?

Xithrius:
Very well. I hope we're able to implement this in the future.

PR #910, opened 2021-10-14
Trivia night event

### LINKED ISSUE #903: Python Trivia Night Command(s)
## Description
One of the events I'm planning to host in 2022 is a Python Trivia Night! People will gather in a stage channel and we'll be asking Python Trivia questions for people to answer. To help with participation and running the event, I'd like to have a cog/extension with a set of commands to help facilitate this.

## Proposed Implementation
I'd like a new set of commands under `.trivianight` or `.triviaevent`.

`.trivianight load` will accept a json file of appropriately formatted questions, options for users to select, and the final answer.
`.trivianight start` will start the trivia event. For this, it will initialize a scoreboard and clear any old scoreboards that may be hanging around.
`.trivianight next` will pick an unanswered question from the question bank, post it in a configured channel, and provide multiple choice buttons for users to select. Users will be able to press the buttons to select their answer and should be allowed to change their answer once. Any subsequent changes aren't allowed and the interaction should fail. After a certain period of time or with a command (`.trivianight pause` or `.trivianight next`) the buttons will be removed and the correct answer will be revealed.
`.trivianight stop` will end the trivia event and produce the final scoreboard results. I'd like to keep track of questions attempted, questions answered correctly, and average time to answer a question. 
`.trivianight #` will select that specific question whether or not it's been answered to be answered again.

## Additional Details
- This should use Discord buttons!
- If an interaction fails, it should be an ephemeral message to just the user.
- Bonus points if there's some validation on the provided question json, but not necessary.

I'm fine with anyone implementing this, but I'd like to get this fully functional by mid-December and it will most likely require a fair amount of iteration and changes. So keep that in mind if you'd like to work on this.

Shom770:
## Relevant Issues

Closes #903

## Description

This is for the trivia night event, and will be the implementation of it into sir lancebot.
There are a few main commands:
`.trivianight load (file)`
This will load the JSON file attached as the questions for the cog to handle for when the event starts.

`.trivianight start`
This will begin the event (this could possibly send an embed regarding the start of the event)

`.trivianight next`
This will head to the next question, displaying the question in a configured channel and letting users pick a choice. They can pick again if they wish to do so but after that, an ephemeral will be sent saying that they cannot pick again after the second time.

`.trivianight stop`
This will display the correct answer and award points accordingly

`.trivianight question #`
This will head to the question specified, displaying the question in a configured channel and letting users pick a choice. They can pick again if they wish to do so but after that, an ephemeral will be sent saying that they cannot pick again after the second time.

`.trivianight list`
This will display all the questions and which ones have been displayed so far

`.trivianight end`
This will stop the event and display the leaderboard.

## Did you:

### DIFF
--- a/bot/exts/events/trivianight/__init__.py
+++ b/bot/exts/events/trivianight/__init__.py
@@ -1,0 +1,7 @@
+class UserScore:
+    """Marker class for passing into the scoreboard to add points/record speed."""
+
+    __slots__ = ("user_id",)
+
+    def __init__(self, user_id: int):
+        self.user_id = user_id
--- a/bot/exts/events/trivianight/__init__.py
+++ b/bot/exts/events/trivianight/__init__.py
@@ -1,7 +1,0 @@
-class UserScore:
-    """Marker class for passing into the scoreboard to add points/record speed."""
-
-    __slots__ = ("user_id",)
-
-    def __init__(self, user_id: int):
-        self.user_id = user_id
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -33,6 +33,10 @@ class AlreadyUpdated(RuntimeError):
     """Exception raised when the user has already updated their guess once."""
 
 
+class AllQuestionsVisited(RuntimeError):
+    """Exception raised when all of the questions have been visited."""
+
+
 class Question:
     """Interface for one question in a trivia night game."""
 
@@ -142,6 +146,8 @@ class TriviaNightGame:
                 question = [q for q in self._all_questions if q.number == number][0]
             except IndexError:
                 raise ValueError(f"Question number {number} does not exist.")
+        elif len(self._questions) == 0:
+            raise AllQuestionsVisited("All of the questions have been visited.")
         else:
             question = self._questions.pop(randrange(len(self._questions)))
 
@@ -161,7 +167,7 @@ class TriviaNightGame:
         self.current_question.stop()
         self.current_question = None
 
-    def list_questions(self) -> None:
+    def list_questions(self) -> str:
         """
         List all the questions.
 
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -144,7 +144,7 @@ class TriviaNightGame:
 
         if number is not None:
             try:
-                question = [q for q in self._all_questions if q.number == number][0]
+                question = [q for q in self._all_questions if q.number == int(number)][0]
             except IndexError:
                 raise ValueError(f"Question number {number} does not exist.")
         elif len(self._questions) == 0:
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -1,0 +1,105 @@
+import time
+from string import ascii_uppercase
+from typing import Iterable, Optional, TypedDict
+
+
+DEFAULT_QUESTION_POINTS = 10
+DEFAULT_QUESTION_TIME = 10
+
+
+class QuestionData(TypedDict):
+    """Representing the different 'keys' of the question taken from the JSON."""
+
+    number: str
+    description: str
+    answers: list[str]
+    correct: str
+    points: Optional[int]
+    time: Optional[int]
+
+
+UserGuess = tuple[
+    str,  # The answer that was guessed
+    bool,  # Whether the answer can be changed again
+    float  # The time it took to guess
+]
+
+
+class Question:
+    """Interface for one question in a trivia night game."""
+
+    def __init__(self, data: QuestionData):
+        self._data = data
+        self._guesses: dict[int, UserGuess] = {}
+        self._started = None
+
+    # These properties are mostly proxies to the underlying data:
+
+    @property
+    def number(self) -> str:
+        """The number of the question."""
+        return self._data["number"]
+
+    @property
+    def description(self) -> str:
+        """The description of the question."""
+        return self._data["description"]
+
+    @property
+    def answers(self) -> list[tuple[str, str]]:
+        """The possible answers for this answer.
+
+        This is a property that returns a list of letter, answer pairs.
+        """
+        return [(ascii_uppercase[i], q) for (i, q) in enumerate(self._data["answers"])]
+
+    @property
+    def correct(self) -> str:
+        """The correct answer for this question."""
+        return self._data["correct"]
+
+    @property
+    def max_points(self) -> int:
+        """The maximum points that can be awarded for this question."""
+        return self._data.get("points") or DEFAULT_QUESTION_POINTS
+
+    @property
+    def time(self) -> float:
+        """The time allowed to answer the question."""
+        return self._data.get("time") or DEFAULT_QUESTION_TIME
+
+    def start(self) -> float:
+        """Start the question and return the time it started."""
+        self._started = time.perf_counter()
+        return self._started
+
+    def _update_guess(self, user: int, answer: str) -> UserGuess:
+        """Update an already existing guess."""
+        if self._started is None:
+            raise RuntimeError("Question is not open for answers.")
+
+        if self._guesses[user][1] is False:
+            raise RuntimeError(f"User({user}) has already updated their guess once.")
+
+        self._guesses[user] = (answer, False, time.perf_counter() - self._started)
+        return self._guesses[user]
+
+    def guess(self, user: int, answer: str) -> UserGuess:
+        """Add a guess made by a user to the current question."""
+        if user in self._guesses:
+            return self._update_guess(user, answer)
+
+        if self._started is None:
+            raise RuntimeError("Question is not open for answers.")
+
+        self._guesses[user] = (answer, True, time.perf_counter() - self._started)
+        return self._guesses[user]
+
+    def stop(self) -> dict[int, UserGuess]:
+        """Stop the question and return the guesses that were made."""
+        guesses = self._guesses
+
+        self._started = None
+        self._guesses = {}
+
+        return guesses
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -18,11 +18,10 @@ class QuestionData(TypedDict):
     time: Optional[int]
 
 
-UserGuess = tuple[
-    str,  # The answer that was guessed
-    bool,  # Whether the answer can be changed again
-    float  # The time it took to guess
-]
+class UserGuess(NamedTuple):
+    answer: str
+    editable: bool
+    elapsed: float
 
 
 class QuestionClosed(RuntimeError):
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -3,7 +3,6 @@ from random import randrange
 from string import ascii_uppercase
 from typing import Iterable, Optional, TypedDict
 
-
 DEFAULT_QUESTION_POINTS = 10
 DEFAULT_QUESTION_TIME = 10
 
@@ -56,7 +55,8 @@ class Question:
 
     @property
     def answers(self) -> list[tuple[str, str]]:
-        """The possible answers for this answer.
+        """
+        The possible answers for this answer.
 
         This is a property that returns a list of letter, answer pairs.
         """
@@ -119,7 +119,11 @@ class TriviaNightGame:
 
     def __init__(self, data: list[QuestionData]) -> None:
         self._questions = [Question(q) for q in data]
+        # A copy of the questions to keep for `.trivianight list`
+        self._all_questions = list(self._questions)
         self.current_question: Optional[Question] = None
+        self._points = {}
+        self._speed = {}
 
     def __iter__(self) -> Iterable[Question]:
         return iter(self._questions)
@@ -135,7 +139,7 @@ class TriviaNightGame:
 
         if number is not None:
             try:
-                question = [q for q in self._questions if q.number == number][0]
+                question = [q for q in self._all_questions if q.number == number][0]
             except IndexError:
                 raise ValueError(f"Question number {number} does not exist.")
         else:
@@ -156,3 +160,24 @@ class TriviaNightGame:
 
         self.current_question.stop()
         self.current_question = None
+
+    def list_questions(self) -> None:
+        """
+        List all the questions.
+
+        This method should be called when `.trivianight list` is called to display the following information:
+            - Question number
+            - Question description
+            - Visited/not visited
+        """
+        formatted_string = ""
+
+        spaces = max(len(q.description) for q in self._all_questions)
+
+        for question in self._all_questions:
+            visited, not_visited = ":checkmark:", ":x:"
+            formatted_string += f"`Q{question.number}: {question.description}" \
+                                f"{' ' * (spaces - len(question.description))}|`" \
+                                f" {visited if question not in self._all_questions else not_visited}\n"
+
+        return formatted_string
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -178,6 +178,6 @@ class TriviaNightGame:
             visited, not_visited = ":checkmark:", ":x:"
             formatted_string += f"`Q{question.number}: {question.description}" \
                                 f"{' ' * (spaces - len(question.description))}|`" \
-                                f" {visited if question not in self._all_questions else not_visited}\n"
+                                f" {visited if question not in self._questions else not_visited}\n"
 
         return formatted_string
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -18,10 +18,11 @@ class QuestionData(TypedDict):
     time: Optional[int]
 
 
-class UserGuess(NamedTuple):
-    answer: str
-    editable: bool
-    elapsed: float
+UserGuess = tuple[
+    str,  # Represents the answer choice the user chose.
+    bool,  # Represents if the user can edit their answer.
+    float  # Represents the amount of time passed since the question began.
+]
 
 
 class QuestionClosed(RuntimeError):
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -177,14 +177,22 @@ class TriviaNightGame:
             - Question description
             - Visited/not visited
         """
+        question_list = []
         formatted_string = ""
 
-        spaces = max(len(q.description) for q in self._all_questions)
+        visited = ":white_check_mark:"
+        not_visited = ":x:"
 
         for question in self._all_questions:
-            visited, not_visited = ":white_check_mark:", ":x:"
-            formatted_string += f"`Q{question.number}: {question.description}" \
-                                f"{' ' * (spaces - len(question.description))}|`" \
-                                f" {visited if question not in self._questions else not_visited}\n"
+            formatted_string += (
+                f"**Q{question.number}** {not_visited if question in self._questions else visited}"
+                f"\n{question.description}\n\n"
+            )
+            if question.number % 5 == 0:
+                question_list.append(formatted_string.rstrip())
+                formatted_string = ""
 
-        return formatted_string
+        if formatted_string:
+            question_list.append(formatted_string)
+
+        return question_list
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -142,7 +142,6 @@ class TriviaNightGame:
             question = self._questions.pop(randrange(len(self._questions)))
 
         self.current_question = question
-        self.current_question.start()
         return question
 
     def end_question(self) -> None:
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -178,21 +178,15 @@ class TriviaNightGame:
             - Visited/not visited
         """
         question_list = []
-        formatted_string = ""
 
         visited = ":white_check_mark:"
         not_visited = ":x:"
 
         for question in self._all_questions:
-            formatted_string += (
+            formatted_string = (
                 f"**Q{question.number}** {not_visited if question in self._questions else visited}"
                 f"\n{question.description}\n\n"
             )
-            if question.number % 5 == 0:
-                question_list.append(formatted_string.rstrip())
-                formatted_string = ""
-
-        if formatted_string:
-            question_list.append(formatted_string)
+            question_list.append(formatted_string.rstrip())
 
         return question_list
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -4,7 +4,7 @@ from string import ascii_uppercase
 from typing import Iterable, Optional, TypedDict
 
 DEFAULT_QUESTION_POINTS = 10
-DEFAULT_QUESTION_TIME = 10
+DEFAULT_QUESTION_TIME = 20
 
 
 class QuestionData(TypedDict):
--- a/bot/exts/events/trivianight/_game.py
+++ b/bot/exts/events/trivianight/_game.py
@@ -1,4 +1,5 @@
 import time
+from random import randrange
 from string import ascii_uppercase
 from typing import Iterable, Optional, TypedDict
 
@@ -23,6 +24,14 @@ UserGuess = tuple[
     bool,  # Whether the answer can be changed again
     float  # The time it took to guess
 ]
+
+
+class QuestionClosed(RuntimeError):
+    """Exception raised when the question is not open for guesses anymore."""
+
+
+class AlreadyUpdated(RuntimeError):
+    """Exception raised when the user has already updated their guess once."""
 
 
 class Question:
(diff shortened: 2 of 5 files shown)


bot/exts/events/trivianight/_questions.py
```diff
@@ -0,0 +1,210 @@
...
+
+        It will put the following into a message:
+            - Question number
+            - Question description
+            - If the question was already 'visited' (displayed)
+        """
+        spaces = len(sorted(self.questions, key=lambda question: len(question['description']))[-1]["description"]) + 3
+        formatted_string = ""
+        for question in self.questions:
+            formatted_string += f"`Q{question['number']}: {question['description']!r}" \
```
janine9vn, line 194: This should not be a repr, since this is being displayed to the end user. It outputs something fairly unreadable.
image

janine9vn (changes requested):
So! I will likely not comment on specific pieces of code because the changes I'm going to request will span a lot and likely require a fundamental change to some of the logic.

Also thank you in advance for working on this! Please don't be overwhelmed by my requested changes.

# New Features
After playing with this for a bit and thinking on the event overall, I am going to ask for some changes that deviate from the original issue.
- Adding a timer to questions. Each question will be on a timer and once the time runs out, people can no longer answer the question. In my ideal world, there will be a default timing assumed, than can be overridden per question via json. But I'm open to other options. (`.trivianight stop` should probably still exist just in case something goes wonky with the timer though.)
- Scoring! see below
### Scoring
I've had a think about it and I've decided to go with a modified kahoot scoring. Each question will be worth X points, we can probably assume a default of 10, with an override being allowed if present in the json. The scoring will operate as follows (they only get points if its correct): 
1. Get what time the player responded since the timer responded and divide that by the amount of time you have for the question. (i.e. 20 second timer, player responded 2 seconds in, 2/20 = 0.1 is the result)
2. Divide that number by 2
3. Subtract from 1 (our number is now 0.95)
4. Multiple by the total value of points left (if points are worth 10 points, this player now gets 9.5 points)

**However!**, the modification I'm going to request is that if a person responds incorrectly within the first 2 seconds of the question, they get that number of points they would have gotten if it were correct, subtracted from their score (Happy to answer with my reasoning in dev-contrib). If someone responds incorrectly 3 seconds after the timer starts or later, there's no penalty.

# The Bugs
So I hit quite a few bugs and things that I think should be handled a bit more gracefully. I'm going to describe how I encountered them and what I think the actual behavior should be.
- `.trivianight list` with no question list loaded in just errors out. This should probably just tell me that no questions are loaded.
- `.trivianight list` with a question list loaded in provides an... odd result. 
image
- the formatting for `.trivianight list` can probably be improved a bit, figuring out the best way to make it readable.
- .`trivianight list` uses a :checkmark: emoji. That emoji does not exist in our main or emoji servers or in the standard unicode emoji list.
- `.trivianight next` twice in a row produces unexpected behavior and kind of locks up the interaction buttons. Should probably be some sort of check to see if a question is active/unresolved, and if it is either throw a warning *or* just end it and start the next one. 
- The scoreboard and what's my rank is broken. It doesn't show me as participating even if I participate.
- `.trivianight` with no args currently outputs nothing. I think it would be nice to show the status of if a game is playing or night or maybe a quick description of what the cog is for.
- `.trivianight question #` should force a question to appear, regardless if it's been used or not. Currently if it's been used, it'll be stopped from showing.

# Other thoughts
- More documentation within the functions and cog itself would go a long way.
- Also, specific documentation on the format of the json it expects should be in there somewhere. Currently you have to look at the code and just kind of hope you parsed it right.

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,202 @@
...
+        """Takes `text` and adds zero-width spaces to prevent copy and pasting the question."""
+        return "".join(
+            f"{letter}\u200b" if letter not in ('\n', '\t', '`', 'p', 'y') else letter
+            for idx, letter in enumerate(text)
+        )
+
+    @commands.group(aliases=["tn"], invoke_without_command=True)
+    async def trivianight(self, ctx: commands.Context) -> None:
+        """No-op subcommand group for organizing different commands."""
+        if ctx.invoked_subcommand is None:
```
Bluenix2, line 35: This is what the `invoke_without_command` kwarg does.

bot/exts/events/trivianight/_scoreboard.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class Scoreboard:
+    """Class for the scoreboard for the Trivia Night event."""
+
+    def __setitem__(self, key: str, value: int):
+        if value.get("points") and key.user_id not in self.view.points.keys():
+            self.view.points[key.user_id] = value["points"]
+        elif value.get("points"):
+            self.view.points[key.user_id] += self.view.points[key.user_id]
```
Bluenix2, line 117:
```suggestion
        if value.get("points"):
            self.view.points[key.user_id] += self.view.points.get(key.user_id, value["points"])
```
If you accept this repeat the same for speed.

bot/exts/events/trivianight/_scoreboard.py
```diff
@@ -0,0 +1,128 @@
...
+    @discord.ui.button(label="What's my rank?", style=ButtonStyle.blurple)
+    async def rank_button(self, button: Button, interaction: Interaction) -> None:
+        """Send an ephemeral message with the user's rank for the overall points/average speed."""
+        await interaction.response.send_message(embed=self._get_rank(interaction.user), ephemeral=True)
+
+
+class Scoreboard:
+    """Class for the scoreboard for the Trivia Night event."""
+
+    def __setitem__(self, key: str, value: int):
```
Bluenix2, line 113: Is this type annotation still correct?

bot/exts/events/trivianight/_scoreboard.py
```diff
@@ -0,0 +1,128 @@
...
+            self.view.points[key.user_id] += self.view.points[key.user_id]
+
+        if value.get("speed") and key.user_id not in self.view.speed.keys():
+            self.view.speed[key.user_id] = [1, value["speed"]]
+        elif value.get("speed"):
+            self.view.speed[key.user_id] = [
+                self.view.speed[key.user_id][0] + 1, self.view.speed[key.user_id][1] + value["speed"]
+            ]
+
+    async def display(self) -> Union[Embed, View]:
```
Bluenix2, line 126: This type annotation isn't correct either.

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,202 @@
...
+        """Resets previous questions and scoreboards."""
+        self.scoreboard.view = ScoreboardView(self.bot)
+        all_questions = self.questions.questions
+        self.questions.view = QuestionView()
+
+        for question in all_questions:
+            if "visited" in question.keys():
+                del question["visited"]
+
+        self.questions.questions = list(all_questions)
```
Bluenix2, line 104: Isn't there a `set_questions()` method?

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,202 @@
...
+        success_embed = Embed(
+            title=choice(POSITIVE_REPLIES),
+            description="The JSON was loaded successfully!",
+            color=Colours.soft_green
+        )
+        await ctx.send(embed=success_embed)
+
+    @trivianight.command()
+    async def reset(self, ctx: commands.Context) -> None:
+        """Resets previous questions and scoreboards."""
```
Bluenix2, line 95: This is meant to be used after the game right?

bot/exts/events/trivianight/_questions.py
```diff
@@ -0,0 +1,256 @@
...
+
+        if number is None:
+            question_number = randrange(0, len(self.questions))
+            while "visited" in self.questions[question_number].keys():
+                question_number = randrange(0, len(self.questions))
+        else:
+            question_number = number - 1
+
+        self.questions[question_number]["visited"] = True
+        self.view.current_question = self.questions[question_number]
```
Bluenix2, line 209: What view?

bot/exts/events/trivianight/_questions.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def current_question(self) -> tuple[Embed, QuestionView]:
+        """Returns an embed entailing the current question as an embed with a view."""
+        return self.view.create_current_question(), self.view
+
+    def end_question(self) -> Embed:
+        """Terminates answering of the question and displays the correct answer."""
+        scores, answer_embed, time_limit, total_points = self.view.end_question()
+        for user, score in scores.items():
+            # Overhead with calculating scores leads to inflated times, subtracts 0.5 to give an accurate depiction
```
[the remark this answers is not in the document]
Shom770: It takes 0.5 seconds usually from the interaction handling on Discord's end.

bot/exts/events/trivianight/trivianight.py
Shom770: Ah, I didn't know that so I used both. Thanks!

bot/exts/events/trivianight/_scoreboard.py
Shom770: Wouldn't this error out if `key.user_id` isn't a key inside `self.view.points`?

bot/exts/events/trivianight/_questions.py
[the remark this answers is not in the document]
Bluenix2: So if someone presses a button the instant Discord receives it they will still have taken 0.5 seconds according to us?

Isn't that just ping? I don't think we should be on relying on this - we aren't measuring exactly how fast people are to answer so that we can see who's the fastest, we just need to be able to give someone who clicked after 5 seconds more points than someone who clicked after 20.

bot/exts/events/trivianight/_scoreboard.py
Bluenix2: That's what the `.get()` is, the second argument is the default when that key isn't already present.

Bluenix2: Oh nvm - I just realized - that does `+=` vs. `=` for when it is missing.

Yeah you're right here.

bot/exts/events/trivianight/_questions.py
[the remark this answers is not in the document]
Shom770: Good point, I'll remove the overhead subtracting

bot/exts/events/trivianight/trivianight.py
Bluenix2: I see your thumbs up, one thing I like to do in my code is to have the reaet logic be shared with the initializing code.

This way, you dont forget to reset anything 😉. So my recommendation would be to have like a `setup()` or `initialize()` method that is called by both the `reset` command and `__init__()`.

Shom770: Gotcha! I'll add this after updating the command checks

bot/exts/events/trivianight/_questions.py
```diff
@@ -0,0 +1,240 @@
...
+class Questions:
+    """An interface to use from the TriviaNight cog for questions."""
+
+    def __init__(self, scoreboard: Scoreboard):
+        self.scoreboard = scoreboard
+        self.questions = []
+
+    def set_questions(self, questions: list) -> None:
+        """Setting `self.questions` dynamically via a function to set it."""
+        self.questions = questions
```
Bluenix2, line 169: Maybe we can just make a copy of the list (using `[:]`) and then pop/remove items from it rather than modifying it with the `visited` key?

Is there a reason you went for setting that key rather than something like this?

bot/exts/events/trivianight/_questions.py
```diff
@@ -0,0 +1,240 @@
...
+            question_description = question["description"]
+            formatted_string += f"`Q{question['number']}: {question_description}" \
+                                f"{' ' * (spaces - len(question_description) + 2)}" \
+                                f"|` {':x:' if not question.get('visited') else ':white_check_mark:'}\n"
+
+        return formatted_string.strip()
+
+    def current_question(self) -> tuple[Embed, QuestionView]:
+        """Returns an embed entailing the current question as an embed with a view."""
+        return self.view.create_current_question(), self.view
```
Bluenix2, line 227: Could technically be a property no?

bot/exts/events/trivianight/_questions.py
```diff
@@ -0,0 +1,240 @@
...
+        """Terminates answering of the question and displays the correct answer."""
+        scores, answer_embed, time_limit, total_points = self.view.end_question()
+        for user, score in scores.items():
+            time_taken = score[2]
+            point_calculation = (1 - (time_taken / time_limit) / 2) * total_points
+            if score[-1] is True:
+                self.scoreboard[UserScore(user)] = {"points": point_calculation, "speed": time_taken}
+            elif score[-1] is False and score[2] <= 2:
+                # Get the negative of the point_calculation to deduct it
+                self.scoreboard[UserScore(user)] = {"points": -point_calculation}
```
Bluenix2, line 239: Wait now why exactly so we remove as much as the user would've won if it wasn't for *whatever the if-statement tests*?

bot/exts/events/trivianight/_questions.py
Shom770: Hmm, yeah it could be. But, I guess if we're gonna make this a property, we'll have to make other things properties too right?

Shom770: i'm down with changing this to a property though

bot/exts/events/trivianight/_questions.py
Shom770: Since if the user responds in less than 2 seconds and gets it wrong, they would've lost the points they were going to gain.

I should honestly make the score into a dictionary so it's clearer or document each part of score.

bot/exts/events/trivianight/_questions.py
Bluenix2: Right, but instead of filtering that if we instead just make a copy as I suggested you can just iterate through all questions.

bot/exts/events/trivianight/_questions.py
Bluenix2: > we'll have to make other things properties too right?

Not have to, it's really not a big deal, just thought I'd bring it up in-case it didn't cross your mind yet.

Shom770: Ah alright. I think we could keep it as is considering most things are functions anyways

bot/exts/events/trivianight/_questions.py
Shom770: That sounds fine, but how would we check if the question has been visited or not? Just compare it to a main list?

That main list would still be mutated though and it would be extra work to switch it though. Not sure if I'm following your suggestion tho

Shom770: Ohh, I see what you're saying. We have a list of all the questions which is just a dictionary and then pop the dictionaries out if they were already used?

Bluenix2: Yes - make a copy of the list from the deserialized JSON (otherwise you would modify the original list) and remove questions as they are picked (for example: `my_list.pop(random.randrange(len(my_list)))`.

Shivansh-007 (commented):
Besides the requested changes, great PR.

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        if self.game is None:
+            await ctx.send(embed=Embed(
+                title=choice(NEGATIVE_REPLIES),
+                description="There is no trivia night running!",
+                color=Colours.soft_red
+            ))
+            return
+
+        if self.game.current_question is not None:
```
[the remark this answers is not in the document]
Bluenix2: That's just more code and another view - the user should understand from this that they need to call `.trivianight stop` or let the question run out.

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,266 @@
...
+                await message.edit(embed=question_embed, view=None)
+
+                self.game.end_question()
+                self.question_closed.clear()
+                return
+
+            if int(duration) > 1:
+                # It is quite ugly to display decimals, the delay for requests to reach Discord
+                # cause sub-second accuracy to be quite pointless.
+                await ctx.send(f"{int(duration)}s remaining...")
```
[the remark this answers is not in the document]
Bluenix2: I don't think they're important enough. Right now the code yields the following rather satisfactory result for a 20s duration:
```
10s remaining...
5s remaining...
2s remaining...
```
There is no need to differentiate between 2.5s and 2s.

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,266 @@
...
+        """
+        if self.game is None:
+            await ctx.send(embed=Embed(
+                title=choice(NEGATIVE_REPLIES),
+                description="There is no trivia night running!",
+                color=Colours.soft_red
+            ))
+            return
+
+        if self.game.current_question is not None:
```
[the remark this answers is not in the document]
Bluenix2: Again, see my reply on [your comment above](https://github.com/python-discord/sir-lancebot/pull/910/files#r773153776). We need this to be merged relatively soon.

bot/exts/events/trivianight/__init__.py
```diff
@@ -0,0 +1,7 @@
+class UserScore:
```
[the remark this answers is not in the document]
Bluenix2: There is no longer a need for this class, so it does not matter. @Shom770 should remove it - we can just pass the user ID directly.

bot/exts/events/trivianight/_game.py
```diff
@@ -0,0 +1,183 @@
...
+        return self._data.get("time") or DEFAULT_QUESTION_TIME
+
+    def start(self) -> float:
+        """Start the question and return the time it started."""
+        self._started = time.perf_counter()
+        return self._started
+
+    def _update_guess(self, user: int, answer: str) -> UserGuess:
+        """Update an already existing guess."""
+        if self._started is None:
```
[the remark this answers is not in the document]
Bluenix2: Yes, it could be a bool instead.

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,251 @@
...
+        - correct: str (represents the correct answer in terms of what the correct answer is in `answers`
+        - time: Optional[int] (represents the timer for the question and how long it should run, default is 10)
+        - points: Optional[int] (represents how many points are awarded for each question, default is 10)
+
+        The load command accepts three different ways of loading in a JSON:
+        - an attachment of the JSON file
+        - a message link to the attachment/JSON
+        - reading the JSON itself via a codeblock or plain text
+        """
+        if self.game is not None:
```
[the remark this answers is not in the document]
janine9vn: That isn't something within the original spec and I don't think it's necessary. Sourcing and finalizing the questions takes quite a bit of work, so I don't imagine I'll do that on the fly or want to.

bot/exts/events/trivianight/_game.py
```diff
@@ -0,0 +1,183 @@
...
+            raise QuestionClosed("Question is not open for answers.")
```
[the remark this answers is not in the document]
Bluenix2: This does not have to necessarily be the case if I remember correctly, I can see `currently` being added to the sentence though.

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class TriviaNightCog(commands.Cog):
+    """Cog for the Python Trivia Night event."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+        self.game: Optional[TriviaNightGame] = None
+        self.scoreboard: Optional[Scoreboard] = None
+        self.question_closed: asyncio.Event = None
```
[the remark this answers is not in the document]
Bluenix2: I believe the original plan was to have it be None when there is no active question. It could just be a re-used event though you're right.

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,266 @@
...
+        if self.game.current_question is None:
+            error_embed = Embed(
+                title=choice(NEGATIVE_REPLIES),
+                description="There is no ongoing question!",
+                color=Colours.soft_red
+            )
+            await ctx.send(embed=error_embed)
+            return
+
+        self.question_closed.set()
```
Bluenix2, line 228: This could probably need a confirmation message.

bot/exts/events/trivianight/_game.py
```diff
@@ -0,0 +1,183 @@
...
+
+        self._guesses[user] = (answer, True, time.perf_counter() - self._started)
+        return self._guesses[user]
+
+    def stop(self) -> dict[int, UserGuess]:
+        """Stop the question and return the guesses that were made."""
+        guesses = self._guesses
+
+        self._started = None
+        self._guesses = {}
```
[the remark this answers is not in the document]
Shivansh-007: By none, I meant `{}`.

jchristgit (commented):
This is a pretty pretty cool feature, thank you for the extensive PR description documenting it!

I've left a few comments for some smaller nitpicks of mine which would be great to address. If those along with the comments from the remaining outstanding review comments are looked through, I would be happy to give this the green approve 🙂

bot/exts/events/trivianight/_game.py
```diff
@@ -18,10 +18,11 @@ class QuestionData(TypedDict):
...
 
-class UserGuess(NamedTuple):
-    answer: str
-    editable: bool
-    elapsed: float
+UserGuess = tuple[
+    str,  # Represents the answer choice the user chose.
+    bool,  # Represents if the user can edit their answer.
+    float  # Represents the amount of time passed since the question began.
+]
```
Bluenix2, line 25: Oh why was this reverted?

Shom770:
@Bluenix2 It was reverted since I don't think there is any `NamedTuple` class per se that you can inherit from, just the `namedtuple` function from collections. Unless there's an underlying class it uses.

Bluenix2:
> @Bluenix2 It was reverted since I don't think there is any `NamedTuple` class per se that you can inherit from, just the `namedtuple` function from collections. Unless there's an underlying class it uses.

There, it's under the `typing` module.

Shom770:
> > @Bluenix2 It was reverted since I don't think there is any `NamedTuple` class per se that you can inherit from, just the `namedtuple` function from collections. Unless there's an underlying class it uses.
> 
> There is, it's under the `typing` module.

Ohhh. 🤦‍♂️ I should've checked

bot/exts/events/trivianight/_scoreboard.py
```diff
@@ -0,0 +1,186 @@
...
+            formatted_string += f"({(time_taken[-1] / time_taken[0]):.1f}s)\n"
+            if (current_placement + 1) % 10 == 0:
+                formatted_string += "⎯⎯⎯⎯⎯⎯⎯⎯\n"
+
+        speed_embed = Embed(
+            title="Average time taken to answer a question",
+            description=formatted_string,
+            color=Colours.python_blue
+        )
+        return speed_embed
```
Bluenix2, line 85: I don't remember, does this account for people not answering all questions? Also we should probably filter out incorrect answers (if possible?)

bot/exts/events/trivianight/trivianight.py
```diff
@@ -0,0 +1,328 @@
...
+    @commands.has_any_role(*TRIVIA_NIGHT_ROLES)
+    async def end_game(self, ctx: commands.Context) -> None:
+        """Ends the ongoing game."""
+        self.game = None
+
+        await ctx.send(embed=Embed(
+            title=choice(POSITIVE_REPLIES),
+            description="The game has been stopped.",
+            color=Colours.soft_green
+        ))
```
Bluenix2, line 323: Was this accidentally left behind? What's the differnence between this and the end method above?

Bluenix2 (commented):
I couldn't find much here @Shom770 but there's a few points to be brought up.

bot/exts/events/trivianight/_game.py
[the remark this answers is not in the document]
Bluenix2: I agree with this, for the sake of keeping the data as long as possible @Shom770. Move this to `start()` (`self._guesses = {}` that is). So that guesses are left intact for as long as possible.

bot/exts/events/trivianight/trivianight.py
janine9vn: It's a force end in case something wonky happens with the interactions or something weird goes on.
It'll let us force end it and then I can call for the scoreboard specifically which will embed.

bot/exts/events/trivianight/_game.py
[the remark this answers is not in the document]
janine9vn: I don't see a real benefit to keeping the guesses around until the next question starts. Trying to int-e our way back to viewing the guesses later is not something I'm going to do mid-event.

janine9vn (approved):
Great feature and thank you so much for your work on it! The core of this is very, very solid and it runs like a charm.

I think the remaining unaddressed feedback can be added into a QoL PR later as it doesn't impact the core functionality.

mbaruh (approved):

ChrisLovering:
Approving since it's tested and works. We don't want to delay further due to the event happening this evening.

However, we'd like to see some of the comments here actioned in a subsequent PR.

Thanks for the input everyone, nice work 👍

PR #936, opened 2021-11-02
Limit user reactions on embed pagination

### LINKED ISSUE #933: possible error/bug at `.wiki` command
## Description

There might be a bug/error when reacting to `.wiki` commands trashcan, I was able to remove the message by clicking trashcan even when I was not the one running the command. 

## Steps to Reproduce

If this is not how it should act there might be a error in check. Maybe adding a check to function to `.wiki` command (if this is not how it should act). It would be nice to be limited, so it would not be clicked to annoy others or be pressed mistakenly to remove the message.

## Expected Behaviour

I would expect the bot to remove the message only if command author presses the trashcan reaction. 

## Actual Behaviour

Instead anyone seems to be able to remove the message when trashcan is pressed.

## Known Impacted Platforms

- [ ] Web
- [x] Desktop
- [ ] Android App
- [ ] iOS App

## Possible Solutions

Checking if there is a check function to see who click the trashcan and see if they are the one who ran the command. So adding the check in there.

Nipa-Code:
## Relevant Issues

Close #933 

## Description

I added `restrict_to_user=ctx.author` to wikipedia.py. Limits user reactions to prevent non-author from removing message by adding user restriction to paginator.

## Did you:

### DIFF
--- a/bot/exts/utilities/wikipedia.py
+++ b/bot/exts/utilities/wikipedia.py
@@ -86,9 +86,7 @@ class WikipediaSearch(commands.Cog):
             )
             embed.set_thumbnail(url=WIKI_THUMBNAIL)
             embed.timestamp = datetime.utcnow()
-            await LinePaginator.paginate(
-                contents, ctx, embed, restrict_to_user=ctx.author
-            )
+            await LinePaginator.paginate(contents, ctx, embed, restrict_to_user=ctx.author)
         else:
             await ctx.send(
                 "Sorry, we could not find a wikipedia article using that search term."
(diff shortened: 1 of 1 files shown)


bot/exts/utilities/wikipedia.py
```diff
@@ -87,7 +87,7 @@ async def wikipedia_search_command(self, ctx: commands.Context, *, search: str)
             embed.set_thumbnail(url=WIKI_THUMBNAIL)
             embed.timestamp = datetime.utcnow()
             await LinePaginator.paginate(
-                contents, ctx, embed
+                contents, ctx, embed, restrict_to_user=ctx.author
```
ChrisLovering, line 90: This works well, but doesn't need to be split up like this. I'm aware you didn't do this yourself, but lets make it better while we're here. The style we currently go with is this

`await LinePaginator.paginate(contents, ctx, embed, restrict_to_user=ctx.author)`

ChrisLovering (changes requested):

hedyhli (approved):

ChrisLovering (approved):

PR #941, opened 2021-11-04
Add option to get specific amount of articles

### LINKED ISSUE #940: .rp improvement
## Description

The `.rp` command currently gives 5 results by default but what if you wanted to get only 1-4? There should be a option to get amount you want by still keeping the max 5 as default. If no amount provided default value, 5, will be returned.

## Reasoning

Getting only specific amount of results if you don't want to see more, and could possibly be good to limit results if wanted.
## Proposed Implementation

Currently I don't have any in code implementations, but some sort of limit would be added in there.

Nipa-Code:
## Relevant Issues

Closes #940 

## Description

Now the `.rp` command has option to get specific amount of articles in between 1-5 (inclusive). I added new parameter to `.rp` which is `amount`. If no argument given when command is ran, the default value 5 will be used. If user tries to give negative, 0 or bigger than 5 , the bot will send a message saying "`amount must be between 1 and 5 (inclusive).`". 
## Did you:

### DIFF
--- a/bot/exts/utilities/realpython.py
+++ b/bot/exts/utilities/realpython.py
@@ -1,5 +1,6 @@
 import logging
 from html import unescape
+from typing import Optional
 from urllib.parse import quote_plus
 
 from discord import Embed
@@ -7,7 +8,6 @@ from discord.ext import commands
 
 from bot.bot import Bot
 from bot.constants import Colours
-from typing import Optional
 
 logger = logging.getLogger(__name__)
 
--- a/bot/exts/utilities/realpython.py
+++ b/bot/exts/utilities/realpython.py
@@ -41,6 +41,7 @@ class RealPython(commands.Cog):
         if not 1 <= amount <= 5:
             await ctx.send("`amount` must be between 1 and 5 (inclusive).")
             return
+
         params = {"q": user_search, "limit": amount, "kind": "article"}
         async with self.bot.http_session.get(url=API_ROOT, params=params) as response:
             if response.status != 200:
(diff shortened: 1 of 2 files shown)


bot/exts/utilities/realpython.py
```diff
@@ -31,9 +32,16 @@ def __init__(self, bot: Bot):
...
     @commands.cooldown(1, 10, commands.cooldowns.BucketType.user)
-    async def realpython(self, ctx: commands.Context, *, user_search: str) -> None:
-        """Send 5 articles that match the user's search terms."""
-        params = {"q": user_search, "limit": 5, "kind": "article"}
+    async def realpython(self, ctx: commands.Context, amount: Optional[int] = 5, *, user_search: str) -> None:
+        """
+        Send `amount` articles that match the search terms (`user_search`).
+
+        `amount` must be between 1 and 5 (inclusive). If it's not, an error message is sent.
+        """
```
ChrisLovering, line 40: I'm not too fond of having annotated doc strings like this for commands, since this is the help that will be given to the user I think it should read as "proper" English. Maybe something more like this
```suggestion
        """
        Send some articles from RealPython that match the search terms.

        By default the top 5 matches are sent, this can be overwritten to
        a number between 1 and 5 by specifying an amount before the search query.
        """
```

ChrisLovering (changes requested):

ChrisLovering (approved):

hedyhli (approved):

PR #957, opened 2021-11-29
Aoc count timestamp

ChrisLovering:
## Description

Swapped out our custom logic to create a human readable delta with a Discord time stamp.

This allows users to hover over the time stamp to get an exact date, along with automatically rounding to the most significant unit.

I have also added a refactor commit to swap the conditional block around, so the "smaller" code path is the one in the if statement, which makes the function easier to read.

## Screenshots
Before:
image
image

After:
image
Discord time stamps round, rather than truncate like we were doing previously, which is why this says 2 days, rather than 1.
image

## Did you:


Akarys42 (approved):

jchristgit (approved):
Clean and concise. I like it.

PR #963, opened 2021-12-02
Make self_placement_name keyword-only in .aoc lb

bsoyka:
## Relevant Issues

Approved by @janine9vn: https://discord.com/channels/267624335836053506/635950537262759947/915806212367482891

Closes #962

## Description

I made `.adventofcode leaderboard`'s `self_placement_name` argument keyword-only to accept multi-word usernames without needing quotation marks.

## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] Read all the comments in this template?
- [x] Ensure there is an issue open, or link relevant discord discussions?
- [x] Read and agree to the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

PR #964, opened 2021-12-02
Add `.aoc link` command

### LINKED ISSUE #958: `.aoc link <aoc_name>` to link Discord Account and Advent of Code Name
## Description
To support something I'd like to do this year regarding assigned roles, there needs to be a nice way for people to associate their Discord account IDs with their name/account of Advent of Code. I'd like to shift the bulk of the work of figuring this out away from myself and to the users themselves. I'd like them to create the "links" between their accounts.

## Proposed Implementation
Steal the Hacktober stats implementation!
The user should be able to run `.aoc link <aoc_name>`. This will create a mapping in Redis from their Discord ID to their Advent of Code name that's on our leaderboard. If a mapping to same aoc_name exists for a different Discord ID, it should throw an error and tell the user to contact an admin if there's an error.

## Additional Details
If someone is feeling really ambitious, `.aoc lb` with no other arguments can check to see if this link exists, and if it does then automatically show their score on top. Essentially a shortcut for `.aoc lb <aoc_name>`

I would like this implemented within the next two weeks ideally, so if someone is up for that challenge please say so. Otherwise I'll pick this up myself.

janine9vn:
## Description
This adds functionality to the Advent of Code cog for users to "link" their Discord Account with their Advent of Code name on the leaderboard. The cog doesn't explicitly make use of it, but it is necessary to exist and be functional for something I want to do. 

This is implemented with a Redis Cache. A user can link their discord account with a name they specify and it's stored as a key-value pair with the Discord ID as the key.

There is no validation that the name they are entering is on our leaderboard. This is an intentional decision as I don't want to rely on the leaderboard cooldown for a user to successfully run this command. If a user fucks up their name, that's on them. It does check if the name they are trying to link is already in the cache and provides the appropriate response.

To test this command you can run the following two commands in different scenarios:
`.aoc link`
`.aoc link aocnamehere`

## Relevant Issues
Closes #958

### DIFF
--- a/bot/exts/events/advent_of_code/_cog.py
+++ b/bot/exts/events/advent_of_code/_cog.py
@@ -30,6 +30,7 @@ AOC_WHITELIST = AOC_WHITELIST_RESTRICTED + (Channels.advent_of_code,)
 class AdventOfCode(commands.Cog):
     """Advent of Code festivities! Ho Ho Ho!"""
 
+    # Redis Cache for linking Discord IDs to Advent of Code usernames
     account_links = RedisCache()
 
     def __init__(self, bot: Bot):
@@ -186,23 +187,9 @@ class AdventOfCode(commands.Cog):
         """
         Link your Discord Account to your Advent of Code name.
 
-        Stored in a Redis Cache, Discord ID: Advent of Code Name
+        Stored in a Redis Cache with the format of `Discord ID: Advent of Code Name`
         """
         cache_items = await self.account_links.items()
-
-        # A short circuit in case the cache is empty
-        if len(cache_items) == 0 and aoc_name:
-            log.info(f"{ctx.author} ({ctx.author.id}) is now linked to {aoc_name}.")
-            await self.account_links.set(ctx.author.id, aoc_name)
-            await ctx.reply(f"You have linked your Discord ID to {aoc_name}.")
-            return
-        elif len(cache_items) == 0:
-            await ctx.reply(
-                "You have not linked an Advent of Code account."
-                "Please re-run the command with one specified."
-            )
-            return
-
         cache_aoc_name = [value for _, value in cache_items]
 
         if aoc_name:
--- a/bot/exts/events/advent_of_code/_cog.py
+++ b/bot/exts/events/advent_of_code/_cog.py
@@ -228,6 +228,27 @@ class AdventOfCode(commands.Cog):
                     " Please re-run the command with one specified."
                 )
 
+    @in_month(Month.NOVEMBER, Month.DECEMBER)
+    @adventofcode_group.command(
+        name="unlink",
+        aliases=("disconnect",),
+        brief="Tie your Discord account with your Advent of Code name."
+    )
+    @whitelist_override(channels=AOC_WHITELIST)
+    async def aoc_unlink_account(self, ctx: commands.Context) -> None:
+        """
+        Unlink your Discord ID with your Advent of Code leaderboard name.
+
+        Deletes the entry that was Stored in the Redis cache.
+        """
+        if aoc_cache_name := await self.account_links.get(ctx.author.id):
+            log.info(f"Unlinking {ctx.author} ({ctx.author.id}) from Advent of Code account {aoc_cache_name}")
+            await self.account_links.delete(ctx.author.id)
+            await ctx.reply(f"We have removed the link between your Discord ID and {aoc_cache_name}.")
+        else:
+            log.info(f"Attempted to unlink {ctx.author} ({ctx.author.id}), but not link was found.")
+            await ctx.reply("You don't have an Advent of Code account linked.")
+
     @in_month(Month.DECEMBER)
     @adventofcode_group.command(
         name="dayandstar",
--- a/bot/exts/events/advent_of_code/_cog.py
+++ b/bot/exts/events/advent_of_code/_cog.py
@@ -210,12 +210,12 @@ class AdventOfCode(commands.Cog):
 
             # Update an existing link
             if old_aoc_name := await self.account_links.get(ctx.author.id):
-                log.info(f"{ctx.author} ({ctx.author.id}) has changed their link from {old_aoc_name} to {aoc_name}.")
+                log.info(f"Changing link for{ctx.author} ({ctx.author.id}) from {old_aoc_name} to {aoc_name}.")
                 await self.account_links.set(ctx.author.id, aoc_name)
                 await ctx.reply(f"Your linked account has been changed to {aoc_name}.")
             else:
                 # Create a new link
-                log.info(f"{ctx.author} ({ctx.author.id}) is now linked to {aoc_name}.")
+                log.info(f"Linking {ctx.author} ({ctx.author.id}) to account {aoc_name}.")
                 await self.account_links.set(ctx.author.id, aoc_name)
                 await ctx.reply(f"You have linked your Discord ID to {aoc_name}.")
         else:
@@ -246,7 +246,7 @@ class AdventOfCode(commands.Cog):
             await self.account_links.delete(ctx.author.id)
             await ctx.reply(f"We have removed the link between your Discord ID and {aoc_cache_name}.")
         else:
-            log.info(f"Attempted to unlink {ctx.author} ({ctx.author.id}), but not link was found.")
+            log.info(f"Attempted to unlink {ctx.author} ({ctx.author.id}), but no link was found.")
             await ctx.reply("You don't have an Advent of Code account linked.")
 
     @in_month(Month.DECEMBER)
--- a/bot/exts/events/advent_of_code/_cog.py
+++ b/bot/exts/events/advent_of_code/_cog.py
@@ -183,7 +183,7 @@ class AdventOfCode(commands.Cog):
         brief="Tie your Discord account with your Advent of Code name."
     )
     @whitelist_override(channels=AOC_WHITELIST)
-    async def aoc_link_account(self, ctx: commands.Context, aoc_name: str = None) -> None:
+    async def aoc_link_account(self, ctx: commands.Context, *, aoc_name: str = None) -> None:
         """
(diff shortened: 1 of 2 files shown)


bot/exts/events/advent_of_code/_cog.py
```diff
@@ -172,6 +175,72 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+                await ctx.reply(f"You have linked your Discord ID to {aoc_name}.")
+        else:
+            # User has not supplied a name, let's check if they're in the cache or not
+            if cache_name := await self.account_links.get(ctx.author.id):
+                await ctx.reply(f"You have already linked an Advent of Code account: {cache_name}.")
+            else:
+                await ctx.reply(
+                    "You have not linked an Advent of Code account."
+                    " Please re-run the command with one specified."
+                )
```
Shivansh-007, line 242: This could probably be the first part of the if statement with `if not aoc_name` and then we could return early to avoid one extra indentation 🙃

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -172,6 +175,72 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+            if old_aoc_name := await self.account_links.get(ctx.author.id):
+                log.info(f"{ctx.author} ({ctx.author.id}) has changed their link from {old_aoc_name} to {aoc_name}.")
+                await self.account_links.set(ctx.author.id, aoc_name)
+                await ctx.reply(f"Your linked account has been changed to {aoc_name}.")
+            else:
+                # Create a new link
+                log.info(f"{ctx.author} ({ctx.author.id}) is now linked to {aoc_name}.")
+                await self.account_links.set(ctx.author.id, aoc_name)
+                await ctx.reply(f"You have linked your Discord ID to {aoc_name}.")
+        else:
```
Shivansh-007, line 234: You can probably move the `await self.account_links.set()` line outside(after it) the if..else statement as it is the same.

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -172,6 +175,72 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+            log.info(f"{ctx.author} ({ctx.author.id}) is now linked to {aoc_name}.")
+            await self.account_links.set(ctx.author.id, aoc_name)
+            await ctx.reply(f"You have linked your Discord ID to {aoc_name}.")
+            return
+        elif len(cache_items) == 0:
+            await ctx.reply(
+                "You have not linked an Advent of Code account."
+                "Please re-run the command with one specified."
+            )
+            return
```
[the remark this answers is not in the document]
janine9vn: If the cache is empty, later code trying to pull items from the cache will fail. Additionally, this will help with future development. Did you test this locally?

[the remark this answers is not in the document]
Shivansh-007: Yeah, it works

Screenshot from 2021-12-03 07-34-49

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -172,6 +175,72 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+        brief="Tie your Discord account with your Advent of Code name."
+    )
+    @whitelist_override(channels=AOC_WHITELIST)
+    async def aoc_link_account(self, ctx: commands.Context, aoc_name: str = None) -> None:
+        """
+        Link your Discord Account to your Advent of Code name.
+
+        Stored in a Redis Cache, Discord ID: Advent of Code Name
+        """
+        cache_items = await self.account_links.items()
```
onerandomusername, line 191: I see no way to unlink an account, I feel like that should be possible.

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -172,6 +175,72 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
             await ctx.message.add_reaction(Emojis.envelope)
 
+    @in_month(Month.NOVEMBER, Month.DECEMBER)
+    @adventofcode_group.command(
+        name="link",
+        aliases=("connect",),
+        brief="Tie your Discord account with your Advent of Code name."
+    )
+    @whitelist_override(channels=AOC_WHITELIST)
+    async def aoc_link_account(self, ctx: commands.Context, aoc_name: str = None) -> None:
```
onerandomusername, line 185: why is aoc_name none here?

onerandomusername (changes requested):

onerandomusername:
Huh, I left a overall message which strangely did not get saved.

I was mentioned that it looks pretty good, but I'd like to see an unlink command. In addition, a way to unlink from a moderator view in case someone adds an account they cannot control. I also do not see an admin/mod or event lead command to be able to force an unlink.

bot/exts/events/advent_of_code/_cog.py
janine9vn: The default is None in case the user just does `.aoc link` with no specified name. I can check the cache if they've already linked or now and then return the status. It doesn't need to error out if they don't remember linking or not.

onerandomusername: I see, that makes sense. Given the changes with #963, I feel like this command the leaderboard now have different input styles, which does not make the most sense.

janine9vn: I'm not sure what you mean by this. All the user has to do is provide their advent of code leaderboard name and it gets linked, if they don't it'll tell them if they've linked it or not. It's the same as the `.aoc lb aoc_name` with that PR being merged.

bot/exts/events/advent_of_code/_cog.py
janine9vn: What would be the benefit of doing so here? I want to make sure that the link happens before I send the message back to the user in case something goes wonky. I don't want to tell the user that the link has happened before it's actually happened.

bot/exts/events/advent_of_code/_cog.py
onerandomusername: What #963 did is it made the user argument keyword only.

This means that a user can run `.aoc link name with spaces` and not need quotes. Given that this is what the bot now has and our users expect it to work (at least in the scope of the aoc commands), this command should not require quotes for a username argument.

This is not the same, as I've tested locally.
image

Current behavior treats it as a kw arg, but also strips the first set of `"` since that was the initial behavior. 
image

janine9vn: Ah, okay. So your feedback is to make this keyword argument only to accept spaces in the name without quotes. I can do that. That wasn't clear from your original statement.

bot/exts/events/advent_of_code/_cog.py
Shivansh-007: No real benefits, would just not duplicate code, looking at the log statements, I think we could even set the new `aoc_name` before the if...else block, since the logging statements are in past, saying the change has already happened.

onerandomusername: I agree with some of what Jason is saying.

since Since the logging statement is before it happens, it should be in present or future tense.

If I was in a late debugging session trying to figure out what happened, I would think that the account successfully got linked, until I dug a bit further.

Changing the tense would save some time, and make more sense, as being past tense implies it has happened, when it hasn't.

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -175,6 +179,79 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+    )
+    @whitelist_override(channels=AOC_WHITELIST)
+    async def aoc_link_account(self, ctx: commands.Context, *, aoc_name: str = None) -> None:
+        """
+        Link your Discord Account to your Advent of Code name.
+
+        Stored in a Redis Cache with the format of `Discord ID: Advent of Code Name`
+        """
+        cache_items = await self.account_links.items()
+        cache_aoc_name = [value for _, value in cache_items]
```
ChrisLovering, line 196:
```suggestion
        cached_aoc_names = [value for _, value in cache_items]
```

This is quite a small change, but imo will help readability where this var is used further on in the function, as it will indicate it's more than more name.

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -175,6 +179,79 @@ async def join_leaderboard(self, ctx: commands.Context) -> None:
...
+                )
+                await ctx.reply(
+                    f"{aoc_name} is already tied to another account."
+                    " Please contact an admin if you believe this is an error."
+                )
+                return
+
+            # Update an existing link
+            if old_aoc_name := await self.account_links.get(ctx.author.id):
+                log.info(f"Changing link for{ctx.author} ({ctx.author.id}) from {old_aoc_name} to {aoc_name}.")
```
ChrisLovering, line 216:
```suggestion
                log.info(f"Changing link for {ctx.author} ({ctx.author.id}) from {old_aoc_name} to {aoc_name}.")
```

ChrisLovering (approved):
Looks good, and tested locally 👍

Two minor changes that would be welcomed, but shouldn't block this from being merged.

Something I noticed while testing is that if you try to run the leaderboard command and give it a name (or now as of this PR you have linked your GitHub), if you haven't join the leaderboard the message given isn't the best.
image

I think this would be better if it said something like `We can't seem to find your profile in our leaderboard. Try running .aoc join to get an invite!`

D0rs4n (approved):
Looks really good to me. I tested it out locally, and it works perfectly fine.
Thank you!

bot/exts/events/advent_of_code/_cog.py
Shivansh-007: No comment on this and it's not mentioned, yet it is resolved. 

Looking at the code again, it wouldn't really create much of a difference so could be skipped.

PR #969, opened 2021-12-03
nit: make last refreshed time above the rest of the embed

### LINKED ISSUE #966: aoc buff: remove the leaderboard command, and make it an updating message
## Description

Too many people are trying to get the leaderboard rapidly. Since it only updates once every 30 minutes, a channel of #aoc-leaderboard would be nice while the game is afoot. This would automatically update every 30 minutes, as the leaderboard currently does. The current aoc lb command would link to the #aoc-leaderboard channel, maybe showing the user where they currently are, but linking to the channel to view the full lb.
## Reasoning

image

image

## Proposed Implementation

Special channel which will feature the leaderboard, updating every 30 minutes

I do not care, if I can get to it tomorrow I will attempt to implement it myself, but if someone else has an idea for it, they can go for it as well.

### LINKED ISSUE #967: aoc leaderboard: use a discord formatted timestamp for when the lb can be next updated
> Additionally, we would be able to use discord timestamps, to show a helpful message in the lb command of how long until the lb can be refreshed again.

This may be worth a second and different issue, given that this can be implemented regardless of the rest of this pr.

_Originally posted by @onerandomusername in https://github.com/python-discord/sir-lancebot/issues/966#issuecomment-985238562_

Given the amount of updating, and times ran the command, it would helpful to see a timestamp that means how long until the leaderboard will be refreshed, rather than "every 30 minutes", which as mentioned in the above issue's comments, isn't entirely accurate anyways.

onerandomusername:
discussed on discord, using this implementation from zig: 
image

closes #967 
closes #966 (indirectly)

new embed: 
image

Edit: did not link the discussion: https://canary.discord.com/channels/267624335836053506/635950537262759947/916405655630389310

### DIFF
--- a/bot/exts/events/advent_of_code/_helpers.py
+++ b/bot/exts/events/advent_of_code/_helpers.py
@@ -332,7 +332,7 @@ async def fetch_leaderboard(invalidate_cache: bool = False, self_placement_name:
         number_of_participants = len(leaderboard)
         formatted_leaderboard = _format_leaderboard(leaderboard)
         full_leaderboard_url = await _upload_leaderboard(formatted_leaderboard)
-        leaderboard_fetched_at = datetime.datetime.utcnow().isoformat()
+        leaderboard_fetched_at = datetime.datetime.now(datetime.timezone.utc).isoformat()
 
         cached_leaderboard = {
             "placement_leaderboard": json.dumps(raw_leaderboard_data),
--- a/bot/exts/events/advent_of_code/_helpers.py
+++ b/bot/exts/events/advent_of_code/_helpers.py
@@ -368,11 +368,13 @@ def get_summary_embed(leaderboard: dict) -> discord.Embed:
     """Get an embed with the current summary stats of the leaderboard."""
     leaderboard_url = leaderboard["full_leaderboard_url"]
     refresh_minutes = AdventOfCode.leaderboard_cache_expiry_seconds // 60
+    refreshed_unix = int(datetime.datetime.fromisoformat(leaderboard["leaderboard_fetched_at"]).timestamp())
 
-    aoc_embed = discord.Embed(
-        colour=Colours.soft_green,
-        timestamp=datetime.datetime.fromisoformat(leaderboard["leaderboard_fetched_at"]),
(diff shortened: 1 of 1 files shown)


Shivansh-007 (approved):

Xithrius (approved):
That's some good stuff right there

bot/exts/events/advent_of_code/_helpers.py
```diff
@@ -368,11 +368,13 @@ def get_summary_embed(leaderboard: dict) -> discord.Embed:
...
 
-    aoc_embed = discord.Embed(
-        colour=Colours.soft_green,
-        timestamp=datetime.datetime.fromisoformat(leaderboard["leaderboard_fetched_at"]),
-        description=f"*The leaderboard is refreshed every {refresh_minutes} minutes.*"
+    aoc_embed = discord.Embed(colour=Colours.soft_green)
+
+    aoc_embed.description = (
+        f"The leaderboard is refreshed every {refresh_minutes} minutes.\n"
+        f"Last Updated: <t:{refreshed_unix}:t>"
```
janine9vn, line 377: So, there's actually a bug with the current implementation of how `leaderboard_fetched_at` is created, impacting this change.

If the bot doesn't run in UTC timezone, then the timestamp returned will be incorrect and X hours (where X is the utc offset) off from whatever timezone the bot is currently being run in. It's so far been unnoticed because in production it does run in UTC. However, this is still a bug as it's dependent on which timezone the bot runs in and impacts local development for those not in the same timezone as UTC.

The fix is changing line 335 in this file to:
```py
leaderboard_fetched_at = datetime.datetime.now(datetime.timezone.utc).isoformat()
```

janine9vn (changes requested):

bot/exts/events/advent_of_code/_helpers.py
onerandomusername: I've seen multiple places in both bots which rely on UTC timezone so I figured this was a non-issue.

Nonetheless, fixed in https://github.com/python-discord/sir-lancebot/pull/969/commits/4c451b170e0005726dac2489835045c1dc43ce62

Additionally did a rebase since there were changes to this cog.

janine9vn (approved):

PR #971, opened 2021-12-03
yank lovefest command

### LINKED ISSUE #970: yank role commands
## Description

Remove the subscribe commands from lance, because as of https://github.com/python-discord/bot/pull/1868, this is now supported on Bot. (Beautifully executed, too)

EDIT: Rather than removing the commands, they will be replaced with a message that tells the user to use the command on bot.

## Reasoning

No sense in having two commands that do the same thing, more maintenance work!

## Proposed Implementation

**TODO:**
- [x] remove lovefest role command
- [ ] remove advent of code command (deferred until January, when the command can no longer be assigned or removed)

onerandomusername:
now that !subscribe exists on Python, the lovefest command is irrelevant

## Relevant Issues

Partially closes #970 -- lovefest migration

## Description

Added a custom error that would be handled by the error handler, in order to deal with this migrations in the future. This has the benefit of only needing to raise an exception with the new command, and the error will be logged and an embed shown to the user.

Because there are not stats on lance, I intentionally left not deleting the embed, as this can be a makeshift stat by searching for instances of lance's messages mentioning the new command to use.

## Did you:


bot/exts/holidays/valentines/be_my_valentine.py
```diff
@@ -30,7 +32,7 @@ def load_json() -> dict:
         return loads(p.read_text("utf8"))
 
     @in_month(Month.FEBRUARY)
-    @commands.group(name="lovefest")
+    @commands.command(name="lovefest", help=f"NOTE: This command has been moved to {MOVED_COMMAND}")
```
ChrisLovering, line 35: Could you please move this help string to the command's docstring. Our help command uses those to populate the help text.

bot/exts/holidays/valentines/be_my_valentine.py
```diff
@@ -30,7 +32,7 @@ def load_json() -> dict:
...
     async def lovefest_role(self, ctx: commands.Context) -> None:
```
ChrisLovering, line 36: Could you add `, *, _)` to the function signature here, so that if a user runs `.lovefest sub` this command still gets invoked and they are told about the move, rather than getting a bag arg error.

ChrisLovering (changes requested):

bot/exts/holidays/valentines/be_my_valentine.py
onerandomusername: The reason its here is because its dynamically using the configured prefix of the other bot. Additionally, this is used for the help command even in this location.

bot/exts/holidays/valentines/be_my_valentine.py
onerandomusername: Not necessary. 
image

bot/exts/holidays/valentines/be_my_valentine.py
onerandomusername: Setting help in the command decorator overrides the callback docstring, as shown below

image
image

ChrisLovering: Ahh alright cool, needing dynamic help here is a good reason. Could you hosted the docstring to reflect the new behaviour? It still references what it used to do.

bot/exts/holidays/valentines/be_my_valentine.py
ChrisLovering: Ah nice, that works

bot/exts/holidays/valentines/be_my_valentine.py
onerandomusername: cfb2cc69

ChrisLovering (approved):

PR #973, opened 2021-12-05
Make error messages more consistent in the AoC daystar view

D0rs4n:
## Description

Changes in the AoC DayAndStarView:
 - From now on, when the interacting user and the original author of the view is different, the bot will send an ephemeral message, regarding the issue, instead of the default error message.
## Did you:


ks129 (changes requested):
It is not so high priority, so I think this should be solved with exception raising and handling in global error handler. This would remove possible future duplication.

D0rs4n:
> It is not so high priority, so I think this should be solved with exception raising and handling in global error handler. This would remove possible future duplication.

I'm not sure how that would work in the interaction environment.
If an error is raised in the `interaction_check`, it'll call the `View.on_error`([See](https://discordpy.readthedocs.io/en/master/api.html?highlight=interaction_check#discord.ui.View.on_error)): `The default implementation prints the traceback to stderr.`
So that would have to be overriden as well, accordingly.
Hence, I believe it's easier, to just send an interaction response.
(You would have to do the same procedure on every View.)

bot/exts/events/advent_of_code/views/dayandstarview.py
```diff
@@ -42,7 +42,10 @@ def generate_output(self) -> str:
 
     async def interaction_check(self, interaction: discord.Interaction) -> bool:
         """Global check to ensure that the interacting user is the user who invoked the command originally."""
-        return interaction.user == self.original_author
+        if interaction.user != self.original_author:
+            await interaction.response.send_message(":x: This is not your view to interact with!", ephemeral=True)
```
janine9vn, line 46: My only comment is that while this message makes sense for someone familiar with interactions and discord.py, it may not be clear to those who don't know what a "view" means.

We could maybe reword this to something like: "You can't interact with someone else's response. Please run the command yourself."

janine9vn (changes requested):
Small change on the wording of the message, but otherwise thank you for the PR~

bot/exts/events/advent_of_code/views/dayandstarview.py
D0rs4n: That's a valid point.
I've pushed a commit that makes the error message a bit more understandable for everyone!

janine9vn (approved):
beautiful, thank you!

wookie184 (approved):
Works nicely, very cool!

Issue #968, opened 2021-12-03
github command enhancements


onerandomusername:
## Description

(Full list of changes is located on #778's review comments)
- constants
Some constants are unused, or not used well. 
- move ratelimits
Ratelimits are only handled after a single call to fetch data, would make most sense to handle them in the fetch_data method
- remove git alias
while we're here, this should go as github is not git, and this allows future use of `git` as a command or tag.
- add wait for deletion
if a user wants to delete the embed, they will now be able to. In case they didn't mean to link the issues.
- include issue number in response
With several issues requested, the issue numbers should be shown
- move the if not message.guild check to before we scan for issues
no reason to check for issues in a dm message, ever
- make only one request per pull instead of two
current implementation makes a second request to get pull info, but all of the info is returned on the first call
- keep the order the same
issue order is currently converted to a set, which loses the initial order.

## Reasoning

General command enhancements, small performance boost. Most notably, only makes one request per pull instead of two.

## Proposed Implementation

Implementation can be found on the review comments of #778 

## Would you like to implement this yourself?

- [x] I'd like to implement this feature myself
- [ ] Anyone can implement this feature

**NOTE: This is stalled until #778 is merged**

onerandomusername:
New change: 

As per this, https://canary.discord.com/channels/267624335836053506/635950537262759947/931292179744297010, github api can return a merged pr as still a draft for some ungodly reason, so the ordering of the emojis will be changed to support merged prs still being a draft..

DMFriends:
In case this hasn't been considered already, I think it would also be nice to be able to delete the embed with the issues/PR's if you type the wrong number by mistake. Or possibly a better idea is to have an edit listener so if you typed `sir-lancebot#654` but you meant to type `sir-lancebot#653`, you could edit the message to say `sir-lancebot#653` and that would automatically edit the embed with the new issue number.

PR #778, opened 2021-06-29
Merge automatic issue linking into the github cog

ToxicKidz:
## Relevant Issues

Closes #711
Closes #670

## Description

The automatic linking of issues was moved into the github cog, and the `.issue` command was removed (see #788).

## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] Read all the comments in this template?
- [x] Ensure there is an issue open, or link relevant discord discussions?
- [x] Read the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

PR #979, opened 2021-12-10
Fixed Key Error: "draft" bug

### LINKED ISSUE #928: KeyError: 'draft'
Sentry Issue: [SIR-LANCEBOT-6X](https://sentry.io/organizations/python-discord/issues/2743337733/?referrer=github_integration)

```
KeyError: 'draft'
  File "discord/ext/commands/core.py", line 167, in wrapped
  File "bot/exts/utilities/issues.py", line 211, in issue
    results = [await self.fetch_issues(number, repository, user) for number in numbers]
  File "bot/exts/utilities/issues.py", line 211, in <listcomp>
    results = [await self.fetch_issues(number, repository, user) for number in numbers]
  File "bot/exts/utilities/issues.py", line 142, in fetch_issues
    if pull_data["draft"]:

Unhandled command error: 'draft'
```

katimoth:
## Relevant Issues

Closes #928 

## Description

Added a key check to the if statement in L142 so that pull_data["draft"] will only be checked if the key exists in the parsed json
## Did you:

### DIFF
--- a/bot/exts/utilities/issues.py
+++ b/bot/exts/utilities/issues.py
@@ -139,15 +139,19 @@ class Issues(commands.Cog):
             log.trace(f"PR provided, querying GH pulls API for additional information: {pulls_url}")
             async with self.bot.http_session.get(pulls_url) as p:
                 pull_data = await p.json()
-                if "draft" in pull_data and pull_data["draft"]:
-                    emoji = Emojis.pull_request_draft
-                elif pull_data["state"] == "open":
-                    emoji = Emojis.pull_request_open
-                # When 'merged_at' is not None, this means that the state of the PR is merged
-                elif pull_data["merged_at"] is not None:
-                    emoji = Emojis.pull_request_merged
-                else:
-                    emoji = Emojis.pull_request_closed
+                try:
+                    if pull_data["draft"]:
+                        emoji = Emojis.pull_request_draft
+                    elif pull_data["state"] == "open":
+                        emoji = Emojis.pull_request_open
+                    # When 'merged_at' is not None, this means that the state of the PR is merged
+                    elif pull_data["merged_at"] is not None:
+                        emoji = Emojis.pull_request_merged
+                    else:
+                        emoji = Emojis.pull_request_closed
+                except KeyError:
+                    # couldn't get pull data
+                    emoji = Emojis.x_square
 
         issue_url = json_data.get("html_url")
 
--- a/bot/exts/utilities/issues.py
+++ b/bot/exts/utilities/issues.py
@@ -139,7 +139,7 @@ class Issues(commands.Cog):
             log.trace(f"PR provided, querying GH pulls API for additional information: {pulls_url}")
             async with self.bot.http_session.get(pulls_url) as p:
                 pull_data = await p.json()
-                if pull_data["draft"]:
+                if "draft" in pull_data and pull_data["draft"]:
                     emoji = Emojis.pull_request_draft
                 elif pull_data["state"] == "open":
                     emoji = Emojis.pull_request_open
--- a/bot/exts/utilities/issues.py
+++ b/bot/exts/utilities/issues.py
@@ -140,12 +140,12 @@ class Issues(commands.Cog):
             async with self.bot.http_session.get(pulls_url) as p:
                 pull_data = await p.json()
                 try:
-                    if pull_data["draft"]:
+                    if pull_data.get("draft"):
                         emoji = Emojis.pull_request_draft
-                    elif pull_data["state"] == "open":
+                    elif pull_data.get("state") == "open":
                         emoji = Emojis.pull_request_open
                     # When 'merged_at' is not None, this means that the state of the PR is merged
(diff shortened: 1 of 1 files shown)


bot/exts/utilities/issues.py
```diff
@@ -139,7 +139,7 @@ async def fetch_issues(
             log.trace(f"PR provided, querying GH pulls API for additional information: {pulls_url}")
             async with self.bot.http_session.get(pulls_url) as p:
                 pull_data = await p.json()
-                if pull_data["draft"]:
+                if "draft" in pull_data and pull_data["draft"]:
```
onerandomusername, line 142: `if pull_data.get("draft"):`

https://docs.python.org/3/library/stdtypes.html#dict.get

onerandomusername (commented):
looks good, minor suggestion

bot/exts/utilities/issues.py
```diff
@@ -139,7 +139,7 @@ async def fetch_issues(
...
                     emoji = Emojis.pull_request_draft
                 elif pull_data["state"] == "open":
```
fisher60, line 144: From what I have collected, `draft` should *always* be returned in a valid response from the API. Therefore if `draft` was missing, it likely means the entire response was something unexpected (perhaps empty). So we could expect each one of these elifs to fail as well. It may be prudent to utilize a try-except here, or explicitly catch a missing key from any if-elif block. From what I can see, the current will just cause the error to occur on the first elif, since that key would probably be missing as well.

fisher60 (changes requested):

bot/exts/utilities/issues.py
fisher60: API docs for reference https://docs.github.com/en/rest/reference/pulls

katimoth: @fisher60 What should the emoji be set to if the pull_data is invalid?

onerandomusername: This would be fixed as part of #968, although I believe #778 would fix this as I think it moves all requests to a helper method which checks for a 200 response code and whatnot.

katimoth: I still think a try-except would be a good idea in case the pull data is malformed.  #778 still leaves the possibility of a KeyError.

katimoth (commented):
Added @fisher60 and @onerandomusername's suggestions

bot/exts/utilities/issues.py
onerandomusername: If its malformed, we shouldn't be doing anything with the pull data. A solution would be to use a helper method to get the data, deal with possible ratelimits, and check for a 200 status code.

Most of the calls are deferred to a helper method with #778, but not all of them, unfortunately. 

Additionally, this particular call to the api affecting this bug will be removed as a result of #968, as the data is already there with the initial issues api request.

The solution, if we are to implement it in this pull request, is to check if [L140](https://github.com/python-discord/sir-lancebot/blob/0d87e9ae6e44937af5e3b64a7cbe329660f9fff9/bot/exts/utilities/issues.py#L140) does not have a 200 response code, and deal with that accordingly.

onerandomusername: If its malformed, we shouldn't be doing anything with the pull data. A solution would be to use a helper method to get the data, deal with possible ratelimits, and check for a 200 status code.

Most of the calls are deferred to a helper method with #778, but not all of them, unfortunately. 

Additionally, this particular call to the api affecting this bug will be removed as a result of #968, as the data is already there with the initial issues api request.

The solution, if we are to implement it in this pull request, is to check if [L140](https://github.com/python-discord/sir-lancebot/blob/0d87e9ae6e44937af5e3b64a7cbe329660f9fff9/bot/exts/utilities/issues.py#L140) does not have a 200 response code, and deal with that accordingly.

onerandomusername (changes requested):

bot/exts/utilities/issues.py
```diff
@@ -139,15 +139,19 @@ async def fetch_issues(
...
+                    elif pull_data.get("state") == "open":
+                        emoji = Emojis.pull_request_open
+                    # When 'merged_at' is not None, this means that the state of the PR is merged
+                    elif pull_data.get("merged_at") is not None:
+                        emoji = Emojis.pull_request_merged
+                    else:
+                        emoji = Emojis.pull_request_closed
+                except KeyError:
+                    # couldn't get pull data
+                    emoji = Emojis.x_square
```
Shivansh-007, line 154: Where is key error raised? All are using `.get()` which returns `None` if the key isn't present on the dict object (it doesn't raise a `KeyError` like `pull_data["draft"]` does).

MaskDuck (changes requested):
we dont need a try except since `.get()` return `None` if not found.

Xithrius:
@katimoth What's your status on this PR?

Xithrius:
User hasn't responded, putting this up for grabs.

Xithrius:
I will be creating another PR to solve this issue.

PR #944, opened 2021-11-14
minor: allow color command in dev-media

onerandomusername:
## Relevant Issues

Context: https://discord.com/channels/267624335836053506/635950537262759947/909560534381252619

## Description

Added decorator to allow bot in the dev and media categories. 
## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] Read all the comments in this template?
- [x] Ensure there is an issue open, or link relevant discord discussions?
- [x] Read and agree to the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

PR #981, opened 2021-12-16
fix: subcommands inherit their parent's whitelist

onerandomusername:
## Relevant Issues

https://discord.com/channels/267624335836053506/635950537262759947/920817429301579807

Additionally, does what #944 was *supposed* to do.

image

## Description

solves issue with adding decorator to the parent
which wouldn't apply to the children

Essentially, whitelists for parent commands now apply to child commands

## Did you:

----
Given the reviews on #944, @ChrisLovering, @Xithrius


Xithrius:
@onerandomusername conflicts.

bot/utils/decorators.py
```diff
@@ -203,20 +203,36 @@ def predicate(ctx: Context) -> bool:
 
         kwargs = default_kwargs.copy()
 
-        # Update kwargs based on override
-        if hasattr(ctx.command.callback, "override"):
+        # determine which command's overrides we will use
+        # as we have groups, we want to ensure that group commands inherit from the parent
+        overridden_command: Union[commands.Command, commands.Group] = None
+        for command in [ctx.command, *ctx.command.parents]:
+            print(command)
```
Xithrius, line 210: Is this intentional? If not,
```suggestion
```

onerandomusername: https://github.com/python-discord/sir-lancebot/pull/981/commits/1e69c3a4c7d799c5a1450e59c9d390097778d3b8

bot/utils/decorators.py
```diff
@@ -199,23 +199,38 @@ def predicate(ctx: Context) -> bool:
         kwargs = default_kwargs.copy()
         allow_dms = False
 
-        # Update kwargs based on override
-        if hasattr(ctx.command.callback, "override"):
+        # determine which command's overrides we will use
+        # as we have groups, we want to ensure that group commands inherit from the parent
```
Bluenix2, line 203: The way this comment is unevenly split the way it is... honestly just irks me 😅.

Not a big deal, but could quickly be changed when you're making other changes.

bot/utils/decorators.py
```diff
@@ -199,23 +199,38 @@ def predicate(ctx: Context) -> bool:
...
+        overridden_command: Union[commands.Command, commands.Group] = None
```
Bluenix2, line 204: Huh, shouldn't the type be `Union[commands.Command, commands.Group, None]`?

---

Also by the looks of it you don't use any `commands.Group`-specific attributes and it's already a subclass of `commands.Command` so you should be able to do `Optional[commands.Command]` unless I am missing something.

Xithrius:
@onerandomusername Status?

bot/utils/decorators.py
```diff
@@ -199,23 +199,38 @@ def predicate(ctx: Context) -> bool:
...
+        if overridden_command is not None:
+            log.debug(f'Command {overridden_command} has overrides')
+            if overridden_command is not ctx.command:
+                log.debug(
+                    f"Command '{ctx.command.qualified_name}' inherited overrides "
+                    "from parent command '{overridden_command.qualified_name}'"
+                )
+
+        # Update kwargs based on override, if one exists
+        if overridden_command and hasattr(overridden_command.callback, "override"):
```
wookie184, line 218: We don't need the hasattr check here, right?
```suggestion
        if overridden_command:
```

bot/utils/decorators.py
wookie184:
```suggestion
        # Determine which command's overrides we will use. Group commands will
        # inherit from their parents if they don't define their own overrides
```

wookie184 (changes requested):

wookie184 (approved):

ChrisLovering (approved):

PR #983, opened 2021-12-18
Epoch command

### LINKED ISSUE #787: .epoch command
## Description
Discord's timestamp feature will be rolling out to all platforms soon... ish ... probably. This timestamp takes an epoch timestamp to use in a rich/dynamic updating field. More information about this can be found [here](https://discord.com/developers/docs/reference#message-formatting-timestamp-styles). I am not ~~Xith~~ a bot, therefore I don't know specific epoch times off the top of my head.

This command would take a simple datetime or a duration and then Lancebot would reply with the equivalent epoch time.

## Proposed Implementation
- Ideally this command should take a simple datetime. Something like 5/6/2021 14:00. A decision will have to be made if that's going to be American or Rest-Of-The-World format though. It will also have to make some assumptions about the seconds and microseconds and whatnot.
- It should take into account which TZ the command was invoked in and not use the bot's local time.
- It would be neat if it also took a duration! Like if I passed it 40M, it would give me the epoch time for 40 minutes from now.
- Lancebot should reply in a non-embed so mobile users can easily copy and paste the provided result.
- Obvious error checks for times before epoch and whatnot. 

Anyone can implement this feature! I am fairly busy and probably won't have time for this until December at best.

Sn4u:
## Relevant Issues

Closes #787 

## Description

Created an "epoch" command which allows you to convert a word date and time to an epoch (Unix timestamp). The parsing for relative dates is done with `arrow.dehumanize`, absolute dates are parsed with `dateutil.parser`. 

If the entered string is parsed successfully, the epoch is sent to chat along with a drop-down interaction which allows you to format the epoch as one of the discord timestamp styles
image
image

## Did you:


bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,123 @@
...
+        These are used in the description of each option in the dropdown
+        """
+        date = date.to('utc')
+        formatted = [str(int(date.timestamp()))]
+        formatted += [date.format(d_format) for d_format in TIMESTAMP_FORMATS]
+        formatted.append(date.humanize())
+        return formatted
+
+
+class TimeStampDropdown(discord.ui.Select):
```
Akarys42, line 89: timestamp is a single word, this should be renamed to `TimestampDropdown`

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,123 @@
...
+
+    async def callback(self, interaction: discord.Interaction):
+        selected = interaction.data["values"][0]
+        if selected == "Epoch":
+            return await interaction.message.edit(content=f"`{self.epoch}`")
+        else:
+            return await interaction.message.edit(content=f"`<t:{self.epoch}:{STYLES[selected]}>`")
+
+
+class TimeStampMenuView(discord.ui.View):
```
Akarys42, line 106: Same as above

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,123 @@
...
+]
+STYLES = {"Epoch": "",
+          "Short Time": "t",
+          "Long Time": "T",
+          "Short Date": "d",
+          "Long Date": "D",
+          "Short Date/Time": "f",
+          "Long Date/Time": "F",
+          "Relative Time": "R"
+          }
```
Akarys42, line 29: This is pretty order dependent. What would you say about having `STYLES` hold a tuple containing the format and the identifier?

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,123 @@
...
+
+        **Relative time**
+            accepted units: "seconds", "minutes", "hours", "days", "weeks", "months", "years"
+            Eg ".epoch in a month 4 days and 2 hours"
+
+        **Absolute time**
+            eg ".epoch 2022/6/15 16:43 -04:00"
+            Absolute times must be entered in descending orders of magnitude.
+            Timezones may take the following forms:
+                Z (UTC)
```
Akarys42, line 57: It is a bit sad that you have to explicitly mention `Z`, would there be any way around this one?

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,123 @@
...
+        formatted += [date.format(d_format) for d_format in TIMESTAMP_FORMATS]
+        formatted.append(date.humanize())
+        return formatted
+
+
+class TimeStampDropdown(discord.ui.Select):
+    def __init__(self, formatted_times: list[str], epoch: int):
+        self.epoch: int = epoch
+        super().__init__(
+            placeholder="Format this epoch as a discord timestamp",
```
Akarys42, line 93: I'd word this as "Select the format of your timestamp", but I don't mind either way

Akarys42 (changes requested):
Works pretty well, just a few things

bot/exts/utilities/epoch.py
Sn4u: Writing the timezone is optional, it will default to UTC anyway. `Z` and `UTC` should both be recognized as well, but other named timezones (PST, EST, etc.) won't be. Do you think I should remove that line from the docstring?

bot/exts/utilities/epoch.py
Sn4u: I'm not sure what you mean here. What would that look like?

Xithrius: A list of tuples would look like this:
```python
STYLES = [
	("Epoch", ""),
	("Short Time", "t"),
	("Long Time", "T"),
	...
]
```

Sn4u: > A list of tuples would look like this:
> 
> ```python
> STYLES = [
> 	("Epoch", ""),
> 	("Short Time", "t"),
> 	("Long Time", "T"),
> 	...
> ]
> ```
I don't see the benefit over a dictionary. That should retain order as well, no?

Xithrius: If my memory serves the correct information, dictionaries retain insertion order. So what you have now *should* be fine.

Akarys42: No no, I was thinking about a dictionary of tuples, something like
```py
STYLES = {
    "Short time": ("t", "h:mm A"),
    "Long time": ("T", "h:mm:ss A")
    ...
}```
To remove `TIMESTAMP_FORMAT`

File: bot/exts/utilities/epoch.py (hunk 4 above)
Comment (Akarys42) [reply bot/exts/utilities/epoch.py:69]: Oh I see, yeah you could mention in the help text that the timezone is optional and will default to UTC

File: bot/exts/utilities/epoch.py (hunk 3 above)
Comment (Xithrius) [reply bot/exts/utilities/epoch.py]: Ah, yeah. That makes more sense. I agree with this suggestion.

File: bot/exts/utilities/epoch.py [hunk 6]
```diff
@@ -0,0 +1,134 @@
...
+        """
+        if not date_time:
+            await invoke_help_command(ctx)
+            return
+
+        if isinstance(date_time, tuple):
+            ignored_tokens, date_time = list(a.strip() for a in date_time[1] if a.strip()), date_time[0]
+            if ignored_tokens:
+                await ctx.send(f"Could not parse the following token(s): `{', '.join(ignored_tokens)}`")
+                await ctx.send(f"The resulting date and time is: `{date_time.format(arrow.FORMAT_RSS)}`")
```
janine9vn, line 76: The message of "The resulting date and time is:" is actually quite helpful as part of the message and it would be nice if this was sent for every epoch invocation that used relative datetimes.

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,134 @@
...
+                ±HH
+
+        Times in the dropdown are shown in UTC
+        """
+        if not date_time:
+            await invoke_help_command(ctx)
+            return
+
+        if isinstance(date_time, tuple):
+            ignored_tokens, date_time = list(a.strip() for a in date_time[1] if a.strip()), date_time[0]
```
janine9vn, line 73: Could you explain a bit more about what this logic is doing/trying to accomplish as a comment? It's a bit non-obvious for me and I think would be helpful for others.

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,134 @@
...
+    """Convert an entered time and date to a unix timestamp."""
+
+    @commands.command(name="epoch")
+    async def epoch(self, ctx: commands.Context, *, date_time: DateString = None) -> None:
+        """
+        Convert an entered date/time string to the equivalent epoch.
+
+        **Relative time**
+            accepted units: "seconds", "minutes", "hours", "days", "weeks", "months", "years"
+            Eg ".epoch in a month 4 days and 2 hours"
```
janine9vn, line 55: A question more than anything, is it an `arrow` limitation that `.epoch 6 hours` and `.epoch in 6 hours` produce different results? In my ideal world, `.epoch 6 hours` would mean the same thing as `.epoch in 6 hours`

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,134 @@
...
+    @staticmethod
+    def _format_dates(date: arrow.Arrow) -> list[str]:
+        """
+        Return a list of date strings formatted according to the discord timestamp styles.
+
+        These are used in the description of each style in the dropdown
+        """
+        date = date.to('utc')
+        formatted = [str(int(date.timestamp()))]
+        formatted += [date.format(format[1]) for format in list(STYLES.values())[1:6]]
```
janine9vn, line 94: This should be `[1:7]`, currently the Long Date/Time is getting cut-off and Relative Time is being labeled as the Long Date/Time

janine9vn (changes requested):
Thank you for this PR!

Really solid code, I just have some minor feedback and a small bug with the dropdowns.

bot/exts/utilities/epoch.py
Sn4u: Yes, it seems a relative time can't be parsed without an `in ..` or `... ago`. I guess it's probably best to mention that in the docstring

onerandomusername: this could maybe be manually adjusted, by inserting "in" if it isn't there, and ago is not there

Shivansh-007 (commented):
Looks good, few minor comments.

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,134 @@
...
+            Absolute times must be entered in descending orders of magnitude.
+            Timezones may take the following forms:
+                Z (UTC)
+                ±HH:MM
+                ±HHMM
+                ±HH
+
+        Times in the dropdown are shown in UTC
+        """
+        if not date_time:
```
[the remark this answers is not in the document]
Sn4u: I thought it would be helpful to bring up the help embed if no date string was given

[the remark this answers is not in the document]
Shivansh-007: I agree we should show the help embed every time there is an invalid input error, like how python does, I could probably open a separate issue to mention that.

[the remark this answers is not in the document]
janine9vn: I prefer it in the current implementation, where if I just invoke the command with no datetime if provides the help command without an error embed.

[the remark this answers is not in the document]
onerandomusername: Which makes sense, sure, but it should be implemented on the error handler rather than on every individual command.

Bot already has it in the error handler, so we should make a issue/pull to migrate that feature over here.

bot/exts/utilities/epoch.py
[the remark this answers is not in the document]
Sn4u: I'm not sure what you mean here. Leave it as a tuple as opposed to converting to a list? We're working with a generator here so it has to be converted. Note that this is outdated code; I've since switched to filter+map. But even then the map object needs to be converted to a list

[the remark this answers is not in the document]
Shivansh-007: `ignored_tokens` is only used on `"".join()` which works with a generator, so we can remove the list call and keep it as
```py
            ignored_tokens, date_time = (a.strip() for a in date_time[1] if a.strip()), date_time[0]
```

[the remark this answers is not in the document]
Sn4u: I would rather keep it as a list. With a generator, the condition `if ignored_tokens:` is True, even when empty

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,134 @@
...
+            if ignored_tokens:
+                await ctx.send(f"Could not parse the following token(s): `{', '.join(ignored_tokens)}`")
+                await ctx.send(f"The resulting date and time is: `{date_time.format(arrow.FORMAT_RSS)}`")
+
+        epoch = int(date_time.timestamp())
+        dropdown = _TimestampDropdown(self._format_dates(date_time), epoch)
+        view = _TimestampMenuView(ctx, dropdown)
+        original = await ctx.send(f"`{epoch}`", view=view)
+        if await view.wait():  # wait until expiration and remove the dropdown
+            await original.edit(view=None)
```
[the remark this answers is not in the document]
Sn4u: It's possible however I think I prefer the current implementation where the message is edited, and the drop-down component is removed after the timeout. I'll leave this conversation open and maybe a few other people can share what they think

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,135 @@
...
+            date_time = date_time[0]
+            if ignored_tokens:
+                await ctx.send(f"Could not parse the following token(s): `{', '.join(ignored_tokens)}`")
+            await ctx.send(f"Date and time parsed as: `{date_time.format(arrow.FORMAT_RSS)}`")
+
+        epoch = int(date_time.timestamp())
+        view = TimestampMenuView(ctx, self._format_dates(date_time), epoch)
+        original = await ctx.send(f"`{epoch}`", view=view)
+        if await view.wait():  # wait until expiration before removing the dropdown
+            await original.edit(view=None)
```
onerandomusername, line 89: this can error if the message has somehow been deleted when we try to delete it. 

Might not worth be changing, given that there's not a wait_for_deletion or anything, so only a mod can delete it.

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,135 @@
...
+            ignored_tokens = list(map(str.strip, filter(str.strip, date_time[1])))
+            date_time = date_time[0]
+            if ignored_tokens:
+                await ctx.send(f"Could not parse the following token(s): `{', '.join(ignored_tokens)}`")
+            await ctx.send(f"Date and time parsed as: `{date_time.format(arrow.FORMAT_RSS)}`")
+
+        epoch = int(date_time.timestamp())
+        view = TimestampMenuView(ctx, self._format_dates(date_time), epoch)
+        original = await ctx.send(f"`{epoch}`", view=view)
+        if await view.wait():  # wait until expiration before removing the dropdown
```
onerandomusername, line 88: shouldn't be an if here. it should always be true, and we want to remove the view either way.

onerandomusername (changes requested):

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,135 @@
...
+        self.dropdown: discord.ui.Select = self.children[0]
+        for label, date_time in zip(STYLES.keys(), formatted_times):
+            self.dropdown.add_option(label=label, description=date_time)
+
+    @discord.ui.select(placeholder="Select the format of your timestamp")
+    async def select_format(self, _: discord.ui.Select, interaction: discord.Interaction) -> discord.Message:
+        """Drop down menu which contains a list of formats which discord timestamps can take."""
+        selected = interaction.data["values"][0]
+        if selected == "Epoch":
+            return await interaction.message.edit(content=f"`{self.epoch}`")
```
onerandomusername, line 121: This should use interaction.response.edit_message() as that acks the response *and* edits the message in one request.

interaction.message.edit is a normal edit request, over the normal route.
Currently, discord.py views check if the interaction was responded after the callback returns, and send a deferred response if it wasn't acked.This ends up being two api requests.

However, if we use interaction.response.edit_message, that will only be one api request.

onerandomusername (changes requested):
My actually most important message wasn't saved by github. Nor was my summary.

Looks pretty good, one comment about api requests \:)

bot/exts/utilities/epoch.py
Sn4u: 👍 That's good to know! Fixed in latest commit

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,135 @@
...
+            await invoke_help_command(ctx)
+            return
+
+        if isinstance(date_time, tuple):
+            # Remove empty strings. Strip extra whitespace from the remaining items
+            ignored_tokens = list(map(str.strip, filter(str.strip, date_time[1])))
+            date_time = date_time[0]
+            if ignored_tokens:
+                await ctx.send(f"Could not parse the following token(s): `{', '.join(ignored_tokens)}`")
+            await ctx.send(f"Date and time parsed as: `{date_time.format(arrow.FORMAT_RSS)}`")
```
janine9vn, line 83: fun fact about this, if it's a relative/humanized date_time passed in (i.e. "in 30 minutes") then the `await ctx.send(f"Date and time parsed as...")` won't send, as the type of that is `arrow.arrow.Arrow` and not tuple. 

If that `ctx.send(...)` was un-indented one more level then it would always fire, which I would like if possible. But I also don't think it *needs* to be changed, so feel free to ignore.

bot/exts/utilities/epoch.py
```diff
@@ -0,0 +1,135 @@
...
+        """Drop down menu which contains a list of formats which discord timestamps can take."""
+        selected = interaction.data["values"][0]
+        if selected == "Epoch":
+            return await interaction.response.edit_message(content=f"`{self.epoch}`")
+        return await interaction.response.edit_message(content=f"`<t:{self.epoch}:{STYLES[selected][0]}>`")
+
+    async def interaction_check(self, interaction: discord.Interaction) -> bool:
+        """Check to ensure that the interacting user is the user who invoked the command."""
+        if interaction.user != self.ctx.author:
+            embed = discord.Embed(description="Sorry, but this interaction can only be used by the original author.")
```
janine9vn, line 127: Similar comment I left on one of D0rs4n's PR, but I'm concerned about someone not familiar with d.py knowing what the term "interaction" necessarily means. Maybe something like "Sorry, this response can only be used by ...") but I'm also cool if you think it's fine and want to keep it as is.

janine9vn (approved):
\*chef's kiss\* Thank you for implementing this! Your local events lead who is always in need of converting dates to epoch is very appreciative.

I have 2 small comments but they are by no means blockers.

bot/exts/utilities/epoch.py
Sn4u: https://github.com/python-discord/sir-lancebot/pull/983#discussion_r773506933
This was a misunderstanding on my part. When I read this comment from earlier, you said that you thought it would be useful if that line was sent for every invocation that used relative datetimes. I incorrectly assumed that you made a mistake and you meant that you wanted it for absolute datetimes. I'll commit that now

HassanAbouelela (approved):
This is really good. Top-notch stuff, thanks

PR #988, opened 2021-12-22
Add Twemoji utility command

### LINKED ISSUE #987: Twemoji utility command
This issue proposes the addition of a `.twemoji <codepoint|emoji>` command. This will be useful for viewing and downloading Twemojis (the default emojis on Discord, which we base a lot of our branding on).

### Invocation
An invocation of the command could look something like this. These are all common formats for specifying an emoji. However, I might limit the available formats, in order to make things simpler.
```
.twemoji 😁
.twemoji \U0001f9d0
.twemoji 1f9d0
.twemoji U+1f9d0
```
This would work with multi-character emojis as well:
```
.twemoji ☺️
.twemoji \u263a\ufe0f
.twemoji U+1F46A U+1F3FF
```
### Bot response
The bot will respond with an embed containing the following information about the given Twemoji:

- Emoji name (if I can find a simple way to fetch it)
- Download link to the source file (as an svg)
- A preview of the emoji
- The emoji codepoint

Including the codepoint will be occasionally useful when the user specifies the emoji of interest through an emoji (eg. "🥰" instead of "1f970"). The color of the embed will be blue, to match Twitter's color scheme, since Twitter are the creators of Twemojis.

## Additional Details

It's possible to add support for specifying multiple emojis in the same call to the command. The emojis could then be separated using pagination.

However, considering the niche use cases for this feature, I think it's better to keep the implementation simple. If we against all odds find a significant need for specifying multiple emojis, in the future, I'm open to discussing the possibility of adding pagination at that point in time.

gustavwilliam:
## Relevant Issues
Closes #987.

## Description
Please refer to #987 for a more detailed explanation of the command.

### Example usage
The following screenshots showcase different formats that can be used to indicate what twemoji should be displayed.

Screenshot 2021-12-22 at 23 57 53
Screenshot 2021-12-22 at 23 58 29
Screenshot 2021-12-22 at 23 59 06

## Did you:


bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,144 @@
+import logging
+import re
+from typing import List, Literal
```
TizzySaurus, line 3: As mentioned on discord, since a spring-cleanup-like PR (I forget the number), we use `list[]` etc. in place of the typing variant.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,144 @@
...
+        argument can be of any format, since it will be trimmed automatically.
+        """
+        if code := Twemoji.trim_code(codepoint):
+            return chr(int(code, 16))
+        return ""
+
+    @staticmethod
+    def codepoint(emoji: str) -> str:
+        """
+        Returns the codepoint of a single emoji.
```
TizzySaurus, line 74: Since there seems to be multiple "codepoint formats" (based on `trim_code`) we should specify which format this returns.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,144 @@
...
+            description=f"{codepoint.replace('-', ' ')}\n[Download svg]({Twemoji.get_url(codepoint, 'svg')})",
+            colour=Colours.twitter_blue,
+        )
+        embed.set_thumbnail(url=Twemoji.get_url(codepoint, "png"))
+        return embed
+
+    @staticmethod
+    def emoji(codepoint: str) -> str:
+        """
+        Returns the emoji corresponding to a given codepoint.
```
TizzySaurus, line 62: NIT: I'd be tempted to reword this to "Returns the emoji corresponding to the `codepoint`." Very minor thing so can leave if you want to.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,144 @@
...
+
+        Example usages:
+        >>> trim_code("U+1f1f8")
+        "1f1f8"
+        >>> trim_code("\u0001f1f8")
+        "1f1f8"
+        >>> trim_code("1f466")
+        "1f466"
+        """
+        if code := CODE.search(str(codepoint)):
```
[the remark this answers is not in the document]
gustavwilliam: It is, and the return value is also a string. See the examples in the docstring.

Not sure if I'm understanding your question correctly?

bot/exts/utilities/twemoji.py
gustavwilliam: This always returns the most trimmed down format. I could, however, specify that more clearly.

bot/exts/utilities/twemoji.py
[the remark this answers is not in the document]
Shivansh-007: He is basically saying why do we need to call `str()` on `codepoint` if the parameter `codepoint` is already a string.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,144 @@
...
+    async def twemoji(self, ctx: commands.Context, *raw_emoji: str) -> None:
+        """Sends a preview of a given Twemoji, specified by codepoint or emoji."""
+        if len(raw_emoji) == 0:
+            await invoke_help_command(ctx)
+            return
+        try:
+            codepoint = self.codepoint_from_input(raw_emoji)
+        except (ValueError, AttributeError):
+            raise commands.BadArgument(
+                "please include a valid emoji or emoji code point."
```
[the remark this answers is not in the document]
gustavwilliam: I'll make it consistent. Autocorrect seems to prefer "code point", but I'm more of a fan of "codepoint"

bot/exts/utilities/twemoji.py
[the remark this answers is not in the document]
TizzySaurus: > 

This ^ @gustavwilliam. If it's already a string then why are we converting it to a string again?

bot/exts/utilities/twemoji.py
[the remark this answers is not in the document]
TizzySaurus: Right, I see. Keep the error lowercase then. I too prefer `codepoint` over `code point` by the way so would keep it one word.

bot/exts/utilities/twemoji.py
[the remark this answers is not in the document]
gustavwilliam: Ah, good catch! I'll investigate.

onerandomusername (changes requested):
Needs an error message.
image

No code review.

*This may not be accurate as I had a hard time checking out this branch, for some reason.

gustavwilliam:
> Needs an error message.

@onerandomusername Not sure how that's possible. I'm pretty sure there's an error embed for when no emoji is found.

I'll double check this.

gustavwilliam:
Seems like there was a bug with the implementation.

Should be good to go now, though.
Screenshot 2021-12-29 at 01 39 17

TizzySaurus (approved):
Haven't tested, but the code looks good :+1:

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,148 @@
...
+
+        `emoji` should be an emoji character, such as "🐍" and "🥰", and
+        not a codepoint like "1f1f8". When working with combined emojis,
+        such as "🇸🇪" and "👨‍👩‍👦", send the component emojis through the method
+        one at a time.
+        """
+        return hex(ord(emoji))[2:]
+
+    @staticmethod
+    def trim_code(codepoint: str | None) -> str:
```
Bluenix2, line 84: You can't use this syntax yet unless you add `from __future__ import annotations` at the top of the file.

Bluenix2, line 84: The return type isn't correct here according to your docstring, it should be `Optional[str]` or `str | None`

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,148 @@
...
+    def emoji(codepoint: str) -> str:
+        """
+        Returns the emoji corresponding to a given `codepoint`.
+
+        The return value is an emoji character, such as "🍂". The `codepoint`
+        argument can be of any format, since it will be trimmed automatically.
+        """
+        if code := Twemoji.trim_code(codepoint):
+            return chr(int(code, 16))
+        return ""
```
Bluenix2, line 69: This can be a ternary operation like this:
```suggestion
        return chr(int(code, 16)) if (code := Twemoji.trim_code(codepoint)) else ""
```

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,148 @@
...
+        "1f1f8"
+        >>> trim_code("1f466")
+        "1f466"
+        """
+        if not codepoint:
+            return None
+        if code := CODE.search(codepoint):
+            return code.group()
+
+    def codepoint_from_input(self, raw_emoji: tuple[str, ...]) -> str:
```
Bluenix2, line 103: This could be a staticmethod like the other ones you made right?

Bluenix2, line 103: Shouldn't this be `list[str]` according to the examples you have showed?

You can use `Iterable[str]` for both to work.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @commands.command(aliases=("tw",))
+    async def twemoji(self, ctx: commands.Context, *raw_emoji: str) -> None:
+        """Sends a preview of a given Twemoji, specified by codepoint or emoji."""
+        if len(raw_emoji) == 0:
+            await invoke_help_command(ctx)
+            return
+        try:
+            codepoint = self.codepoint_from_input(raw_emoji)
+        except (ValueError, AttributeError):
```
Bluenix2, line 138: When do you get an AttributeError? It may be worth to it to attempt to fix what's causing it (unless it's supposed to raise AttributeError but I am failing to see where that would be).

Bluenix2 (commented):
Oops forgot to press "Submit review" on this one

bot/exts/utilities/twemoji.py
gustavwilliam: Good catch. I was writing this in 3.10.

gustavwilliam: The docstring might be *the culprit* here, since this always returns a string.

I'll look into some options for fixing this.

bot/exts/utilities/twemoji.py
gustavwilliam: Do you think this is better? It looks more compact, but I'm not sure the readability is quite as good.

Either way, I'll look into changing the return type to `str | None`, so this might disappear.

bot/exts/utilities/twemoji.py
onerandomusername: > You can't use this syntax yet unless you add `from __future__ import annotations` at the top of the file.

Even with the import, it would break if anything tried typing.get_type_hints as the syntax itself was added in 3.10.

bot/exts/utilities/twemoji.py
gustavwilliam: I'll update the docstring.

bot/exts/utilities/twemoji.py
gustavwilliam: What would the response be?

bot/exts/utilities/twemoji.py
gustavwilliam: I don't quite remember, but I'll investigate. It might've been a past thing.

bot/exts/utilities/twemoji.py
onerandomusername:
```py
Python 3.9.9 (main, Nov 23 2021, 11:52:34)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from __future__ import annotations
>>> import typing
>>> def meth(foo: str | None):
...   pass
...
>>> typing.get_type_hints(meth)
  File "<stdin>", line 1, in <module>
  File "/home/.../.pyenv/versions/3.9.9/lib/python3.9/typing.py", line 1496, in get_type_hints
    value = _eval_type(value, globalns, localns)
  File "/home/.../.pyenv/versions/3.9.9/lib/python3.9/typing.py", line 292, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
  File "/home/.../.pyenv/versions/3.9.9/lib/python3.9/typing.py", line 554, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

gustavwilliam: Interesting. I might use `Optional[]` for now then 👍🏼

bot/exts/utilities/twemoji.py
gustavwilliam: I'm rejecting this for the sake or readability. Unresolved this if you have a strong opinion regarding this.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,147 @@
...
+            return "-".join(emojis)
+
+        emoji = "".join(Twemoji.emoji(Twemoji.trim_code(code)) for code in raw_emoji)
+        if is_emoji(emoji):
+            return "-".join(Twemoji.codepoint(e) for e in emoji)
+
+        raise ValueError("No codepoint could be obtained from the given input")
+
+    @commands.command(aliases=("tw",))
+    async def twemoji(self, ctx: commands.Context, *raw_emoji: str) -> None:
```
Bluenix2, line 130: You need to add the correct decorators - otherwise we can't use this in dev-branding for example.

Bluenix2 (changes requested):

bot/exts/utilities/twemoji.py
gustavwilliam: Ah, decorators to allow it by default in all channels? Which ones do we use for that?

Bluenix2: It seems like [`whitelist_check`](https://github.com/python-discord/sir-lancebot/blob/main/bot/utils/decorators.py#L191) is what you're looking for.

ChrisLovering: [`whitelist_override()`](https://github.com/python-discord/sir-lancebot/blob/main/bot/utils/decorators.py#L286-L304) is the decorator that we use for this.

HassanAbouelela: Whitelist override (the one below that) is what we actually use. On second thought, it might be best to hide the other function, but that’s a problem for later.

Because it can be confusing, what you need to do is set the role override to the everyone role. You can see an example of that here:

https://github.com/python-discord/sir-lancebot/blob/c7b93780e509e05121ce1c3deacc38c6f884a6c9/bot/exts/utilities/bookmark.py#L90

gustavwilliam: Thanks!

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,149 @@
...
+from bot.constants import Colours, Roles
+from bot.utils.decorators import whitelist_override
+from bot.utils.extensions import invoke_help_command
+
+log = logging.getLogger(__name__)
+BASE_URLS = {
+    "png": "https://raw.githubusercontent.com/twitter/twemoji/master/assets/72x72/",
+    "svg": "https://raw.githubusercontent.com/twitter/twemoji/master/assets/svg/",
+}
+CODE = re.compile(r"[a-f1-9][a-f0-9]{3,5}$")
```
ShakyaMajumdar, line 19: IMO `CODE` feels a bit vague as a variable name.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,149 @@
...
+    def codepoint(emoji: str) -> str:
+        """
+        Returns the codepoint, in a trimmed format, of a single emoji.
+
+        `emoji` should be an emoji character, such as "🐍" and "🥰", and
+        not a codepoint like "1f1f8". When working with combined emojis,
+        such as "🇸🇪" and "👨‍👩‍👦", send the component emojis through the method
+        one at a time.
+        """
+        return hex(ord(emoji))[2:]
```
ShakyaMajumdar, line 82:
```suggestion
        return hex(ord(emoji)).removeprefix("0x")
```

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @staticmethod
+    def trim_code(codepoint: Optional[str]) -> Optional[str]:
```
ShakyaMajumdar, line 85: `codepoint` can be just `str`, right? I don't see anywhere the method can be called with a `None`/`Optional`. Even if there was, it would make more sense to handle the `None` at the call-site, not in the function.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,149 @@
...
+        Example usages:
+        >>> trim_code("U+1f1f8")
+        "1f1f8"
+        >>> trim_code("\u0001f1f8")
+        "1f1f8"
+        >>> trim_code("1f466")
+        "1f466"
+        """
+        if not codepoint:
+            return None
```
ShakyaMajumdar, line 100: Would become redundant if the previous suggestion is incorporated.

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Transform a unicode alias to an emoji name.
+
+        Example usages:
+        >>> alias_to_name(":falling_leaf:")
+        "Falling leaf"
+        >>> alias_to_name(":family_man_girl_boy:")
+        "Family man girl boy"
+        """
+        name = alias[1:-1].replace("_", " ")
```
ShakyaMajumdar, line 44:
```suggestion
        name = alias.strip(":").replace("_", " ")
```

bot/exts/utilities/twemoji.py
```diff
@@ -0,0 +1,149 @@
...
+    def emoji(codepoint: str) -> str:
+        """
+        Returns the emoji corresponding to a given `codepoint`, or `""` if no emoji was found.
+
+        The return value is an emoji character, such as "🍂". The `codepoint`
+        argument can be of any format, since it will be trimmed automatically.
+        """
+        if code := Twemoji.trim_code(codepoint):
+            return chr(int(code, 16))
+        return ""
```
ShakyaMajumdar, line 70: I don't like the idea of returning an empty string to indicate failure. Maybe `raise ValueError` or `return None` instead?

ShakyaMajumdar (changes requested):
Also as a general comment, the `staticmethod`s might be better suited as top-level functions, I think.

bot/exts/utilities/twemoji.py
gustavwilliam: In what way is this better? Clarity of what's going on?

bot/exts/utilities/twemoji.py
gustavwilliam: Is this also for clarity? The actual result won't change.

bot/exts/utilities/twemoji.py
gustavwilliam: Sure, raising an exception would probably be optimal when only looking at this function, but the output of this function is used in a place where returning an empty string makes things a lot simpler.

Considering this is Sir-Lancebot, I think it's fine to leave it like this, instead of making any larger changes to the code.

gustavwilliam:
> Also as a general comment, the `staticmethod`s might be better suited as top-level functions, I think.

@ShakyaMajumdar True. In an ideal world, I'd put a few of the `@staticmethod`s in a `bot/utils/emojis.py` file. Might change that.

bot/exts/utilities/twemoji.py
ShakyaMajumdar: Yeah, I feel these strings methods are more explicit than slicing.

bot/exts/utilities/twemoji.py
ShakyaMajumdar: If you make it return `None`, it wouldn't need a lot of change - just `Twemoji.emoji(e)` to `Twemoji.emoji(e) or ""`.

gustavwilliam: True!

bot/exts/utilities/twemoji.py
codephile1221: Yeah, `CODEPOINT`/`CODEPOINT_RE`/ `CODEPOINT_REGEX` would be clearer.

Bluenix2:
Is the command meant to support unicode characters?

gustavwilliam:
> Is the command meant to support unicode characters?

Yes, such as the first example in the PR description.

Bluenix2 (approved):
I must've forgotten to Approve.

Regarding my last question, for reference, I was doing `\:sweat_smile:` which becomes `\😅` in the message content (displayed as `😅` instead of the Twemoji representation). That was the only blocker for my approval.

Nice work here, thanks!

ShakyaMajumdar (approved):

Akarys42 (approved):

Xithrius (approved):

PR #1008, opened 2022-01-09
Fix aoc join in date logic

ChrisLovering:
## Description

Not using `.value` meant it was pulling the enum name, and therefore was failing the check.

## Did you:

### DIFF
--- a/bot/exts/events/advent_of_code/_cog.py
+++ b/bot/exts/events/advent_of_code/_cog.py
@@ -187,9 +187,10 @@ class AdventOfCode(commands.Cog):
     async def join_leaderboard(self, ctx: commands.Context) -> None:
         """DM the user the information for joining the Python Discord leaderboard."""
         current_date = datetime.now()
-        if (
-            current_date.month not in (Month.NOVEMBER, Month.DECEMBER) and current_date.year != AocConfig.year or
-            current_date.month != Month.JANUARY and current_date.year != AocConfig.year + 1
+        allowed_months = (Month.NOVEMBER.value, Month.DECEMBER.value)
+        if not (
+            current_date.month in allowed_months and current_date.year == AocConfig.year or
+            current_date.month == Month.JANUARY.value and current_date.year == AocConfig.year + 1
         ):
             # Only allow joining the leaderboard in the run up to AOC and the January following.
             await ctx.send(f"The Python Discord leaderboard for {current_date.year} is not yet available!")


ToxicKidz (approved):
LGTM

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -187,9 +187,10 @@ async def about_aoc(self, ctx: commands.Context) -> None:
     async def join_leaderboard(self, ctx: commands.Context) -> None:
         """DM the user the information for joining the Python Discord leaderboard."""
         current_date = datetime.now()
+        allowed_months = (Month.NOVEMBER.value, Month.DECEMBER.value)
         if (
```
D0rs4n, line 191:
```suggestion
        if not (
```
Since, the first batch of condition will result in False, and the second one as well, since it's January and `2021+1`

D0rs4n (changes requested):
One tiny detail.

bot/exts/events/advent_of_code/_cog.py
ToxicKidz: Why should the bot say the leaderboard isn't available if it's the right month?

ChrisLovering: I'm not sure I follow, this `if` block is to determine if it's not current AoC, if it's not, then we send the error and return.

Current AoC is defined by the AOC_YEAR var, which we set in the env. So currently, it's set to 2021.

Both parts would fail right now, so the user invoking can join the 2021 board. Come Feb, neither of the conditionals will pass, so the error comes up and the command returns early.

D0rs4n: I tested it out locally, and it did not work.
Must have overlooked something.

D0rs4n (approved):

bot/exts/events/advent_of_code/_cog.py
ChrisLovering: Did you set the AOC_YEAR env var to 2021 locally?

D0rs4n: Yes, of course.
Event tested out with 2022 just to make sure it won't work. (And it will with 2021)

ChrisLovering: No wait, you are right, sorry. i need to look into this more.

ChrisLovering: Alright, pushed a fix.

D0rs4n: Oh, alright.
The logic checked out but something was off 👍

ChrisLovering: Yea, too much negation, ands and ors combined. This should read a little easier now.

PR #1010, opened 2022-01-10
Latex Command

### LINKED ISSUE #1009: Latex Command Revamped
## Description
A slightly different implementation of the .latex command, hopefully with no memory leaks this time.
see also: #663 #668

## Proposed Implementation
Previously, we were using matplotlib to render the latex, which seemed to hold references to the data even after we were done with it if the input contained invalid latex. This led to an increase in memory usage every time there was malformed input, and ultimately the bot would come crashing down because of memory limits on the process. The only way to mitigate it was with manual calls to `gc.collect` and `del`s, which was error-prone and hacky.

This time around, I propose that we keep a separate python script for the latex logic, and call into that with `asyncio.subprocess`. That way, memory usage should return to normal after every invocation, since a new process is created each time. Each process would load matplotlib, render latex, save the output in a cache directory, and die.

ShakyaMajumdar:
## Relevant Issues
Closes #1009 

## Description
(See proposed implementation in #1009)

## Did you:


bot/exts/fun/latex/_latex_cog.py
```diff
@@ -0,0 +1,73 @@
...
+class Latex(commands.Cog):
+    """Renders latex."""
+
+    @commands.command()
+    @commands.max_concurrency(1, commands.BucketType.guild, wait=True)
+    async def latex(self, ctx: commands.Context, *, query: str) -> None:
+        """Renders the text in latex and sends the image."""
+        query = _prepare_input(query)
+        query_hash = hashlib.md5(query.encode()).hexdigest()
+        image_path = CACHE_DIRECTORY / f"{query_hash}.png"
```
rf20008, line 56: I think you meant to do ```py
image_path = f"CACHE_DIRECTORY/{query_hash}.png"```. Right now, you're dividing 2 strings by each other, which will cause an exception.

rf20008 (changes requested):
I like the code!

I have a few suggestions.

Firstly, you could delete the file immediately after it gets rendered for memory purposes or we would run into a similar problem as before.

Secondly, on line 56 of bot/exts/fun/latex/_latex_cog.py, it seems that you are dividing CACHE_DIRECTORY by `{query_hash}.png`. This will raise an exception.


Other than that, I like the code so far.

bot/exts/fun/latex/_latex_cog.py
onerandomusername: As per https://github.com/python-discord/sir-lancebot/blob/89ebfc684d257286bc7cfcb9023658952314cd3b/bot/exts/fun/latex/_latex_cog.py#L21-L23

`CACHE_DIRECTORY` is a [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html#pathlib.Path) instance, which supports the dividing operator.

I could not find the documentation for this, but the source for this is here: https://github.com/python/cpython/blob/2e6798f35260ff90129861ef1f289ac40c0396c2/Lib/pathlib.py#L851-L861

dawnofmidnight: If you're curious, the docs are the first entry under [the operators heading](https://docs.python.org/3/library/pathlib.html#operators),

rf20008: Thanks for the info!

rf20008 (approved):
Thanks for the info! I'm excited to see this merged :) (I'll find this useful)

wookie184 (dismissed):
This doesn't currently seem to enforce any memory limits on the subprocesses created, which means the command is exploitable by running a command like `.latex $\genfrac{}{}{10000000}{1}{1}{1}$` (tailoring that number to allocate as much memory as is available).

This can probably be done using the resources module, https://docs.python.org/3/library/resource.html, and setting some limits before running the render in `_renderer.py`.

`resources`, however, does not support windows. Because of this, I would suggest we disable this cog on windows. I think setting such limits is more complicated on windows and this code already doesn't run on windows as `create_subprocess_exec` doesn't work with the event loop we are using on windows https://docs.python.org/3/library/asyncio-subprocess.html#subprocess-and-threads
> On Windows subprocesses are provided by ProactorEventLoop only (default), SelectorEventLoop has no subprocess support.
We seem to use the selector event loop
https://github.com/python-discord/sir-lancebot/blob/9202b432ccef1c13e476afd892fbd718397afa0c/bot/__init__.py#L41-L42

HassanAbouelela:
@wookie184 the exact change you describe was implemented by me on the [latex-limits](https://github.com/python-discord/sir-lancebot/commits/latex-limits) branch, but it proved unreliable. The code would still often OOM completely for reasons beyond me. IMO, the desire for a latex command is mostly artificial, and we can get by fine without it. If we were to implement it, I think the path of least resistance would be using an existing framework. @vcokltfre expressed interest in writing an API like that, outside of PyDis.

wookie184:
Just for reference, i'll post some findings on the cause of the surprisingly high memory usage here.

To limit memory usage we set the rlimit [RLIMIT_AS](https://docs.python.org/3/library/resource.html#resource.RLIMIT_RSS). This corresponds to "virtual memory", which includes memory in in ram or swap space, but also memory that isn't actually used (i.e. allocated to the process but not actually allocated to any hardware). I believe this "not actually allocated" memory is what numpy seems to be taking up a lot of on import and why the rlimit needs to be set so high even though there is no visible filling up of ram or swap.

A potential solution would be to use the [RLIMIT_RSS](https://docs.python.org/3/library/resource.html#resource.RLIMIT_RSS) rlimit instead, which only includes memory actually allocated in RAM, however swap space is a thing, and there doesn't seem to a a specific rlimit for it on linux.

If we *wanted* to continue with this route, we'd probably have to set up some cgroups limits, but that's pretty out of scope for sir-lancebot :) so this will probably end up implemented through an external API

Akarys42 (approved):

jb3 (changes requested):
Adding explicit changes request because I'm in agreement that without effective memory management we should try offload this to an API which reduces attack vectors on our infrastructure.

.gitignore
```diff
@@ -1,7 +1,7 @@
 # bot (project-specific)
 log/*
 data/*
-_latex_cache/*
+bot/exts/fun/latex/cache/*
```
ChrisLovering, line 4: Changing this will require us to alter the deployment manifest too, as by default sir-lance has a read-only file system. We have specifically given a writable volume to _latex_cache/

bot/exts/fun/latex/template.txt
```diff
@@ -0,0 +1,5 @@
+\documentclass{article}
+\begin{document}
+    \pagenumbering{gobble}
+    $text
+\end{document}
```
ChrisLovering, line 5: This file should be in `bot/resources/fun`.

bot/exts/fun/latex/_latex_cog.py
```diff
@@ -0,0 +1,103 @@
...
+
+FORMATTED_CODE_REGEX = re.compile(
+    r"(?P<delim>(?P<block>```)|``?)"        # code delimiter: 1-3 backticks; (?P=block) only matches if it's a block
+    r"(?(block)(?:(?P<lang>[a-z]+)\n)?)"    # if we're in a block, match optional language (only letters plus newline)
+    r"(?:[ \t]*\n)*"                        # any blank (empty or tabs/spaces only) lines before the code
+    r"(?P<code>.*?)"                        # extract all code inside the markup
+    r"\s*"                                  # any more whitespace before the end of the code markup
+    r"(?P=delim)",                          # match the exact same delimiter from the start again
+    re.DOTALL | re.IGNORECASE,              # "." also matches newlines, case insensitive
+)
```
ChrisLovering, line 20: Ideally we would use the regex from bot-core for this.
However, it hasn't been installed on sir-lance yet. So if you don't want to implement that as part of this PR it's fine.

bot/exts/fun/latex/__init__.py
```diff
@@ -0,0 +1,7 @@
+from bot.bot import Bot
+from bot.exts.fun.latex._latex_cog import Latex
+
+
+def setup(bot: Bot) -> None:
+    """Load the Latex Cog."""
+    bot.add_cog(Latex(bot))
```
ChrisLovering, line 7: Do we still need to separate this cog out like this? Seems like it with it being a single file cog now, there is little point it being in it's own folder.

ChrisLovering (changes requested):
Purely a code-review at this point, haven't tested it.

.gitignore
ChrisLovering: This is more of a note for my reference. I think if we want to change to this new path, we should have it be _cache though, so that the walk_extensions func doesn't even attempt to import it as a module

bot/exts/fun/latex/__init__.py
ShakyaMajumdar: Should the cache directory by bot/exts/fun/cache then?

ChrisLovering: Maybe something like `bot/exts/fun/_latex_cache` to make it clear.

.gitignore
ChrisLovering: Updated deployment here https://github.com/python-discord/kubernetes/pull/118

ChrisLovering (changes requested):
Testing this locally it works very well, and is good to go.

However, the output from the API really isn't designed for Discord dark mode!

Darkmode:
image

Lightmode:
image

Could we use PIL (or similar) to put a white background on the image returned by the API? I think this will make it much easier to use. Preferably this is done in a non-blocking way.

ShakyaMajumdar:
@ChrisLovering that's weird - it looks a lot better for me
image

But yeah, I'll look into adding a background

ChrisLovering:
> @ChrisLovering that's weird - it looks a lot better for me image
> 
> But yeah, I'll look into adding a background

Yea, that example looks fine for me too, guess it's dependent on what it's outputing.
image

ChrisLovering (approved):
Looks good to me 👌

bot/exts/fun/latex.py
```diff
@@ -0,0 +1,120 @@
...
+THIS_DIR = Path(__file__).parent
+CACHE_DIRECTORY = THIS_DIR / "_latex_cache"
+CACHE_DIRECTORY.mkdir(exist_ok=True)
+TEMPLATE = string.Template(Path("bot/resources/fun/latex_template.txt").read_text())
+
+BG_COLOR = (54, 57, 63, 255)
+PAD = 10
+
+
+def _prepare_input(text: str) -> str:
```
mbaruh, line 36: Here and also in `_process_image`, even though it's not required by the linter, it would be nice if there was a short description of what each function does.

bot/exts/fun/latex.py
```diff
@@ -0,0 +1,120 @@
...
+        except Exception:
+            # 400 (Bad Request) means there are too many characters
+            pass
+
+    @commands.command()
+    @commands.max_concurrency(1, commands.BucketType.guild, wait=True)
+    async def latex(self, ctx: commands.Context, *, query: str) -> None:
+        """Renders the text in latex and sends the image."""
+        query = _prepare_input(query)
+        query_hash = hashlib.md5(query.encode()).hexdigest()
```
mbaruh, line 98: A comment about the caching mechanism would be nice here.

bot/exts/fun/latex.py
```diff
@@ -0,0 +1,120 @@
...
+CACHE_DIRECTORY.mkdir(exist_ok=True)
+TEMPLATE = string.Template(Path("bot/resources/fun/latex_template.txt").read_text())
+
+BG_COLOR = (54, 57, 63, 255)
+PAD = 10
+
+
+def _prepare_input(text: str) -> str:
+    if match := FORMATTED_CODE_REGEX.match(text):
+        return match.group("code")
```
mbaruh, line 38: Can you clarify why this is needed?

bot/exts/fun/latex.py
```diff
@@ -0,0 +1,120 @@
...
+    else:
+        return text
+
+
+def _process_image(data: bytes, out_file: BinaryIO) -> None:
+    image = Image.open(BytesIO(data)).convert("RGBA")
+    width, height = image.size
+    background = Image.new("RGBA", (width + 2 * PAD, height + 2 * PAD), "WHITE")
+    background.paste(image, (PAD, PAD), image)
```
mbaruh, line 47: Can you add a comment on why using the image as a mask is needed here?

mbaruh (changes requested):
Code looks good, I don't have much to say :D
Could use a bit more documentation however

mbaruh (approved):
Looks good, thank you :)

PR #1011, opened 2022-01-12
Fix AoC name lookup for anon users

ChrisLovering:
## Description

When a user doesn't set a name, the AoC API doesn't return a name key at all, so we need to make use of the ID field instead, to build the name based on a similar tempalte that AoC uses for it's leaderboard.
## Did you:

### DIFF
--- a/bot/exts/events/advent_of_code/_cog.py
+++ b/bot/exts/events/advent_of_code/_cog.py
@@ -96,7 +96,9 @@ class AdventOfCode(commands.Cog):
                 # Only give the role to people who have completed all 50 stars
                 continue
 
-            member_id = aoc_name_to_member_id.get(member_aoc_info["name"], None)
+            aoc_name = member_aoc_info["name"] or f"Anonymous #{member_aoc_info['id']}"
+
+            member_id = aoc_name_to_member_id.get(aoc_name)
             if not member_id:
                 log.debug(f"Could not find member_id for {member_aoc_info['name']}, not giving role.")
                 continue


bot/exts/events/advent_of_code/_cog.py
```diff
@@ -96,7 +96,9 @@ async def completionist_task(self) -> None:
                 # Only give the role to people who have completed all 50 stars
                 continue
 
-            member_id = aoc_name_to_member_id.get(member_aoc_info["name"], None)
+            aoc_name = member_aoc_info["name"] if member_aoc_info["name"] else f"Anonymous #{member_aoc_info['id']}"
```
ToxicKidz, line 99: This is more of a nitpick, but this can be simplified.
```suggestion
            aoc_name = member_aoc_info["name"] or f"Anonymous #{member_aoc_info['id']}"
```

ToxicKidz (approved):
LGTM, just one small suggestion.

brad90four (approved):
LGTM

bot/exts/events/advent_of_code/_cog.py
```diff
@@ -96,7 +96,9 @@ async def completionist_task(self) -> None:
                 # Only give the role to people who have completed all 50 stars
                 continue
 
-            member_id = aoc_name_to_member_id.get(member_aoc_info["name"], None)
+            aoc_name = member_aoc_info["name"] or f"Anonymous #{member_aoc_info['id']}"
```
Bluenix2, line 99:
```suggestion
            aoc_name = member_aoc_info.get("name") or f"Anonymous #{member_aoc_info['id']}"
```

Why isn't the name missing from `member_aoc_info` or was that an incorrect assumption?

ChrisLovering: I should have been more specific, the key is present, but is empty.

D0rs4n (approved):
Looks good! Seems to be working locally just fine.

PR #1036, opened 2022-02-20
Handle Missing Logs in Latex API Response

### LINKED ISSUE #1035: KeyError: 'log'
Sentry Issue: [SIR-LANCEBOT-7Y](https://sentry.io/organizations/python-discord/issues/3030389077/?referrer=github_integration)

```
KeyError: 'log'
  File "discord/ext/commands/core.py", line 167, in wrapped
  File "bot/exts/fun/latex.py", line 111, in latex
    await self._generate_image(TEMPLATE.substitute(text=query), out_file)
  File "bot/exts/fun/latex.py", line 76, in _generate_image
    raise InvalidLatexError(logs=response_json["log"])

Unhandled command error: 'log'
```
This was generated by running `.latex \🍏`

The API responded with
```json
{
    description: 'The server broke. This is bad.', 
    status: 'error'
}
```

ShakyaMajumdar:
## Relevant Issues
Closes #1035 

## Description
Let the `logs` attribute of invalid latex errors be `None` if its missing in the API response, and handle the case with a slightly different embed.

## Did you:


bot/exts/fun/latex.py
```diff
@@ -110,12 +110,15 @@ async def latex(self, ctx: commands.Context, *, query: str) -> None:
...
-                        embed.description = f"[View Logs]({logs_paste_url})"
+                    if err.logs is None:
+                        embed.description = "No logs available."
                     else:
-                        embed.description = "Couldn't upload logs."
+                        logs_paste_url = await self._upload_to_pastebin(err.logs)
+                        if logs_paste_url:
+                            embed.description = f"[View Logs]({logs_paste_url})"
+                        else:
+                            embed.description = "Couldn't upload logs."
```
Akarys42, line 121: I think this should raise a warning

ChrisLovering: I disagree. If the logs had too many characters and couldn't be uploaded to pastebin, then the core dev team don't need to be told by raising a warning and a sentry issue.

If it fails for another reason, then sure. At the moment however there is no way to tell.

I think we should lower this to an info at least. Maybe in a subsequent PR we could try to truncate logs before uploading to the paste service, this would mean an error being raised would be something that the core dev team would care about.

Akarys42: Right. I didn't realize it would also return None if the text is too long. 

Although, I'd argue that there should be a character limit that is being checked by the log, and we should either have another description message or truncate the logs.

ChrisLovering: Yea, I agree on the last point, and is what I mentioned in the last paragraph of my last comment.

For now I'd say revert the log change and then we can raise an issue to do the truncation.

MaskDuck (approved):
lgtm

ChrisLovering (approved):

Bluenix2 (approved):
This looks good and handles the issue cleanly in the code, thanks for the fix!

Xithrius (approved):

PR #1039, opened 2022-02-20
Add run bot step during lint

ChrisLovering:
## Description

This adds a step to our linting action to catch errors in imports/cog setup functions before they are merged by running the bot package, but not actually calling `bot.run()`.

Adding it to the lint flow means we already have the code checked out and a poetry env, so it adds ~1s to CI time.

See here fir a passing action https://github.com/ChrisLovering/sir-lancebot/runs/5264386120?check_suite_focus=true
This is a failing one https://github.com/ChrisLovering/sir-lancebot/runs/5264408507?check_suite_focus=true

## Did you:

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -143,6 +143,7 @@ class Client(NamedTuple):
     prefix = environ.get("PREFIX", ".")
     token = environ.get("BOT_TOKEN")
     debug = environ.get("BOT_DEBUG", "true").lower() == "true"
+    in_ci = environ.get("IN_CI", "false").lower() == "true"
     github_bot_repo = "https://github.com/python-discord/sir-lancebot"
     # Override seasonal locks: 1 (January) to 12 (December)
     month_override = int(environ["MONTH_OVERRIDE"]) if "MONTH_OVERRIDE" in environ else None
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -12,7 +12,7 @@ concurrency:
 
 jobs:
   lint:
-    name: Run pre-commit & flake8
+    name: Run linting & tests
     runs-on: ubuntu-latest
     env:
       # List of licenses that are compatible with the MIT License and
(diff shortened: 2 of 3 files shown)


Akarys42 (changes requested):
I would argue that this new step should be inserted higher up the workflow.

In the case of a syntax error, the error would be reported much better by this step, while flake8 doesn't give any detail about such an error.

Considering this step doesn't take much time, I think it is fine to have it run first.

ChrisLovering:
> I would argue that this new step should be inserted higher up the workflow.
> 
> In the case of a syntax error, the error would be reported much better by this step, while flake8 doesn't give any detail about such an error.
> 
> Considering this step doesn't take much time, I think it is fine to have it run first.

On the flip side flake8 gives errors that this wouldn't give any detail about too.

So is the argument here to order the steps fastest to slowest, so that the workflow fails fast? If so, we can move this new step up, and the license check down.

Akarys42:
My argument is flake8 doesn't give any meaningful error when a file has a syntax error. 

The new run step would give a normal traceback, though. So I think it should be moved up, so if a syntax error is present the run step would fail instead of being stopped at the flake8 step.

ChrisLovering:
> My argument is flake8 doesn't give any meaningful error when a file has a syntax error.
> 
> The new run step would give a normal traceback, though. So I think it should be moved up, so if a syntax error is present the run step would fail instead of being stopped at the flake8 step.

Ahh right, that makes sense now. Yea, I agree.

Akarys42 (approved):
LGTM

bot/constants.py
```diff
@@ -143,6 +143,7 @@ class Client(NamedTuple):
     prefix = environ.get("PREFIX", ".")
     token = environ.get("BOT_TOKEN")
     debug = environ.get("BOT_DEBUG", "true").lower() == "true"
+    in_ci = environ.get("IN_CI").lower() == "true"
```
jchristgit, line 146: What happens if this is unset?

ChrisLovering: It would fail 😓

Fixed in https://github.com/python-discord/sir-lancebot/pull/1039/commits/3aa76661c68787dac9b7c9fa490f94941fa7760d

jchristgit (approved):

jchristgit:
Thanks!

ChrisLovering:
Force push was to squash a fixup commit.

bot/constants.py
jchristgit: **As a...**
Python Discord Admin

**I want to...**
  Only allow starting our community bot in CI

**Such that...**
    I have less robots that can mess with my work

D0rs4n (approved):

PR #1049, opened 2022-04-26
Bookmark enhancements

### LINKED ISSUE #918: un bookmark command
## Description
There would be a way to remove a bookmark. This would be useful while mistakenly creating a bookmark or if you have many of bookmarks and don't want to have them there anymore. 
## Reasoning

If you happen to create a bookmark for a wrong message, you could delete that bookmark. Or if you just have a old bookmark that you don't anymore need. It would keep the bookmarks clear.

## Proposed Implementation

Possibly creating a tag for each bookmark, as example `#4321`. With that you could do `.unbm 4321` or `.unbm #4321`

ChrisLovering:
## Relevant Issues

Closes #918

## Description

This simplifies the bookmark implementation, and adds the ability to delete bookmarks in DMs.

To reduce complexity, the delete bookmark functionality actually allows users to delete any sir-lancebot message within their own DMs, and is labeled as such.

## Did you:

### DIFF
--- a/bot/exts/core/help.py
+++ b/bot/exts/core/help.py
@@ -13,6 +13,7 @@ from bot import constants
 from bot.bot import Bot
 from bot.constants import Emojis
 from bot.utils.commands import get_command_suggestions
+from bot.utils.decorators import whitelist_override
 from bot.utils.pagination import FIRST_EMOJI, LAST_EMOJI, LEFT_EMOJI, LinePaginator, RIGHT_EMOJI
 
 DELETE_EMOJI = Emojis.trashcan
@@ -512,6 +513,7 @@ class Help(DiscordCog):
     """Custom Embed Pagination Help feature."""
 
     @commands.command("help")
+    @whitelist_override(allow_dm=True)
     async def new_help(self, ctx: Context, *commands) -> None:
         """Shows Command Help."""
         try:
--- a/bot/exts/core/help.py
+++ b/bot/exts/core/help.py
@@ -286,7 +286,7 @@ class HelpSession:
             else:
                 results.append(f"<{name}>")
 
-        return f"{cmd.name} {' '.join(results)}"
+        return f"{cmd.qualified_name} {' '.join(results)}"
 
     async def build_pages(self) -> None:
         """Builds the list of content pages to be paginated through in the help message, as a list of str."""
@@ -313,6 +313,8 @@ class HelpSession:
         prefix = constants.Client.prefix
 
         signature = self._get_command_params(self.query)
+        paginator.add_line(f"**```\n{prefix}{signature}\n```**")
+
         parent = self.query.full_parent_name + " " if self.query.parent else ""
         paginator.add_line(f"**```\n{prefix}{parent}{signature}\n```**")
         aliases = [f"`{alias}`" if not parent else f"`{parent}{alias}`" for alias in self.query.aliases]
--- a/bot/exts/core/help.py
+++ b/bot/exts/core/help.py
@@ -316,7 +316,6 @@ class HelpSession:
         paginator.add_line(f"**```\n{prefix}{signature}\n```**")
 
         parent = self.query.full_parent_name + " " if self.query.parent else ""
-        paginator.add_line(f"**```\n{prefix}{parent}{signature}\n```**")
         aliases = [f"`{alias}`" if not parent else f"`{parent}{alias}`" for alias in self.query.aliases]
         aliases += [f"`{alias}`" for alias in getattr(self.query, "root_aliases", ())]
         aliases = ", ".join(sorted(aliases))
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -74,10 +74,9 @@ class Bookmark(commands.Cog):
             await member.send(embed=embed)
         except discord.Forbidden:
             error_embed = self.build_error_embed(f"{member.mention}, please enable your DMs to receive the bookmark.")
+            await channel.send(embed=error_embed)
         else:
             log.info(f"{member} bookmarked {target_message.jump_url} with title '{title}'")
-            return
-        await channel.send(embed=error_embed)
 
     @commands.group(name="bookmark", aliases=("bm", "pin"), invoke_without_command=True)
     @commands.guild_only()
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -49,30 +49,31 @@ class Bookmark(commands.Cog):
         return embed
 
     @staticmethod
-    def build_error_embed(user: discord.Member) -> discord.Embed:
-        """Builds an error embed for when a bookmark requester has DMs disabled."""
+    def build_error_embed(message: str) -> discord.Embed:
+        """Builds an error embed for a given message."""
         return discord.Embed(
             title=random.choice(ERROR_REPLIES),
-            description=f"{user.mention}, please enable your DMs to receive the bookmark.",
+            description=message,
             colour=Colours.soft_red
         )
 
     async def action_bookmark(
         self,
         channel: discord.TextChannel,
-        user: discord.Member,
+        member: discord.Member,
         target_message: discord.Message,
         title: str
     ) -> None:
-        """Sends the bookmark DM, or sends an error embed when a user bookmarks a message."""
+        """Sends the bookmark DM, or sends an error embed when a member bookmarks a message."""
+        embed = self.build_bookmark_dm(target_message, title)
         try:
-            embed = self.build_bookmark_dm(target_message, title)
-            await user.send(embed=embed)
+            await member.send(embed=embed)
         except discord.Forbidden:
-            error_embed = self.build_error_embed(user)
-            await channel.send(embed=error_embed)
+            error_embed = self.build_error_embed(f"{member.mention}, please enable your DMs to receive the bookmark.")
         else:
-            log.info(f"{user} bookmarked {target_message.jump_url} with title '{title}'")
+            log.info(f"{member} bookmarked {target_message.jump_url} with title '{title}'")
+            return
+        await channel.send(embed=error_embed)
 
     @staticmethod
     async def send_reaction_embed(
@@ -117,11 +118,7 @@ class Bookmark(commands.Cog):
         permissions = target_message.channel.permissions_for(ctx.author)
         if not permissions.read_messages:
             log.info(f"{ctx.author} tried to bookmark a message in #{target_message.channel} but has no permissions.")
-            embed = discord.Embed(
-                title=random.choice(ERROR_REPLIES),
-                color=Colours.soft_red,
-                description="You don't have permission to view this channel."
-            )
+            embed = self.build_error_embed(f"{ctx.author.mention} You don't have permission to view this channel.")
             await ctx.send(embed=embed)
             return
 
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -17,7 +17,7 @@ log = logging.getLogger(__name__)
 TIMEOUT = 120
 BOOKMARK_EMOJI = "📌"
 MESSAGE_NOT_FOUND_ERROR = (
-    "You must either provide a valid message to bookmark, or reply to one."
+    "You must either provide a reference to a valid message, or reply to one."
     "\n\nThe lookup strategy for a message is as follows (in order):"
     "\n1. Lookup by '{channel ID}-{message ID}' (retrieved by shift-clicking on 'Copy ID')"
     "\n2. Lookup by message ID (the message **must** be in the current channel)"
@@ -94,7 +94,7 @@ class Bookmark(commands.Cog):
 
         Members can either give a message as an argument, or reply to a message.
 
-        Bookmarks can subsequently be deleted by using the `bookmark delete` command.
+        Bookmarks can subsequently be deleted by using the `bookmark delete` command in DMs.
         """
         target_message: Optional[discord.Message] = target_message or getattr(ctx.message.reference, "resolved", None)
         if target_message is None:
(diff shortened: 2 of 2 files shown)


wookie184 (changes requested):
Deleting bookmarks doesn't seem to work for me. When I try it in DMs I get told I can only use it in the `bot-commands` channel and when I try it in the `bot-commands` channel I get told i'm not authorised to use it. Could this be because there is a `guild_only` decorator on the bookmark group and a `dm_only` decorator on the `delete` command? Or do I have something set up wrong locally?

I get a similar thing with `.help`. Ideally I think `help` should be able to be used anywhere for all commands, but that might be something for a different issue.

wookie184 (approved):
LGTM

bot/exts/utilities/bookmark.py
```diff
@@ -16,6 +16,13 @@
 # Number of seconds to wait for other users to bookmark the same message
 TIMEOUT = 120
 BOOKMARK_EMOJI = "📌"
+MESSAGE_NOT_FOUND_ERROR = (
+    "You must either provide a valid message to bookmark, or reply to one."
+    "\n\nThe lookup strategy for a message is as follows (in order):"
+    "\n1. Lookup by '{channel ID}-{message ID}' (retrieved by shift-clicking on 'Copy ID')"
+    "\n2. Lookup by message ID (the message **must** be in the current channel)"
+    "\n3. Lookup by message URL"
+)
```
ichard26, line 25: This is a bit confusing when it's sent during a `.bm delete` invocation since lookup method one is not possible AFAIK, and two, we are trying to **un**bookmark a message. Not sure how to deal with this without duplication though :confused:

bot/exts/utilities/bookmark.py
```diff
@@ -95,30 +89,40 @@ async def bookmark(
...
-                    "\n2. Lookup by message ID (the message **must** be in the context channel)"
-                    "\n3. Lookup by message URL"
-                )
-            target_message = ctx.message.reference.resolved
+        """
+        Send the author a link to the specified message via DMs.
+
+        Members can either give a message as an argument, or reply to a message.
+
+        Bookmarks can subsequently be deleted by using the `bookmark delete` command.
```
ichard26, line 97:
```suggestion
        Bookmarks can subsequently be deleted by using the `bookmark delete` command in DMs.
```

ichard26 (commented):
Overall it works nicely! Thanks for working on this! I have a few documentation suggestions to improve user usability, and also I'd like to repeat the help command unideal behaviour. 

image

Using `.help` is a bit confusing since it doesn't work in DMs nor any channel except for #bot-commands.

Xithrius:
@ChrisLovering Enhance *faster*!

Will you be finishing up this PR sometime soon?

ChrisLovering:
> Overall it works nicely! Thanks for working on this! I have a few documentation suggestions to improve user usability, and also I'd like to repeat the help command unideal behaviour.
> 
> image
> 
> Using `.help` is a bit confusing since it doesn't work in DMs nor any channel except for #bot-commands.

IMO it's even more confusing if we allow it in DMs, since the Help command only shows the command directly, which makes users think they can run `&delete` directly 
image
Not too sure what to do here.

bot/exts/utilities/bookmark.py
ChrisLovering: https://github.com/python-discord/sir-lancebot/pull/1049/commits/0e4b03a9f2aa1d7a91dc3baa5a189ee4f9c380f5

bot/exts/utilities/bookmark.py
ChrisLovering: https://github.com/python-discord/sir-lancebot/pull/1049/commits/0e4b03a9f2aa1d7a91dc3baa5a189ee4f9c380f5

bot/exts/utilities/bookmark.py
HassanAbouelela: I could’ve sworn I wrote a comment on this, but it seems I didn’t. Anyways, this change is wrong because:
1. This command is decorated with a `guild_only`.
2. We don’t allow commands in DMs unless absolutely necessary.

And speaking of 1, that’s redundant since it’s the default for all commands.

wookie184: This is referring to the `bookmark delete` command, not the `bookmark` command. For the `bookmark delete` command the defaults were overwritten to allow it in (and only in) DMs.

HassanAbouelela: In that case we should probably reconsider. It's policy not to do this, and with decent reason. Core-devs please refer to the [policy doc](https://www.notion.so/pythondiscord/Internal-Policies-86f26339ba86483a90b1b6dd1d157886).

wookie184: Having to jump between DMs and the server to delete a bookmark is not great UX, so in this case I think it makes sense to allow it in DMs. It's a very simple command (implementation) so I don't think there should be any issues caused by this.

HassanAbouelela: I don't think this is better UX, considering all that comes with this (Chris left a comment about it below).

ChrisLovering: When you book mark a command, the link to that message is sent to you in DM. The best UX is to allow users to delete the message directly from the same place.

The UX issue described below is related to the help command, rather than this command directly.

We have two options to fix that, We could allow the help command to be ran in DMs, which leads to the screenshot I posted about below, or we allow the bookmark delete command to also be ran in server, so at least users can get the help info about it.

Regardless of options, i believe this is a very valid use case for a command that can be ran in DMs.

wookie184:
> IMO it's even more confusing if we allow it in DMs, since the Help command only shows the command directly, which makes users think they can run `&delete` directly

Looked into that bug and it seems to be because specifying `root_aliases` causes `bookmark delete` to be registered as a command rather than subcommand, which causes it to be listed even though subcommands usually aren't, and this isn't done correctly (this isn't specific to DMs).

Tentative patch that seems to fix it:
```diff
diff --git a/bot/exts/core/help.py b/bot/exts/core/help.py
index db3c2aa6..f7df0760 100644
--- a/bot/exts/core/help.py
+++ b/bot/exts/core/help.py
@@ -277,7 +277,7 @@ class HelpSession:
             else:
                 results.append(f"<{name}>")

-        return f"{cmd.name} {' '.join(results)}"
+        return f"{cmd.qualified_name} {' '.join(results)}"

     async def build_pages(self) -> None:
         """Builds the list of content pages to be paginated through in the help message, as a list of str."""
@@ -304,8 +304,9 @@ class HelpSession:
         prefix = constants.Client.prefix

         signature = self._get_command_params(self.query)
+        paginator.add_line(f"**```\n{prefix}{signature}\n```**")
+
         parent = self.query.full_parent_name + " " if self.query.parent else ""
-        paginator.add_line(f"**```\n{prefix}{parent}{signature}\n```**")
         aliases = [f"`{alias}`" if not parent else f"`{parent} {alias}`" for alias in self.query.aliases]
         aliases += [f"`{alias}`" for alias in getattr(self.query, "root_aliases", ())]
         aliases = ", ".join(sorted(aliases))
```

If this seems right I don't mind if you want to kaizen this in here or want me to open a separate PR.

ichard26 (commented):
Assuming we are OK with the feature as designed (hi Scale), I'm fine with leaving the help situation as-is. It's not great, but it's acceptable. The rest of this PR looks good to me!

I'm not approving until someone else chimes in on how we'll want to let users get help for this feature. It would be nice to get help working in DMs, but I might be underestimating the impact of such a change.

ChrisLovering:
I've gone ahead and applied @wookie184's suggested fix and rebased onto main.

bot/exts/core/help.py
```diff
@@ -313,6 +313,8 @@ async def _add_command_signature(self, paginator: LinePaginator) -> None:
         prefix = constants.Client.prefix
 
         signature = self._get_command_params(self.query)
+        paginator.add_line(f"**```\n{prefix}{signature}\n```**")
+
         parent = self.query.full_parent_name + " " if self.query.parent else ""
         paginator.add_line(f"**```\n{prefix}{parent}{signature}\n```**")
```
wookie184, line 319: Missed this line being removed in my patch, currently the top line is duplicated
image

```suggestion
```

wookie184: Also with that we can enable the help command in DMs

image

ChrisLovering: [`1de56aa` (#1049)](https://github.com/python-discord/sir-lancebot/pull/1049/commits/1de56aa67f0d5480b2016736bf6dcd6dd4bbff72) and [`050ec2b` (#1049)](https://github.com/python-discord/sir-lancebot/pull/1049/commits/050ec2bf36bc70415f07f4f0aaf3e911e4af118f)

wookie184 (approved):

bot/exts/utilities/bookmark.py
```diff
@@ -42,51 +49,38 @@ def build_bookmark_dm(target_message: discord.Message, title: str) -> discord.Em
...
-                ),
-                colour=Colours.soft_green
-            )
-        )
-
-        await message.add_reaction(BOOKMARK_EMOJI)
-        return message
+            log.info(f"{member} bookmarked {target_message.jump_url} with title '{title}'")
+            return
+        await channel.send(embed=error_embed)
```
HassanAbouelela, line 80: I think we should move this send into the except block, so it's easier to understand what is happening here.

bot/exts/utilities/bookmark.py
```diff
@@ -152,6 +151,34 @@ def event_check(reaction: discord.Reaction, user: discord.Member) -> bool:
...
+        Instead of giving a `target_message` the user may also reply to the message to delete.
+
+        This command allows deleting any message sent by Sir-Lancebot in the user's DMs.
+        """
+        target_message: Optional[discord.Message] = target_message or getattr(ctx.message.reference, "resolved", None)
+        if target_message is None:
+            raise commands.UserInputError(MESSAGE_NOT_FOUND_ERROR)
+
+        if not isinstance(ctx.channel, discord.DMChannel):
+            raise commands.UserInputError("You can only run this command your own DMs!")
```
HassanAbouelela, line 174: Is this actually possible with the whitelist above? If the purpose is for type-casting, we can be more explicit about that.

bot/exts/utilities/bookmark.py
```diff
@@ -16,6 +16,13 @@
 # Number of seconds to wait for other users to bookmark the same message
 TIMEOUT = 120
 BOOKMARK_EMOJI = "📌"
+MESSAGE_NOT_FOUND_ERROR = (
+    "You must either provide a reference to a valid message, or reply to one."
+    "\n\nThe lookup strategy for a message is as follows (in order):"
+    "\n1. Lookup by '{channel ID}-{message ID}' (retrieved by shift-clicking on 'Copy ID')"
+    "\n2. Lookup by message ID (the message **must** be in the current channel)"
```
HassanAbouelela, line 23: This lookup fails for any non-cached message for me. Looking at the source code for the converter, after a cache-miss, it'll try to get the channel, then fetch the message from it, but it always fails to get the channel, even when it has already done other operations in that channel (such as sending a bookmark or a failure).

The reason I commented on this portion specifically is that I think it's best to just not mention this, and not worry about it.

HassanAbouelela (changes requested):

bot/exts/utilities/bookmark.py
ChrisLovering: This is fixed in a more recent version of d.py https://github.com/Rapptz/discord.py/pull/7688

Until we bump up to that, this monkey patch is supposed to resolve the issue you are seeing.
https://github.com/python-discord/sir-lancebot/blob/main/bot/monkey_patches.py#L72

ChrisLovering: Testing it myself, it seems like this message converter just does not have a good time with messages in DM channels for whatever reason.

Possibly we just make it only work with a reply, since that seems to work consistently

bot/exts/utilities/bookmark.py
ChrisLovering: Yea, it is somehow.
image

bot/exts/utilities/bookmark.py
ChrisLovering: Yea, good idea. [`e5bbd71` (#1049)](https://github.com/python-discord/sir-lancebot/pull/1049/commits/e5bbd7135d147bf054c711b9010546f12c4b22e0)

ChrisLovering:
Force push is to rebase onto main

bot/exts/utilities/bookmark.py
HassanAbouelela: I think that’s fine, you’re going to be at the message when running the command anyway

bot/exts/utilities/bookmark.py
HassanAbouelela: I think this is something worth investigating as a bug in the whitelist code, or maybe d.py, but I won’t hold up this PR on it. If you have reproduction steps that would be great.

ChrisLovering: Agreed, I might hold off until after I bump lance up to stable d.py 2.0. Just in the hopes that it disappears.

bot/exts/utilities/bookmark.py
ChrisLovering: [python-discord/sir-lancebot@`36eb3d1` (#1049)](https://github.com/python-discord/sir-lancebot/pull/1049/commits/36eb3d19c742dac3351ff90e44509115867a512b)

HassanAbouelela (approved):
A few things to do after this PR, but besides my comment, everything looks fine. I'll leave you to clean up the PR history, or just merge it, whichever you prefer.

bot/exts/utilities/bookmark.py
```diff
@@ -16,6 +16,13 @@
 # Number of seconds to wait for other users to bookmark the same message
 TIMEOUT = 120
 BOOKMARK_EMOJI = "📌"
+MESSAGE_NOT_FOUND_ERROR = (
```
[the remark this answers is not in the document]
ChrisLovering: This is still used in the bookmark command, where the strategy described actually works. It's only falling over for DM messages.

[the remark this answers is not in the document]
HassanAbouelela: 👍

PR #884, opened 2021-09-26
Fix error with `.help`

TizzySaurus:
## Relevant Issues

Closes #863 

## Description

Fixed the traceback caused by `process.extract` returning 3 items when `dict()` can only take 2.

Ended up changing the data-structure of the argument sent to `HelpQueryNotFound to `list[str]` since only the keys where ever used anyway.

Also noticed there was some inconsistency within the file in regards to `commands.Command` so fixed that too.

NB: I couldn't figure out how to get the bot running with dpy2.0 so this is currently untested, but *should* work.
EDIT: Have since managed to get it to run, and this does work as intended.

## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] Read all the comments in this template?
- [x] Ensure there is an issue open, or link relevant discord discussions?
- [x] Read and agree to the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

PR #1064, opened 2022-06-21
Help command fix, normalize suggestions for unknown commands

RohanJnr:
## Relevant Issues
issue: #863 

Refer to https://github.com/python-discord/sir-lancebot/pull/884

## Fixes
- The help command used to raise an error when provided with an invalid command name (fixed)
- There used to be different command suggestions from the help command and the error handler when an unknown command was provided.
  - This has been fixed by using the same method for finding similar commands.

## Changes
- Providing an invalid subcommand but valid parent command (for example: `!ext listt`, ext is valid but subcommand `listt` is invalid) now sends the help command output for the parent command.

- Using `bot.all_commands` attribute instead of `bot.walk_commands()` generator.
  - `walk_commands()` is not useful in the error_handler because (as it provides us with all subcommands also), when there is an invalid subcommand, the parent command (`commands.Group`) handles this case and sends a help message, hence, never reaching the error_handler.
  - `all_commands` does not give us subcommands (it gives aliases which is pretty useful).

- Using `rapidfuzz` for finding similar commands instead of `difflib` (reasons being speed).

- Error handler now sends multiple command suggestions instead of 1 when user is trying to run an unknown command.

## Yet to decide/to discuss
The help command lists/suggests commands that may not in the user's power to invoke them. Should we filter commands when listing them such that to only show them the commands which they can invoke ?

This is the current output when user invokes help command on a restricted command.

## Did you:

### DIFF
--- a/bot/exts/core/error_handler.py
+++ b/bot/exts/core/error_handler.py
@@ -158,27 +158,32 @@ class CommandErrorHandler(commands.Cog):
 
     async def send_command_suggestion(self, ctx: commands.Context, command_name: str) -> None:
         """Sends user similar commands if any can be found."""
-        if similar_command_data := get_command_suggestions(list(self.bot.all_commands.keys()), command_name, limit=1):
-            similar_command_name = similar_command_data[0]
-            similar_command = self.bot.get_command(similar_command_name)
+        command_suggestions = []
+        if similar_command_names := get_command_suggestions(list(self.bot.all_commands.keys()), command_name):
+            for similar_command_name in similar_command_names:
+                similar_command = self.bot.get_command(similar_command_name)
 
-            if not similar_command:
-                return
-
-            log_msg = "Cancelling attempt to suggest a command due to failed checks."
-            try:
-                if not await similar_command.can_run(ctx):
-                    log.debug(log_msg)
+                if not similar_command:
                     return
-            except commands.errors.CommandError as cmd_error:
-                log.debug(log_msg)
-                await self.on_command_error(ctx, cmd_error)
-                return
+
+                log_msg = "Cancelling attempt to suggest a command due to failed checks."
+                try:
+                    if not await similar_command.can_run(ctx):
+                        log.debug(log_msg)
+                        return
+                except commands.errors.CommandError as cmd_error:
+                    log.debug(log_msg)
+                    await self.on_command_error(ctx, cmd_error)
+                    return
+
+                command_suggestions.append(similar_command_name)
 
             misspelled_content = ctx.message.content
             e = Embed()
             e.set_author(name="Did you mean:", icon_url=QUESTION_MARK_ICON)
-            e.description = misspelled_content.replace(command_name, similar_command_name, 1)
+            e.description = "\n".join(
+                misspelled_content.replace(command_name, cmd, 1) for cmd in command_suggestions
+            )
             await ctx.send(embed=e, delete_after=RedirectOutput.delete_delay)
 
 
--- a/bot/exts/core/error_handler.py
+++ b/bot/exts/core/error_handler.py
@@ -1,4 +1,3 @@
-import difflib
 import logging
 import math
 import random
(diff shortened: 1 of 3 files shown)


bot/exts/core/error_handler.py
```diff
@@ -158,31 +158,32 @@ async def on_command_error(self, ctx: commands.Context, error: commands.CommandE
...
-            log_msg = "Cancelling attempt to suggest a command due to failed checks."
-            try:
-                if not await similar_command.can_run(ctx):
+        command_suggestions = []
+        if similar_command_names := get_command_suggestions(list(self.bot.all_commands.keys()), command_name):
+            for similar_command_name in similar_command_names:
+                similar_command = self.bot.get_command(similar_command_name)
+
+                if not similar_command:
+                    return
```
Bluenix2, line 167: There are other similar command names we can try if we can't find this one, I would suggest changing this to `continue`:
```suggestion
                    continue
```

bot/exts/core/error_handler.py
```diff
@@ -158,31 +158,32 @@ async def on_command_error(self, ctx: commands.Context, error: commands.CommandE
...
+
+                log_msg = "Cancelling attempt to suggest a command due to failed checks."
+                try:
+                    if not await similar_command.can_run(ctx):
+                        log.debug(log_msg)
+                        return
```
Bluenix2, line 173: Same as above, we can assume that there may be other suggestions:
```suggestion
                        continue
```

bot/exts/core/help.py
```diff
@@ -115,6 +119,9 @@ def __init__(
     def _get_query(self, query: str) -> Union[Command, Cog]:
         """Attempts to match the provided query with a valid command or cog."""
         command = self._bot.get_command(query)
+
+        log.info(query)
+        log.info(command)
```
Bluenix2, line 124: Were these just local debugs? They don't make a lot of sense without any explanation. I think it is best if we just remove them?

bot/exts/core/help.py
```diff
@@ -153,12 +160,17 @@ def _handle_not_found(self, query: str) -> None:
 
         Will pass on possible close matches along with the `HelpQueryNotFound` exception.
         """
-        # Combine command and cog names
-        choices = list(self._bot.all_commands) + list(self._bot.cogs)
+        # Check if parent command in valid incase subcommand is invalid.
+        if " " in query:
+            parent, sub_command = query.split()
```
Bluenix2, line 165: Don't we have deeper nested subcommands than once? I think you should pass `maxsplit=1`.

brad90four (approved):
Content wise this looks good to me, have not verified with independent testing yet.

bot/exts/core/help.py
```diff
@@ -41,14 +41,18 @@ class HelpQueryNotFound(ValueError):
...
-    query, where keys are the possible matched command names and values are the likeness match scores.
+    Params:
+    possible_matches: list of similar command names.
+    parent_command: parent command of an invalid subcommand. Only available when an invalid subcommand
+                    has been passed.
     """
 
-    def __init__(self, arg: str, possible_matches: dict = None):
+    def __init__(
+            self, arg: str, possible_matches: list[str] = None, *, parent_command: Command = None
```
GDWR, line 51: Missing `Optional` or `| None` typing.
```suggestion
            self, arg: str, possible_matches: list[str] | None = None, *, parent_command: Command | None = None
```

GDWR (changes requested):
Looks good, just a slight type hint that could catch somebody out.

ichard26:
I'm in the middle of a review (sorry for taking so long!) but I ran out of time to finish it today :/ I'll submit my review tomorrow.

bot/exts/core/help.py
RohanJnr: Good catch, though we need to use `Optional` from the `typing` module since the project is running on 3.9 and not 3.10.

GDWR (commented):
Looks good, I'll leave the final approval to @ichard26 so it doesn't auto-merge. 
(I think that is how this works)

ichard26 (changes requested):
I wasn't able to find time to do a full review with manual testing (and I feel bad about delaying the PR even more), so consider this just a read-through review. On that note, it looks good except for a poorly written comment.

wookie184 (approved):
Had a quick look over, looks good and seems to work nicely, thanks 👍

PR #1070, opened 2022-07-09
Update .uwu to work with replies

### LINKED ISSUE #1069: Add message replies support for `.uwu`
## Description
I think it would be great that if you replied to a message with `.uwu` and no other text, that lancebot would uwu-ify the replied-to message. Alternatively, add support for `.uwu` in a context menu!

## Reasoning
We must allow `.uwu` in all its forms. But I think it would be a fun feature to implement and a good way to get familiar with our code base.

## Proposed Implementation
If there is no message content, check if there is a message that has been replied to, and if there has use the text from *that* message as the text to uwu-ify.

Or context menu, but I have no idea how context menus work in d.py

shenanigansd:
## Relevant Issues

Fixes #1069 

## Description

Updates `.uwu` to check for content in a replied to message if the current message is empty.

image

## Did you:

### DIFF
--- a/bot/exts/fun/uwu.py
+++ b/bot/exts/fun/uwu.py
@@ -1,7 +1,9 @@
 import random
 import re
+import typing as t
 from functools import partial
 
+import discord
 from discord.ext import commands
 from discord.ext.commands import Cog, Context, clean_content
 
@@ -105,7 +107,7 @@ class Uwu(Cog):
         return input_string
 
     @commands.command(name="uwu", aliases=("uwuwize", "uwuify",))
-    async def uwu_command(self, ctx: Context, *, text: clean_content(fix_channel_mentions=True)) -> None:
+    async def uwu_command(self, ctx: Context, *, text: t.Optional[str] = None) -> None:
         """
         Echo an uwuified version the passed text.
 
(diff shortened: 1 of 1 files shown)


bot/exts/fun/uwu.py
```diff
@@ -105,15 +107,25 @@ def _uwuify(self, input_string: str, *, stutter_strength: float = 0.2, emoji_str
         return input_string
 
     @commands.command(name="uwu", aliases=("uwuwize", "uwuify",))
-    async def uwu_command(self, ctx: Context, *, text: clean_content(fix_channel_mentions=True)) -> None:
+    async def uwu_command(self, ctx: Context, *, text: t.Optional[str] = None) -> None:
```
HassanAbouelela, line 110: Shouldn't we make this an optional `clean_content` as it was originally, or call the converter in the body of the function?

HassanAbouelela: Seeing what we do below with the replies, we should probably call it in the body itself after getting the final text we'll be working with.

shenanigansd: fa5c359e7b4adc36d963d951c5ae79908c82afa5

ChrisLovering (approved):

HassanAbouelela (approved):
Yes

PR #1077, opened 2022-07-12
GitHub issues embeds: preserve order and display repo name and issue number

### LINKED ISSUE #743: Issue/PR hyperlink embed upgrade
### What this issue solves

This issue aims to solve two things:

- The format in which issues and pull requests are displayed in the embed that the bot automatically sends.
- The order in which the issues and pull requests are sent. This wasn't part of the original issue but I figure if this is correctly implemented then this could be fixed too.

### Current implementation

Upon mentioning an issue or pull requests in a message such as `sir-lancebot#69` or `bot#420`, the issue/PRs are displayed in an embed in a random order without knowing which issue/PR is which. This is not a problem when a single issue/PR is mentioned but with anymore it becomes hard to figure out which is which quickly. Here is an example:

image

It is unclear which pull request is `site#474` and which is `site#477`.

Currently the [code](https://github.com/python-discord/sir-lancebot/blob/a5d138ef76fec71b2d737e24889c7de494fdbc0d/bot/exts/evergreen/issues.py#L246) uses a `set` and so the order is not preserved and we can't link a URL back to the issue or pull request without doing some fancy URL stuff.

### Proposed implementation

[This line](https://github.com/python-discord/sir-lancebot/blob/a5d138ef76fec71b2d737e24889c7de494fdbc0d/bot/exts/evergreen/issues.py#L246) should change from 
```py
issues = set(issues)
```
to
```py
issues = list(dict.fromkeys(issues))
```
And on [this line](https://github.com/python-discord/sir-lancebot/blob/main/bot/exts/evergreen/issues.py) we could simply just change the format of the string as it is added onto the embed to include the issue/PR number.

### Proposed Implementation 2

This implementation involves the [`more_itertools`](https://more-itertools.readthedocs.io/en/stable/) library. Specifically, the `more_itertools.unique_everseen` function which would mean the previous change would instead need to be to:
```py
issues = list(more_itertools.unique_everseen(issues))
```
This comes with the advantage of `more_itertools` being used in other parts of Sir Lancebot.

### End result
Going with either implementation, we would like to reach an embed that looks more like this, namely, it includes the issue/PR number in the hyperlink. Example:

3DC12400-A012-4CD2-B6DA-A8249850D016
## Reasoning
I think this would be useful as it prevents community members from clicking on each hyperlink and checking if they managed to somehow click the correct link. Otherwise, going back and trying again. It just saves time and effort.

### LINKED ISSUE #1076: Embeds for multiple GitHub issues don't maintain ordering
## Description
When Sir Lancebot creates an embed listing multiple GitHub issues, the issues are not in the same order that they were in in the orginal message

shenanigansd:
## Relevant Issues
Fixes #743  

## Description
Display the name of the repo and the number of the issue in GitHub issues embeds.
This now also preserves the order that the issues were added in.

## Did you:


Numerlor:
I'd consider using `more_itertools.unique_everseen` here if the library can be added as it should be clearer than the dict. Adding the issue number before the issue title should also help with matching up the issue numbers to their issues, as it is more tasking to match them up yourself even when ordered

shenanigansd:
Example:
image

bot/exts/utilities/githubinfo.py
```diff
@@ -192,7 +195,7 @@ async def on_message(self, message: discord.Message) -> None:
 
             log.trace(f"Found {issues = }")
             # Remove duplicates
-            issues = set(issues)
+            issues = unique_everseen(issues)
```
Numerlor, line 198:
```suggestion
            issues = list(unique_everseen(issues))
```
The function returns an iterator which does not have a length that is checked below

Numerlor (approved):

Xithrius (approved):
Merge conflicts must be resolved. Otherwise, looks good to me.

bot/exts/utilities/githubinfo.py
```diff
@@ -192,7 +195,7 @@ async def on_message(self, message: discord.Message) -> None:
 
             log.trace(f"Found {issues = }")
             # Remove duplicates
-            issues = set(issues)
+            issues = list(unique_everseen(issues))
```
ChrisLovering, line 198: I'm not sure adding more_itertools just for this is worth it, when the below uses only built-ins, and is faster.

```suggestion
            issues = list(dict.fromkeys(issues))
```

ChrisLovering (approved):

PR #1091, opened 2022-08-20
Use buttons for the bookmark command

### LINKED ISSUE #932: Replace reactions in the `bookmark` command with buttons.
## Description
Right now, bookmarking messages as someone who is not the command invoker requires adding a reaction. I believe that the system can be made much more better using buttons as it greatly improves the UI and makes it a bit more simpler.

## Reasoning
While using interactions, we now have an option to make any error only visible to the interaction author by making the error message ephemeral. This would mean that the error message will not be publicly visible and a confirmation message that is only visible to the interaction author can be sent as well.

## Proposed Implementation
image

## Additional Details
We can even make use of the link buttons and attach them to the bookmarked message. When clicked, they would redirect the user to the message. Adding a button for it would make the embed solely limited to the message content and a `View Message` link button would improve the UI as well as make navigation to the message a bit more easier.

Diabolical5777:
## Relevant Issues

Closes #932

## Description

I replaced the reactions used currently with buttons

## Did you:

### DIFF
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -1,7 +1,6 @@
-import asyncio
 import logging
 import random
-from typing import Optional
+from typing import Callable, Optional
 
 import discord
 from discord.ext import commands
@@ -15,7 +14,6 @@ log = logging.getLogger(__name__)
 
 # Number of seconds to wait for other users to bookmark the same message
 TIMEOUT = 120
-BOOKMARK_EMOJI = "📌"
 MESSAGE_NOT_FOUND_ERROR = (
     "You must either provide a reference to a valid message, or reply to one."
     "\n\nThe lookup strategy for a message is as follows (in order):"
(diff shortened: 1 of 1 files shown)


bot/exts/utilities/bookmark.py
```diff
@@ -25,6 +23,53 @@
 )
 
 
+class LinkTargetMessage(discord.ui.View):
+    """The button that relays the user to the bookmarked message."""
+
+    def __init__(
+        self,
+        target_message: discord.Message
```
Dorukyum, line 31: Let's add a trailing comma if we choose to use this paranthesis syntax
```suggestion
        target_message: discord.Message,
```

bot/exts/utilities/bookmark.py
```diff
@@ -25,6 +23,53 @@
...
+
+        self.bookmark_function = bookmark_function
+        self.clicked: list[int] = [author.id]
+        self.channel = channel
+        self.target_message = target_message
+
+        super().__init__(label="Receive Bookmark", style=discord.ButtonStyle.green)
+
+    async def callback(self, interaction: discord.Interaction) -> None:
+        """The callback to the function."""
```
Dorukyum, line 58: Maybe to be more clear?
```suggestion
        """The button callback."""
```

bot/exts/utilities/bookmark.py
Diabolical5777: I considered that but to be consistent with how they make the init currently, I left it out. And I am not sure what you mean by paranthesis syntax

Dorukyum: I mean we don't really need to seperate this into multiple lines, line below is 56 chars:
```python
    def __init__(self, target_message: discord.Message):
```

Robin5605:
It looks like we have a full on `discord.ui.View` subclass for a simple link button - but for the actual button that sends the bookmark in the DM, we're just doing:
```py
view = discord.ui.View(timeout=TIMEOUT)
view.add_item(
    SendBookmark(self.action_bookmark, ctx.author, ctx.channel, target_message)
)
```
I think this should be flipped around. There's no need for a full fledged subclass for a simple link button, but maybe we can use a view subclass for the actual bookmark. I doubt we'll also need a button subclass since we're not doing anything dynamic

Diabolical5777:
> It looks like we have a full on `discord.ui.View` subclass for a simple link button - but for the actual button that sends the bookmark in the DM, we're just doing:
> 
> ```python
> view = discord.ui.View(timeout=TIMEOUT)
> view.add_item(
>     SendBookmark(self.action_bookmark, ctx.author, ctx.channel, target_message)
> )
> ```
> 
> I think this should be flipped around. There's no need for a full fledged subclass for a simple link button, but maybe we can use a view subclass for the actual bookmark. I doubt we'll also need a button subclass since we're not doing anything dynamic

I think you have it flipped. I used the button subclass for the button that is sent in the channel where the bookmark is invoked in so others can press the button. For the button in DM's that redirects to the bookmarked message, I used a view directly.

Robin5605:
You're right that I had it flipped. The only other thing I might recommend we change is maybe subclassing `discord.ui.View` instead of a `discord.ui.Button`? Usually that's what we do for non-dynamic components. And I think all the parameters we pass into `SendBookmark` can be passed into the view subclass, and we shouldn't lose any functionality

PR #1092, opened 2022-08-23
Python 3.10 & bot-core migration

ChrisLovering:
## Description
This bumps the bot up to 3.10, bumps all deps. The bumped deps include bot-core which comes with a move to the stable Discord.py 2.0 release.

Sir-Lancebot never had the async cog loading treatment, so this PR also includes that.

There are also miscellaneous fixes included in here for breaking changes in dep bumps.

Something worth mentioning is that I had to change our approach to dry running the bot in CI, see https://github.com/python-discord/sir-lancebot/pull/1092/commits/ab218f2fca9149df2cbde9b8af75f4ba23b6ae8b. This is because cog laoding is managed by bot-core in the bot's setup_hook, which is called after logging in. Obviously we don't want to give CI a bot token, so these changes were required.
This new approach does output some warnings, see the recent successful action, which can safely be ignored, but may confuse users if they actually get an error in that step, and see all the extra output.

## Did you:


bot/__init__.py
```diff
@@ -41,3 +45,5 @@
     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
 
 apply_monkey_patches()
+
+instance: "Bot" = None  # Global Bot instance.
```
wookie184, line 49: Why do we need this?

ChrisLovering: We use it in a few places in bot, such as utils where we don't have access to the bot var very easily.

I've done this here mostly for consistency.

ChrisLovering: https://sourcegraph.com/search?q=context:global+repo:python-discord.bot+%5Cbinstance%5Cb&patternType=regexp for examples

wookie184: Ah, I see

wookie184 (approved):
Great stuff 🎉

ChrisLovering:
Rebased onto main. Only commit added was https://github.com/python-discord/sir-lancebot/commit/aaa96359448a39a5bbb9a0fd403f23b59b5df860 which doesn't conflict with any of these changes.

HassanAbouelela (approved):
This was a big PR. Some would say it was too big. Others are simply wrong.

PR #1126, opened 2022-10-17
Remove calls to deprected `.flatten()` method.

### LINKED ISSUE #1120: AttributeError: 'async_generator' object has no attribute 'flatten'
Sentry Issue: [SIR-LANCEBOT-8S](https://sentry.io/organizations/python-discord/issues/3674722961/?referrer=github_integration)

```
AttributeError: 'async_generator' object has no attribute 'flatten'
  File "discord/client.py", line 409, in _run_event
    await coro(*args, **kwargs)
  File "bot/utils/decorators.py", line 81, in guarded_listener
    return await listener(*args, **kwargs)
  File "bot/exts/holidays/halloween/candy_collection.py", line 93, in on_reaction_add
    await self.hacktober_channel.history(limit=10).flatten()

Ignoring exception in on_reaction_add
```

TizzySaurus:
## Relevant Issues

Closes #1120.

## Description

Replaces `.flatten()` calls with an `async for` loop (the easterquiz ones had to be expanded due to a constraint with having an async comprehension within a non-async comprehension).

## Did you:


bot/resources/utilities/ryanzec_colours.json
```diff
@@ -957,6 +958,7 @@
     "Ochre": "CC7722",
     "Off Green": "E6F8F3",
     "Off Yellow": "FEF9E3",
+    "OG Blurple": "7289DA",
```
wookie184, line 961: I don't think these changes should be here?

bot/exts/holidays/halloween/candy_collection.py
```diff
@@ -90,7 +90,7 @@ async def on_reaction_add(self, reaction: discord.Reaction, user: Union[discord.
 
             recent_message_ids = map(
                 lambda m: m.id,
-                await self.hacktober_channel.history(limit=10).flatten()
+                [message async for message in self.hacktober_channel.history(limit=10)]
             )
```
wookie184, line 94: Shouldn't need the `map` any more, just get the id in the comprehension.

bot/exts/holidays/easter/egghead_quiz.py
```diff
@@ -65,23 +65,43 @@ async def eggquiz(self, ctx: commands.Context) -> None:
...
             results.append(string)
 
+        users = []
+        for reaction in msg.reactions:
+            if str(reaction.emoji) != correct:
+                continue
+
+            async for user in reaction.users():
+                users.append(user)
+
```
wookie184, line 102: Putting a break here like on the last one would make the behaviour more obvious.

TizzySaurus: I don't think a `break` works here, since we *do* want to iterate over all the reactions in this case.

What I will do though, is make `users` a set, since we don't the same user showing up twice.

ChrisLovering (approved):

bot/exts/holidays/easter/egghead_quiz.py
```diff
@@ -97,7 +121,11 @@ async def eggquiz(self, ctx: commands.Context) -> None:
     @staticmethod
     async def already_reacted(message: discord.Message, user: Union[discord.Member, discord.User]) -> bool:
         """Returns whether a given user has reacted more than once to a given message."""
-        users = [u.id for reaction in [await r.users().flatten() for r in message.reactions] for u in reaction]
+        users = []
+        for reaction in message.reactions:
+            async for usr in reaction.users():
+                users.append(usr.id)
+
         return users.count(user.id) > 1  # Old reaction plus new reaction
```
ionite34, line 129: Time complexity in this part seems like it can be better, currently iterates for 2N times for N total users reacted. For 2 reactions, each with 500 users, currently, we iterate a constant of 2000 times (for loop that builds list + list.count) on each new reaction.

Suggest this as an alternative:
In the same case of 1000 users, this would iterate a best case of 1 time and a worst case of 500 times.
```py
@staticmethod
async def already_reacted(new_reaction: discord.Reaction, user: Union[discord.Member, discord.User]) -> bool:
    """Returns whether a given user has reacted more than once to a given message."""
    message = new_reaction.message
    for reaction in message.reactions:
        if reaction.emoji == new_reaction.emoji:
            continue  #  Skip new reaction, duplicate impossible
        async for usr in reaction.users():
            # Duplicate if user id matches anywhere else
            if usr.id == user.id:
                return True
    return False
```

ionite34 (changes requested):

bot/exts/holidays/easter/egghead_quiz.py
TizzySaurus: Done, thanks :+1:

ionite34 (approved):
lgtm

PR #1148, opened 2022-11-15
Add ability to view previous topics with .topic command

### LINKED ISSUE #1145: Show previous conversation prompts after .topic command is reacted to
## Description

After a member invokes the "topic" command, they have the ability to re-roll the embed to choose a different conversation prompt. When they do this, the previous contents of the message is overwritten. Instead, the content should be edited to include the new prompt with the old prompt(s) above it, with a number corresponding to the order it was created in.

## Reasoning

When the person who invokes the command sees a topic they don't want to comment on, they tend to re-roll it by reacting to the message. However, other people in the chat may see it and have something to say, but the message is edited before they can respond. If the change I proposed is added, this person could just put the number of that prompt at the start of his or her message.

## Proposed Implementation

**Current System**
Bot message:
"What is your favorite PyPI module?"

invoker reacts to re-roll

New bot message:
"How has Python changed your life?"

**Proposed Change**
Bot message:
"What is your favorite PyPI module?"

invoker reacts to re-roll

New bot message:
"1. What is your favorite PyPI module?
2. How has Python changed your life?"

Anonymous4045:
## Relevant Issues

Closes #1145 

## Description

After a user invokes the ".topic" command and reacts with the reroll emoji, the embed will now show the previous few conversation prompts. It cannot display more than 256 characters because of the limit on embed title character length, which is usually around 5 prompts.
## Did you:

### DIFF
--- a/bot/exts/utilities/conversationstarters.py
+++ b/bot/exts/utilities/conversationstarters.py
@@ -43,7 +43,7 @@ class ConvoStarters(commands.Cog):
         self.bot = bot
 
     @staticmethod
-    def _build_topic_embed(channel_id: int) -> discord.Embed:
+    def _build_topic_embed(channel_id: int, previous_topic: None | str) -> discord.Embed:
         """
         Build an embed containing a conversation topic.
 
@@ -56,21 +56,46 @@ class ConvoStarters(commands.Cog):
             color=discord.Colour.og_blurple()
         )
 
-        try:
-            channel_topics = TOPICS[channel_id]
-        except KeyError:
-            # Channel doesn't have any topics.
-            embed.title = f"**{next(TOPICS['default'])}**"
+        if previous_topic is None:
+            # Message first sent
+            try:
+                channel_topics = TOPICS[channel_id]
+            except KeyError:
+                # Channel doesn't have any topics.
+                embed.title = f"**{next(TOPICS['default'])}**"
+            else:
+                embed.title = f"**{next(channel_topics)}**"
         else:
-            embed.title = f"**{next(channel_topics)}**"
+            # Message is being edited
+
+            try:
+                channel_topics = TOPICS[channel_id]
+            except KeyError:
+                # Channel doesn't have any topics.
+                new_topic = f"**{next(TOPICS['default'])}**"
+            else:
+                new_topic = f"\n**{next(channel_topics)}**"
+
+            total_topics = previous_topic.count("\n") + 1
+
+            # Add 1 before first topic
+            if total_topics == 1:
+                previous_topic = f"1. {previous_topic}"
+
+            embed.title = previous_topic + f"\n{total_topics + 1}. {new_topic}"
+
+            # When the embed will be larger than the limit, use the previous embed instead
+            if len(embed.title) > 256:
+                embed.title = previous_topic
+
         return embed
 
     @staticmethod
     def _predicate(
-        command_invoker: Union[discord.User, discord.Member],
-        message: discord.Message,
-        reaction: discord.Reaction,
-        user: discord.User
+            command_invoker: Union[discord.User, discord.Member],
+            message: discord.Message,
+            reaction: discord.Reaction,
+            user: discord.User
     ) -> bool:
         user_is_moderator = any(role.id in MODERATION_ROLES for role in getattr(user, "roles", []))
         user_is_invoker = user.id == command_invoker.id
@@ -83,9 +108,9 @@ class ConvoStarters(commands.Cog):
         return is_right_reaction
 
     async def _listen_for_refresh(
-        self,
-        command_invoker: Union[discord.User, discord.Member],
-        message: discord.Message
+            self,
+            command_invoker: Union[discord.User, discord.Member],
+            message: discord.Message
     ) -> None:
         await message.add_reaction("🔄")
         while True:
@@ -101,7 +126,9 @@ class ConvoStarters(commands.Cog):
                 break
 
             try:
-                await message.edit(embed=self._build_topic_embed(message.channel.id))
+                # The returned discord.Message object from discord.Message.edit is different than the current
+                # discord.Message object, so it must be reassigned to update properly
+                message = await message.edit(embed=self._build_topic_embed(message.channel.id, message.embeds[0].title))
             except discord.NotFound:
                 break
 
@@ -109,7 +136,7 @@ class ConvoStarters(commands.Cog):
                 await message.remove_reaction(reaction, user)
 
     @commands.command()
-    @commands.cooldown(1, 60*2, commands.BucketType.channel)
+    @commands.cooldown(1, 60 * 2, commands.BucketType.channel)
     @whitelist_override(channels=ALL_ALLOWED_CHANNELS)
     async def topic(self, ctx: commands.Context) -> None:
         """
@@ -117,7 +144,7 @@ class ConvoStarters(commands.Cog):
 
         Allows the refresh of a topic by pressing an emoji.
         """
-        message = await ctx.send(embed=self._build_topic_embed(ctx.channel.id))
+        message = await ctx.send(embed=self._build_topic_embed(ctx.channel.id, None))
         self.bot.loop.create_task(self._listen_for_refresh(ctx.author, message))
 
 


Xithrius:
I'll keep this open, but be sure next time to get your issue approved by a core dev before continuing with a PR.

ChrisLovering:
Could you share a screenshot of what this looks like after being used a few times?

I think we will also need to add a limit to the number of previous topics that get shown, if there isn't already present. Users could easily spam the button and cause a very large embed from the bot otherwise.

Anonymous4045:
> 

Discord sets the limit on the number of characters in the title of an embed to be 256. From my testing, that seems to mean there's around 4 or 5 topics that can be on display at once. Personally I think 4 or 5 is enough for people to have something to talk about, so it seems good to me.

Here's some screenshots of my bot running the code

Screenshots

bot/exts/utilities/conversationstarters.py
```diff
@@ -83,9 +107,9 @@ def _predicate(
         return is_right_reaction
 
     async def _listen_for_refresh(
-        self,
-        command_invoker: Union[discord.User, discord.Member],
-        message: discord.Message
+            self,
+            command_invoker: Union[discord.User, discord.Member],
+            message: discord.Message
```
ChrisLovering, line 112: Could you revert this change?

bot/exts/utilities/conversationstarters.py
```diff
@@ -101,23 +125,25 @@ async def _listen_for_refresh(
...
+                message = await message.edit(embed=self._build_topic_embed(message.channel.id, message.embeds[0].title))
             except discord.NotFound:
                 break
 
             with suppress(discord.NotFound):
                 await message.remove_reaction(reaction, user)
 
     @commands.command()
-    @commands.cooldown(1, 60*2, commands.BucketType.channel)
+    @commands.cooldown(1, 60 * 2, commands.BucketType.channel)
```
ChrisLovering, line 138: Could you revert this change?

bot/exts/utilities/conversationstarters.py
```diff
@@ -56,21 +56,45 @@ def _build_topic_embed(channel_id: int) -> discord.Embed:
...
+
+            # Add 1 before first topic
+            if total_topics == 1:
+                previous_topic = f"1. {previous_topic}"
+
+            embed.title = previous_topic + f"\n{total_topics + 1}. {new_topic}"
+
+            # When the embed will be larger than the limit, use the previous embed instead
+            if len(embed.title) > 256:
+                embed.title = previous_topic
```
ChrisLovering, line 88: This logic means that clicking on the topic refresh button doesn't actually output a new topic, it just uses the previous input, if the title is beyond the limit.

I think a better way to deal with this would be when a user clicks on the emoji the new topic becomes number 1 and the rest get pushed down a number.
Then, when adding the topics if the last topic would make the title go beyond 256 characters, don't add it.

This would mean the newest topic is always the number 1 topic, and then the older ones are there a reference, until the button is pressed too many times.

bot/exts/utilities/conversationstarters.py
```diff
@@ -56,21 +56,45 @@ def _build_topic_embed(channel_id: int) -> discord.Embed:
...
         else:
-            embed.title = f"**{next(channel_topics)}**"
+            # Message is being edited
+            try:
+                channel_topics = TOPICS[channel_id]
+            except KeyError:
+                # Channel doesn't have any topics.
+                new_topic = f"**{next(TOPICS['default'])}**"
+            else:
+                new_topic = f"**{next(channel_topics)}**"
```
ChrisLovering, line 76: This is duplicated code, doing the shared logic first and returning early if it's the first message means you don't need to do this twice.
```python
        try:
            channel_topics = TOPICS[channel_id]
        except KeyError:
            # Channel doesn't have any topics.
            embed.title = f"**{next(TOPICS['default'])}**"
        else:
            embed.title = f"**{next(channel_topics)}**"

        if previous_topic is None:
            # This is the first topic being sent
            return embed

        total_topics = previous_topic.count("\n") + 1
        # Add 1 before first topic
        if total_topics == 1:
            previous_topic = f"1. {previous_topic}"

        embed.title = previous_topic + f"\n{total_topics + 1}. {embed.title}"

        # When the embed will be larger than the limit, use the previous embed instead
        if len(embed.title) > 256:
            embed.title = previous_topic

        return embed
```

ChrisLovering (changes requested):

Anonymous4045:
> I think a better way to deal with this would be when a user clicks on the emoji the new topic becomes number 1 and the rest get pushed down a number.

I considered this,but this would change the ordering of the older messages. One point of adding this feature is so that people wanting to respond to a specific topic could include the number of that topic in with their reply, e.g. someone responding to the second conversation starter could say something like "For topic 2, ...". Changing the numbers could be confusing or change the sentence someone was referring to.

ChrisLovering:
> > I think a better way to deal with this would be when a user clicks on the emoji the new topic becomes number 1 and the rest get pushed down a number.
> 
> I considered this,but this would change the ordering of the older messages. One point of adding this feature is so that people wanting to respond to a specific topic could include the number of that topic in with their reply, e.g. someone responding to the second conversation starter could say something like "For topic 2, ...". Changing the numbers could be confusing or change the sentence someone was referring to.

If we're going to go with that behaviour, we should remove the reaction button when it will no longer send a new topic.

wookie184:
This seems good to me, the only thing I would suggest is limiting the number of prompts to 3, as 5 seems like quite a lot (i'd still leave the detection for when it goes over the title limit, just in case there are 3 really long topics).

Let's see what others think first though, as I don't mind too much and some people may prefer more.

Robin5605:
Could you move the previous topics as part of the description instead? I feel like it might be too distracting to be in all bold and big like that

bot/exts/utilities/conversationstarters.py
```diff
@@ -63,7 +65,24 @@ def _build_topic_embed(channel_id: int) -> discord.Embed:
...
+            return embed, False
+
+        total_topics = previous_topic.count("\n") + 1
+        # Add 1 before first topic
+        if total_topics == 1:
+            previous_topic = f"1. {previous_topic}"
+
+        embed.title = previous_topic + f"\n{total_topics + 1}. {embed.title}"
+
+        # When the embed will be larger than the limit, use the previous embed instead
```
Ibrahim2750mi, line 80: The comment seems wrong and unnecessary. It is wrong because the comment talks about the embed as a whole, while we are operating on embed's title. It is unnecessary because it can be inferred from the code that we reassign the `embed.title` if it exceeds the title character limit

bot/exts/utilities/conversationstarters.py
Ibrahim2750mi: @Anonymous4045 Also resolve this, by implementing it.

Anonymous4045:
> Could you move the previous topics as part of the description instead? I feel like it might be too distracting to be in all bold and big like that

We could do that, but there's a few things to work out first. What would the title in bold be? And what would we set the limit for the topics to show as, since we wouldn't be bound by the title character limit?

bot/exts/utilities/conversationstarters.py
Anonymous4045: @Ibrahim2750mi I thought we were going with preserving the order that the topics appeared in, so people could reference them by number and not have that number change

Ibrahim2750mi: But that will not allow any new topics to be shown in the embed. Therefore as Chris said the new topics should be added on the number 1 position and oldest topic should be at number 5.

Anonymous4045: @Ibrahim2750mi I'm not sure I get what you mean. New topics are currently added to the end of the embed's title in a way in which their number doesn't change, so if it says something like "2. How did you start Python?" the number 2 won't change after a third topic is added. If that's not what you meant, could you explain a bit more?

Ibrahim2750mi: If the embed's title exceed the character limit you directly use the `previous_topic` which means that it would not append any new topics, so it will stay stuck at those already present topics no matter how many times it is called, because your function just returns the `previous_topic` when it has reached the character limit. Eg: assume this has exceeded the character limit
```
1. How to do python
2. How to not do python
3. beep boop
```
Now even if your function is called it would just return the same embed that is shown in the codeblock, a better approach would be to keep the newest topic at top and the oldest at bottom like this
```
1. beep boop
2. How to not do python
3. How to do python
```
And now if your function is called it should omit `How to do python` so that a new topic can be added to it. Keep in mind you can omit multiple topics if the new topic still exceeds the character limit.
```
1. New cool topic
2. beep boop
3. How to not do python
```

Anonymous4045: That's certainly an option, but that would change the order of each topic, which could cause some confusion for people unaware of how it works or with responding to a particular prompt. At the moment, the reaction disappears once the limit is reached to dissuade people from trying to get too many prompts. The reason being that someone in the chat should have something to say about one of the four or so prompts. 

I think we're also considering moving the topics from the embed title into the main body/description to not have so much bolded text and to extend the character limit. That way, we wouldn't really need to worry about the limit and can instead just stop at something like 5 prompts.

I don't want to make any changes until we decide which route to go.

hedyhli:
Hey @Anonymous4045, are you still planning to work on this?

I agree with wookie that allowing five topics simultaneously is a little too much, perhaps 3 could do. If a topic is approved in the repo, they should already be considered plausible topics. Keep in mind that people also like to reference the topic embed when replying to the topic. Having too many topics to choose from in that embed can lead to fragmented conversations, which one might say defeats part of the purpose of this `.topic` command in the first place.

Which also means, regarding Chris' suggestion to pop the first topic off and append new ones to the bottom:
Breaking the topic number reference is a problem, like you said. So removing the ability to reroll altogether _on the third[^1] time_ the emoji is reacted seems good enough.

This addresses both Chris' concern (with reaction not actually getting a new topic) and yours (people responding to Topic N).

[^1]: Or whatever the decided limit is.

> What would the title in bold be?

Since titles are not mandatory, moving the topic from the title to the description could look better once the reroll emoji is reacted IMO.

So all that this PR needs now as far as I can tell is to 1) improve the UI (bold text) and 2) implement a reasonable behaviour to limit the number of topic rerolls allowed. What do you think?

Anonymous4045:
Sounds good to me. I can start working on this and see how it feels.

Xithrius:
@Anonymous4045 Hello! Any updates on the progress of this PR? Thanks!

Anonymous4045:
@Xithrius thanks for the reminder! I've implemented most of what you said. I also tried switching the suggestion form to be a footer, but this seems to break the hyperlink. Do you know of a way to have links in the footer? We could also just make it in the description, but I think it looks better as a footer for formatting purposes.
IMG_1315

Xithrius:
Thank you for the update!

Footers in embeds do not support hyperlinks, unfortunately. I believe previously the link was just a part of the description.

Anonymous4045:
@Xithrius bummer. How's this instead?

IMG_1316

Xithrius:
Looks good to me!

PR #1167, opened 2022-12-14
Ignore ellipses when suggesting commands

### LINKED ISSUE #1135: Message prefixed by any length of dot's is responded to
## Description
The bot currently responds to message prefixed with multiple dots that clearly aren't intended as commands, see:
image
This behaviour is unhelpful and takes up space in the chat.

## Expected Behaviour
Any message prefixed with at least three dots should be ignored, this could be changed to 2 dots depending on the frequency of typos. I've said three here since (in my opinion) it's the standard prefix for messages like the one above.

## Possible Solutions
Add a simple check to see if the message begins with more then 2/3 dots, perhaps with `.startswith("...")`.

camcaswell:
Closes #1135 

Adds a check to see if the errored command invocation starts with a period before suggesting commands.

Technically, to match the intention of ignoring messages that start with ".." this assumes that the prefix will remain `.`, but that seems like a safe assumption and it's not a huge problem if we're ignoring messages that look something like `>.` in the far future.


ChrisLovering (approved):
Nice change, always annoying when this happens!

wookie184 (approved):
Looks good, thanks

PR #1176, opened 2022-12-23
Make bookmark a context menu command

### LINKED ISSUE #1089: Allow bookmarking a message from a context menu
We would like to update the existing bookmark command to also be available from the context menu.
#dev-contrib discussion here: https://discord.com/channels/267624335836053506/635950537262759947/1010024284971864104

shtlrs:
**NOTE**
This needs to wait for [this PR](https://github.com/python-discord/bot-core/pull/171) before it can be merged, since it will be the one syncing the command tree.
Once that has been completed & merged, bot-core version needs to be bumped and then merged into main, then this branch will be updated

Closes #1089

This adds the support of bookmarking a message from a context menu, but preserves the old behavior as well.

To not lose the functionality of choosing a custom title, a modal has been added where a user is prompted for a title, which they can freely skip.

# Workflow
1. As usual, a user needs to identify a message to bookmark
image

2. A user hovers on the message, then clicks on the ellipsis (three dots) -> Apps -> Bookmark 
image

3. A user will be asked to choose a custom title for their bookmark, but they can freely ignore that step by clicking on `Submit` and the default title `Bookmark` will be chosen
image

4. Once submitted, the bookmark is sent as usual as a DM to the user, the only difference is that the bookmark embed in the invoked channel is now **ephemeral**
image
image

### DIFF
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -138,7 +138,10 @@ class Bookmark(commands.Cog):
             log.info(f"{member} bookmarked {target_message.jump_url} with title '{title}'")
 
     @staticmethod
-    async def run_permission_check(author: discord.Member | discord.User, channel: discord.TextChannel) -> bool:
+    async def user_is_permitted_to_bookmark(
+            author: discord.Member | discord.User,
+            channel: discord.TextChannel
+    ) -> bool:
         """
         Check if users have the right to bookmark a message in a particular channel.
 
@@ -158,7 +161,7 @@ class Bookmark(commands.Cog):
 
     async def book_mark_context_menu_callback(self, interaction: discord.Interaction, message: discord.Message) -> None:
         """The callback that will be invoked upon using the bookmark's context menu command."""
-        if not await self.run_permission_check(interaction.user, message.channel):
+        if not await self.user_is_permitted_to_bookmark(interaction.user, message.channel):
             return
 
         title = "Bookmark"
@@ -189,7 +192,7 @@ class Bookmark(commands.Cog):
         if target_message is None:
             raise commands.UserInputError(MESSAGE_NOT_FOUND_ERROR)
 
-        if not await self.run_permission_check(ctx.author, target_message.channel):
+        if not await self.user_is_permitted_to_bookmark(ctx.author, target_message.channel):
             return
 
         await self.action_bookmark(ctx.channel, ctx.author, target_message, title)
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -30,15 +30,6 @@ class BookmarkTitleSelectionForm(discord.ui.Modal):
     This form is only available when the command is invoked from a context menu.
     """
 
-    def __init__(
-            self,
-            message: discord.Message,
-            action_bookmark_function: Callable[[discord.TextChannel, discord.Member, discord.Message, str], None],
-    ):
-        super().__init__(timeout=1000, title="Name your bookmark")
-        self.message = message
-        self.action_bookmark = action_bookmark_function
-
     bookmark_title = discord.ui.TextInput(
         label="Choose a title for you bookmark",
         placeholder="Type your bookmark title here",
@@ -47,6 +38,15 @@ class BookmarkTitleSelectionForm(discord.ui.Modal):
         min_length=0,
         required=False
     )
+
+    def __init__(
+            self,
+            message: discord.Message,
+            action_bookmark_function: Callable[[discord.TextChannel, discord.Member, discord.Message, str], None],
+    ):
+        super().__init__(timeout=1000, title="Name your bookmark")
+        self.message = message
+        self.action_bookmark = action_bookmark_function
 
     async def on_submit(self, interaction: discord.Interaction) -> None:
         """Sends the bookmark embed to the user with the newly chosen title."""
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -126,8 +126,8 @@ class Bookmark(commands.Cog):
         """Build the ephemeral embed to the bookmark requester."""
         return discord.Embed(
             description=(
-                f"A bookmark for [this message]({target_message.jump_url})"
-                f"has been successfully sent your way"
+                f"A bookmark for [this message]({target_message.jump_url}) "
+                f"has been successfully sent your way.\n"
                 f"Please check your DMs to retrieve it."
             ),
             colour=Colours.soft_green,
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -31,7 +31,7 @@ class BookmarkTitleSelectionForm(discord.ui.Modal):
     """
 
     bookmark_title = discord.ui.TextInput(
-        label="Choose a title for you bookmark",
+        label="Choose a title for you bookmark (optional)",
         placeholder="Type your bookmark title here",
         default="Bookmark",
         max_length=50,
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -128,7 +128,6 @@ class Bookmark(commands.Cog):
             description=(
                 f"A bookmark for [this message]({target_message.jump_url}) "
                 f"has been successfully sent your way.\n"
-                f"Please check your DMs to retrieve it."
             ),
             colour=Colours.soft_green,
         )
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -73,7 +73,7 @@ class Bookmark(commands.Cog):
         self.bot = bot
         self.book_mark_context_menu = discord.app_commands.ContextMenu(
             name="Bookmark",
-            callback=self.book_mark_context_menu_callback
+            callback=self._bookmark_context_menu_callback
         )
         self.bot.tree.add_command(self.book_mark_context_menu)
 
@@ -159,7 +159,7 @@ class Bookmark(commands.Cog):
             return False
         return True
 
-    async def book_mark_context_menu_callback(self, interaction: discord.Interaction, message: discord.Message) -> None:
+    async def _bookmark_context_menu_callback(self, interaction: discord.Interaction, message: discord.Message) -> None:
         """The callback that will be invoked upon using the bookmark's context menu command."""
         if not await self.user_is_permitted_to_bookmark(interaction.user, message.channel):
             return
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -77,20 +77,17 @@ class Bookmark(commands.Cog):
         )
         self.bot.tree.add_command(self.book_mark_context_menu)
 
-    async def cog_load(self) -> None:
-        """Carry out cog asynchronous initialisation."""
-        await self.bot.tree.sync()
-
     @staticmethod
     def build_bookmark_embed(target_message: discord.Message) -> discord.Embed:
         """Build the channel embed to the bookmark requester."""
-        return discord.Embed(
+        embed = discord.Embed(
             description=(
                 f"Click the button to be sent your very own bookmark to "
                 f"[this message]({target_message.jump_url})."
             ),
             colour=Colours.soft_green,
         )
+        return embed
 
     @staticmethod
     def build_bookmark_dm(target_message: discord.Message, title: str) -> discord.Embed:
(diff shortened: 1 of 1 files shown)


bot/exts/utilities/bookmark.py
```diff
@@ -72,6 +72,10 @@ class Bookmark(commands.Cog):
     def __init__(self, bot: Bot):
         self.bot = bot
 
+    async def cog_load(self) -> None:
```
shtlrs, line 80: For reviewers:
Should we maybe move this to `bot.__main__`?

Robin5605: I think it makes the most sense to only sync the tree either A) on command or B) in `setup_hook` on bot startup

bot/exts/utilities/bookmark.py
```diff
@@ -117,6 +137,40 @@ async def action_bookmark(
         else:
             log.info(f"{member} bookmarked {target_message.jump_url} with title '{title}'")
 
+    @staticmethod
+    async def user_is_permitted_to_bookmark(
```
Robin5605, line 141: I'm not sure this function had to be extracted out into it's own function. It also does more than the name implies, not only does it *check* if a user is permitted to bookmark, but also sends an error message.

I think we should just move the logic back into the context menu callback and original text command, since a method to build the error embed is already there:
```py
if not permissions.read_messages:
    log.info(...)
    error_embed = Bookmark.build_error_embed(...)

    # In the context menu callback...
    interaction.response.send_message(embed=error_embed, ephemeral=True)

    # OR in the text-based command:
    await ctx.send(embed=error_embed)
```
The way we have it now, the interaction would timeout and error because we use `ctx.send` rather than replying to the interaction with `interaction.reply.send_message`

bot/exts/utilities/bookmark.py
```diff
@@ -71,6 +71,26 @@ class Bookmark(commands.Cog):
...
+    @staticmethod
+    def build_bookmark_embed(target_message: discord.Message) -> discord.Embed:
+        """Build the channel embed to the bookmark requester."""
+        return discord.Embed(
+            description=(
+                f"Click the button to be sent your very own bookmark to "
+                f"[this message]({target_message.jump_url})."
+            ),
+            colour=Colours.soft_green,
+        )
```
Robin5605, line 93: For consistency with the other build_* methods
```suggestion
        embed = discord.Embed(
            description=(
                f"Click the button to be sent your very own bookmark to "
                f"[this message]({target_message.jump_url})."
            ),
            colour=Colours.soft_green,
        )
        return embed
```

Robin5605 (changes requested):
Hello, thanks for the PR!
Just a few issues :slightly_smiling_face:

bot/exts/utilities/bookmark.py
shtlrs: That's what I had in mind, in a "general" setup hook for the bot, not per Cog, but I thought `__main__` would be that.
But I just browser and I saw that `Bot` has a setup hook as well, I'll move it there

ChrisLovering: this likely makes more sense in bot core

shtlrs: That's true, I'll move this to a different PR

ChrisLovering: https://github.com/python-discord/bot-core/pull/171

bot/exts/utilities/bookmark.py
shtlrs: Done in dc5fa83c8720a8ea90867705b1b5463b0511d40b

ChrisLovering: I don't understand this suggestion. Some of the other build functions save embed to a param, since they need to add authors or thumbnails etc. However, this and `build_error_embed` don't, so just return the embed directly.

bot/exts/utilities/bookmark.py
```diff
@@ -23,6 +23,40 @@
...
+        self.action_bookmark = action_bookmark_function
+
+    bookmark_title = discord.ui.TextInput(
+        label="Choose a title for you bookmark",
+        placeholder="Type your bookmark title here",
+        default="Bookmark",
+        max_length=50,
+        min_length=0,
+        required=False
+    )
```
Robin5605, line 49: As a rule of thumb we want the `bookmark_title` to come before `__init__`
```suggestion
    bookmark_title = discord.ui.TextInput(
        label="Choose a title for you bookmark",
        placeholder="Type your bookmark title here",
        default="Bookmark",
        max_length=50,
        min_length=0,
        required=False
    )

    def __init__(
            self,
            message: discord.Message,
            action_bookmark_function: Callable[[discord.TextChannel, discord.Member, discord.Message, str], None],
    ):
        super().__init__(timeout=1000, title="Name your bookmark")
        self.message = message
        self.action_bookmark = action_bookmark_function
```

bot/exts/utilities/bookmark.py
```diff
@@ -23,6 +23,40 @@
...
+            self,
+            message: discord.Message,
+            action_bookmark_function: Callable[[discord.TextChannel, discord.Member, discord.Message, str], None],
+    ):
+        super().__init__(timeout=1000, title="Name your bookmark")
+        self.message = message
+        self.action_bookmark = action_bookmark_function
+
+    bookmark_title = discord.ui.TextInput(
+        label="Choose a title for you bookmark",
```
Robin5605, line 43: I'd like to be very clear that this is completely optional. Users would only know this was optional if they knew that required modal inputs had a little red asterisk, which is easy to miss and not something everyone knows about
```suggestion
        label="Choose a title for you bookmark (optional)",
```

Robin5605 (changes requested):

bot/exts/utilities/bookmark.py
```diff
@@ -23,6 +23,40 @@
...
+    ):
+        super().__init__(timeout=1000, title="Name your bookmark")
+        self.message = message
+        self.action_bookmark = action_bookmark_function
+
+    async def on_submit(self, interaction: discord.Interaction) -> None:
+        """Sends the bookmark embed to the user with the newly chosen title."""
+        title = self.bookmark_title.value or self.bookmark_title.default
+        await self.action_bookmark(interaction.channel, interaction.user, self.message, title)
+        view = SendBookmark(self.action_bookmark, interaction.user, interaction.channel, self.message, title)
```
Robin5605, line 55: Since this message is ephemeral, I don't think there's a need for the button that says "Click this to get your very own bookmark" or whatever it was. The whole point of the button was that multiple people wouldn't be running the .bookmark command in succession to the same message and flooding the channel - but with ephemeral messages, that's not an issue. So I think this button would just lead to more confusion and is unnecessary.

bot/exts/utilities/bookmark.py
```diff
@@ -71,6 +105,22 @@ class Bookmark(commands.Cog):
...
+        )
+        self.bot.tree.add_command(self.book_mark_context_menu)
+
+    @staticmethod
+    def build_bookmark_embed(target_message: discord.Message) -> discord.Embed:
+        """Build the channel embed to the bookmark requester."""
+        return discord.Embed(
+            description=(
+                f"Click the button to be sent your very own bookmark to "
+                f"[this message]({target_message.jump_url})."
```
Robin5605, line 120: See my comment above. I don't think this description is necessary either. We just need something like "Bookmark sent! Check your DMs" or something similar

Robin5605 (changes requested):

bot/exts/utilities/bookmark.py
```diff
@@ -23,6 +23,40 @@
 )
 
 
+class BookmarkTitleSelectionForm(discord.ui.Modal):
+    """
+    The form where a user can fill in a custom title for their bookmark & submit it.
```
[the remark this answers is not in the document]
shtlrs: PEP 257 says that "The summary line may be on the same line as the opening quotes or on the next line." 

But Flake8 doesn't seem to like it , and it seems they've chosen to force it on a separate line.
image

bot/exts/utilities/bookmark.py
shtlrs: Makes sense, done in cf4903561a222de9b7413c406b5d6e186f977332

bot/exts/utilities/bookmark.py
shtlrs: This one has also been resolved in cf4903561a222de9b7413c406b5d6e186f977332

I just happened to make a new static method that builds an embed that's only used when the context menu command is invoked

shtlrs (commented):
Alright, comments have been addressed and it's ready for review again

wookie184 (changes requested):
Played around with the command and noticed a few things:

- The command didn't seem to sync automatically, I had to run ``.int eval `await bot.tree.sync()` `` for it to appear.
- The first time I use the command, this seems to stay there forever:
image
For bookmarks run after that it disappears normally, although after refreshing the client the same thing happens again.
- Running the command in DMs with the bot gives a message saying you don't have permission to view that channel
image
It's pretty minor, and there's no point in allowing using the command in DMs, but the error message could be better.

Not sure if any of these are discord bugs rather than issues with the code, although worth looking into.

bot/exts/utilities/bookmark.py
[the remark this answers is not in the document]
wookie184: Our style guide specifies to put it on a new line https://www.pythondiscord.com/pages/guides/pydis-guides/contributing/style-guide/#line-structure

I like it more that way too.

bot/exts/utilities/bookmark.py
```diff
@@ -71,6 +104,34 @@ class Bookmark(commands.Cog):
...
+        )
+
+    @staticmethod
+    def build_ephemeral_bookmark_embed(target_message: discord.Message) -> discord.Embed:
+        """Build the ephemeral embed to the bookmark requester."""
+        return discord.Embed(
+            description=(
+                f"A bookmark for [this message]({target_message.jump_url}) "
+                f"has been successfully sent your way.\n"
+                f"Please check your DMs to retrieve it."
```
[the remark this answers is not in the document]
shtlrs: Good point, I've removed the call to action for the DMS

shtlrs:
> * The command didn't seem to sync automatically, I had to run `` .int eval `await bot.tree.sync()`  `` for it to appear.

I think you might have missed the description, this relies on the newest release of pydis_core, where the tree is synced upon loading extensions. So that might be the reason.

> * The first time I use the command, this seems to stay there forever:
>   image
>   For bookmarks run after that it disappears normally, although after refreshing the client the same thing happens again.

This has also happened to me, but since it threw no errors, i thought it could be a discord bug or something.
I couldn't seem to figure out why since, like you said, it only happens the first time.

> * Running the command in DMs with the bot gives a message saying you don't have permission to view that channel
>   image
>   It's pretty minor, and there's no point in allowing using the command in DMs, but the error message could be better.
Well, we could check if the interaction channel is part of a guild or not, if not then it's a DM --> we change the message
But the thing is, we are completely overlooking the `dm_permissions` method here, which does include read_permissions.

So I think it's just better to agree on not allowing the usage of such command in DMs, and if that's the case, we send a message saying so.

wookie184:
> I think you might have missed the description, this relies on the newest release of pydis_core, where the tree is synced upon loading extensions. So that might be the reason.

I noticed that but since the PR was merged I assumed `poetry install` would do fine. Is it just that the version used by this project needs to be bumped now?

wookie184:
Looks like there are some commits from other PRs that shouldn't be here

shtlrs:
> Looks like there are some commits from other PRs that shouldn't be here

I think those are the results of the rebase ?

wookie184 (approved):
Also seems like the hanging command problem isn't there for me either any more👍

ChrisLovering (approved):

bot/exts/utilities/bookmark.py
```diff
@@ -23,6 +23,39 @@
...
 
+class BookmarkTitleSelectionForm(discord.ui.Modal):
+    """
+    The form where a user can fill in a custom title for their bookmark & submit it.
+
+    This form is only available when the command is invoked from a context menu.
+    """
+
+    bookmark_title = discord.ui.TextInput(
+        label="Choose a title for you bookmark (optional)",
```
onerandomusername, line 34:
```suggestion
        label="Choose a title for your bookmark (optional)",
```

ChrisLovering: Fixed in https://github.com/python-discord/sir-lancebot/pull/1211/commits/5d0ac27ed0432359ee339382255eeb71a88f4b81 (https://github.com/python-discord/sir-lancebot/pull/1211)

PR #1180, opened 2022-12-27
Bump botcore version

shtlrs:
Bumps botcore from 8.2.1 to 9.3.1 (now called pydis-core)


ChrisLovering (approved):
Looks good to me! 9.3.1 is available on pypi now too. Feel free to bump to that version, consider my approval is still valid if you do.

Xithrius (approved):
yep

PR #1197, opened 2023-01-29
Add possibility of deleting a bm with a button

### LINKED ISSUE #1196: Deleting bookmarks
I have noticed that sir lancebot only allows us to add bookmarks, but not deleting them. 

I figured that this could be useful for whatever reasons people would have w be that the DM channel is too cluttered, or they have bookmarked some message explaining Middlewares for example but then have read a better one which they'd like to replace with the old one. 

This should be a pretty easy change considering we'd use add another button to the view we already send in the DMs. 

shtlrs:
Closes #1196

The change is pretty simple, and makes it more intuitive to delete a book mark start from the DMs.

I'll keep this in draft for now since the original issue hasn't been approved, but I thought i'd do it still since it's only 4 lines of code


ChrisLovering:
This might be better served with another application command, similar to how we changed the main bookmark command to one.

We can also deprecate the existing text command that way too.

If we make this delete command a global command, we can remove perms for people to run it in-server so it will only appear in DMs.

shtlrs:
> we can remove perms for people to run it in-server so it will only appear in DMs.

Humm I see, just like the perms issue that happened for some people with the new bookmark command? That's a good idea.

ChrisLovering:
> > we can remove perms for people to run it in-server so it will only appear in DMs.
> 
> Humm I see, just like the perms issue that happened for some people with the new bookmark command? That's a good idea.

At a per command level you can control what users have permission to run it 
image

shtlrs:
That's neat.

Alright, sounds pretty good!

ChrisLovering:
> That's neat.
> 
> Alright, sounds pretty good!

You can set it in-app using [this](https://discordpy.readthedocs.io/en/stable/interactions/api.html#discord.app_commands.ContextMenu.default_permissions)
> Setting an empty permissions field will disallow anyone except server administrators from using the command in a guild.

bot/exts/utilities/bookmark.py
```diff
@@ -105,6 +109,26 @@ def build_error_embed(message: str) -> discord.Embed:
             colour=Colours.soft_red,
         )
 
+    async def _delete_bookmark_context_menu_callback(
+            self,
+            interaction: discord.Interaction,
+            message: discord.Message
+    ) -> None:
```
ChrisLovering, line 116: his shouldn't be double-indented (maybe add a trailing comma while you're editing this too)

ChrisLovering (changes requested):

bot/exts/utilities/bookmark.py
shtlrs: Done in 6047ad7165a16ae499d95766ad3dfc20a64c5a63

bot/exts/utilities/bookmark.py
```diff
@@ -105,6 +109,26 @@ def build_error_embed(message: str) -> discord.Embed:
...
+        """
+        Delete the Sir-Lancebot message that the command invocation is replying to.
+
+        This command allows deleting any message sent by Sir-Lancebot in the user's DM channel with the bot.
+        The command invocation must be a reply to the message that is to be deleted.
```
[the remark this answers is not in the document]
shtlrs: Updated in a6fe30adc6aa611e042006659f4dcc8aa135b7b9

bot/exts/utilities/bookmark.py
```diff
@@ -73,7 +72,12 @@ def __init__(self, bot: Bot):
             name="Bookmark",
             callback=self._bookmark_context_menu_callback,
         )
+        self.delete_book_mark_context_menu = discord.app_commands.ContextMenu(
+            name="Delete bookmark",
+            callback=self._delete_bookmark_context_menu_callback,
+        )
```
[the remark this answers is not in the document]
shtlrs: The docs say:
> Setting an empty permissions field will disallow anyone except server administrators from using the command in a guild.

By default, upon instantiation, it defaults to none since the callback doesn't have the `__discord_app_commands_default_permissions__` attribute, so I'm wondering if that counts as the same ?

If not, i'll just add this then
```py
self.delete_book_mark_context_menu.default_permissions = []
```

bot/exts/utilities/bookmark.py
```diff
@@ -105,6 +109,25 @@ def build_error_embed(message: str) -> discord.Embed:
...
+        """
+        if interaction.channel.type != discord.ChannelType.private:
+            raise commands.UserInputError("You can only run this command your own DMs!")
+        elif message.channel != interaction.channel:
+            raise commands.UserInputError("You can only delete messages in your own DMs!")
+        elif message.author != self.bot.user:
+            raise commands.UserInputError("You can only delete messages sent by Sir Lancebot!")
+
+        await message.delete()
+
```
Xithrius, line 130: After the message is deleted, I get the response below.
image

Might just be me, but this could be worrying to a user, thinking the bot broke (I had thought this lol). If this is a convention for the rest of the bot (I haven't worked on this project in a bit), please let me know. In the meantime, here's a suggestion (probably best to put in a builder method like the others):

```suggestion
        await interaction.response.send_message(
            embed=discord.Embed(
                description="Bookmark successfully deleted.",
                colour=Colours.soft_green,
            ),
            ephemeral=True,
        )
```
        
        
image

Xithrius (changes requested):
Just a singular comment. Besides that, original functionality is in place, very nice.

Xithrius (approved):
yuh

PR #1211, opened 2023-02-19
Deprecate bookmark text command

ChrisLovering:
## Description

This deprecates the text-based bookmark command, now that we have the context menu command.

Any users that invoke the old text-based command will be told how to use the new context menu command in sir-lance-playground.

This also simplifies a lot of the implementation of the bookmark cog.

I suggest reviewing commit-by-commit as some code has been moved around.

## Did you:

### DIFF
--- a/bot/exts/utilities/bookmark.py
+++ b/bot/exts/utilities/bookmark.py
@@ -106,9 +106,9 @@ class Bookmark(commands.Cog):
         self.bot = bot
         self.book_mark_context_menu = discord.app_commands.ContextMenu(
             name="Bookmark",
-            callback=self._bookmark_context_menu_callback
+            callback=self._bookmark_context_menu_callback,
         )
-        self.bot.tree.add_command(self.book_mark_context_menu)
+        self.bot.tree.add_command(self.book_mark_context_menu, guild=discord.Object(bot.guild_id))
 
     @staticmethod
     def build_bookmark_embed(target_message: discord.Message) -> discord.Embed:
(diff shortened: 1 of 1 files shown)


bot/exts/utilities/bookmark.py
```diff
@@ -153,85 +102,39 @@ def build_error_embed(message: str) -> discord.Embed:
...
+        response = await ctx.guild.get_channel(Channels.sir_lancebot_playground).send(
+            f"{ctx.author.mention} The bookmark text command has been replaced with a context menu command!\n\n"
+            "To bookmark a message simply right-click (press and hold on mobile) "
+            "on a message, open the 'Apps' menu, and click 'Bookmark'."
+        )
+        await ctx.send(
+            embed=self.build_error_embed(
+                f"See [this message]({response.jump_url}) for information on how to bookmark messages."
+            )
+        )
```
wookie184, line 137: The message redirecting the user is about the same size as the information message itself, so I feel it would make more sense to just send the information directly to the channel where it's invoked- it doesn't seem too noisy.

(There's also currently a slightly weird case currently when using the command in `sir_lancebot_playground` as it sends the info message and then an embed directly below linking to the message above.

ChrisLovering: Yea I like this suggestion, pushed the change.

wookie184 (approved):
LGTM!

shtlrs (approved):
Byoutifoul !

PR #1217, opened 2023-02-25
Add default unicode emoji for trashcan

Ibrahim2750mi:
https://github.com/python-discord/site/pull/882
@wookie184

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -132,10 +132,10 @@ class Emojis:
     check = "\u2611"
     envelope = "\U0001F4E8"
 
-    trashcan_default = "<:trashcan:637136429717389331>"
-    if Client.debug:
-        trashcan_default = "\U0001f5d1\ufe0f"
-    trashcan = environ.get("TRASHCAN_EMOJI", trashcan_default)
+    trashcan = environ.get(
+        "TRASHCAN_EMOJI",
+        f"\N{WASTEBASKET}" if Client.debug else "<:trashcan:637136429717389331>",
+    )
 
     ok_hand = ":ok_hand:"
     hand_raised = "\U0001F64B"
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -134,7 +134,7 @@ class Emojis:
 
     trashcan = environ.get(
         "TRASHCAN_EMOJI",
-        f"\N{WASTEBASKET}" if Client.debug else "<:trashcan:637136429717389331>",
+        "\N{WASTEBASKET}" if Client.debug else "<:trashcan:637136429717389331>",
     )
 
     ok_hand = ":ok_hand:"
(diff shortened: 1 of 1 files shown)


wookie184:
We still want to use the custom one on the server, so this would require setting the env variable in production.

An alternative approach would be to use the `Client.debug` constant, which defaults to true, but is set to false in production. If this is true we would use the unicode default if the env var wasn't set (this wouldn't require any changes to production environment variables)

Do @python-discord/devops have a preference?

Ibrahim2750mi:
> An alternative approach would be to use the `Client.debug` constant, which defaults to true, but is set to false in production. If this is true we would use the unicode default if the env var wasn't set (this wouldn't require any changes to production environment variables)

I don't see `Client.debug` being false anywhere, if anything, we need to set it false by ourselves. Am I right? Or does this happen outside `constants.py`

wookie184:
The `BOT_DEBUG` environment variable is set in production, production environment variables can only be set by a member of the devops team. Since it's already set that's why I suggested using it to determine what default to use for the reaction (keeping the reaction production constant here also makes it easier to change in the future if necessary).

bot/constants.py
```diff
@@ -131,7 +131,12 @@ class Emojis:
     cross_mark = "\u274C"
     check = "\u2611"
     envelope = "\U0001F4E8"
-    trashcan = environ.get("TRASHCAN_EMOJI", "<:trashcan:637136429717389331>")
+
+    trashcan_default = "<:trashcan:637136429717389331>"
+    if Client.debug:
+        trashcan_default = "\U0001f5d1\ufe0f"
+    trashcan = environ.get("TRASHCAN_EMOJI", trashcan_default)
```
wookie184, line 138:
```suggestion
    trashcan = environ.get(
        "TRASHCAN_EMOJI",
        f"\N{WASTEBASKET}" if Client.debug else "<:trashcan:637136429717389331>",
    )
```
Normally an if would probably be nicer, but here ideally we should avoid any extra variables being left over in the namespace so people don't accidentally use them in code or get confused. (alternatively we could prefix the default with a `_` and `del` it afterwards)

wookie184:
Sorry, I forgot that isn't a f-string thing, can just remove the f, my bad 😅

wookie184:
Hmm wait i'm not even sure if it's working without, let me try it

Ibrahim2750mi:
> Hmm wait i'm not even sure if it's working without, let me try it

okay, I will be waiting.

wookie184:
> > Hmm wait i'm not even sure if it's working without, let me try it
> 
> okay, I will be waiting.

Never mind, just fixing the lint error should be fine. I just had `BOT_DEBUG=false` in my `.env` for some reason.

wookie184 (approved):
Thanks!

shenanigansd (approved):
:shipit:

PR #1218, opened 2023-02-26
fixed embed interaction and randomly choosing a kata

### LINKED ISSUE #1170: Fixes for challenge command
Running the `.challenge` command. the following issues occur:
-  Interactions edit the message but still show as failed
image
- The same challenge seems to be shown every time for a given query, where it should be random.

## Possible Solutions

I think the failed interaction thing should be fairly simple to fix, I haven't looked at the code though.

The same challenge being shown could be due to a change in the behaviour of the codewars search. It seems there's a "random sample" option on the page 
image
so we may need to use `sample=true` in the query? I haven't tested this.

henri-climber:
## Relevant Issues
Closes #1170 

## Description

The issue mentioned two problems:

1. interactions edit the message but still show as failed
The problem is that discord automatically says the interaction failed if the bot doesn't respond to the interaction. Adding interaction.response.defer() fixes this. (See this stackoverflow discussion: https://stackoverflow.com/questions/71177867/discord-py-error-message-this-interaction-failed)

2. The same challenge seems to be shown every time for a given query, where it should be random.
I believe this happened because of the if statements that chose the kata_id. Whenever a query resulted in more than 3 katas, the first elif statement was called, resulting in the same 3 katas being displayed over and over. My solution is to remove the elif statements and always make a random selection from all the katas found.

## Did you:

### DIFF
--- a/bot/exts/utilities/challenges.py
+++ b/bot/exts/utilities/challenges.py
@@ -84,6 +84,7 @@ class InformationDropdown(ui.Select):
         # The attribute is not set during initialization.
         result_embed = self.mapping_of_embeds[self.values[0]]
         await self.original_message.edit(embed=result_embed)
+        await interaction.response.defer()
 
 
 class Challenges(commands.Cog):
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
LGTM, thank you!

PR #1229, opened 2023-03-09
Correctly handle when the movie API doesn't return any movies

### LINKED ISSUE #1121: TypeError: object of type 'NoneType' has no len()
Sentry Issue: [SIR-LANCEBOT-8R](https://sentry.io/organizations/python-discord/issues/3673171895/?referrer=github_integration)

```
TypeError: object of type 'NoneType' has no len()
  File "discord/ext/commands/core.py", line 190, in wrapped
  File "bot/exts/holidays/halloween/scarymovie.py", line 23, in random_movie
    selection = await self.select_movie()
  File "bot/exts/holidays/halloween/scarymovie.py", line 50, in select_movie
    selection_id = random.choice(data.get("results")).get("id")
  File "random.py", line 378, in choice
    return seq[self._randbelow(len(seq))]

Unhandled command error: object of type 'NoneType' has no len()
```

TizzySaurus:
## Relevant Issues

Closes #1121 

## Description

Changes code so that it handles when the two `.get()` calls return `None`, sending an error to the command invoke and also logging a warning.

Also limits the selected page to get the movie from to 500, since the API has that as a limit.

## Did you:


bot/exts/holidays/halloween/scarymovie.py
```diff
@@ -44,10 +51,16 @@ async def select_movie(self) -> dict:
             total_pages = data.get("total_pages")
 
         # Get movie details from one random result on a random page
-        params["page"] = random.randint(1, total_pages)
+        params["page"] = max(random.randint(1, total_pages), 500)
```
wookie184, line 54: Should be `min`, no?

wookie184 (changes requested):

wookie184 (approved):
LGTM 👍

bot/exts/holidays/halloween/scarymovie.py
```diff
@@ -44,10 +51,16 @@ async def select_movie(self) -> dict:
             total_pages = data.get("total_pages")
 
         # Get movie details from one random result on a random page
-        params["page"] = random.randint(1, total_pages)
+        params["page"] = min(random.randint(1, total_pages), 500)
```
mbaruh, line 54: While this works the result is no longer random for some inputs. Wouldn't something like this be better?
```suggestion
        params["page"] = random.randint(1, min(total_pages, 500))
```

bot/exts/holidays/halloween/scarymovie.py
```diff
@@ -21,6 +21,13 @@ async def random_movie(self, ctx: commands.Context) -> None:
         """Randomly select a scary movie and display information about it."""
         async with ctx.typing():
             selection = await self.select_movie()
+            if not selection:
+                await ctx.send(embed=Embed(
+                    title=random.choice(NEGATIVE_REPLIES),
+                    description=":warning: Failed to select a movie from the API",
+                    color=Colours.soft_red
+                ))
```
mbaruh, line 29: It seems like there's an APIError class that can be raised which is handled by the error handler. I think it would fit here.

mbaruh (changes requested):

bot/exts/holidays/halloween/scarymovie.py
TizzySaurus: The APIError exception (as per the docstring) is for when "an external API returns an error message". Since this isn't an error returned by the API, I didn't opt to use the exception. 

If we do want to use it, what should the `status_code` argument be? 200? The [error handling](https://github.com/python-discord/sir-lancebot/blob/main/bot/exts/core/error_handler.py#L127-L134) for it also seems pretty bare (doesn't actually use the `.error_msg` or `.status_code`).

mbaruh: I see. Nevermind then.

mbaruh (approved):

PR #1230, opened 2023-03-10
Allow .latex in python-help channel

matiaslagoevia:
## Relevant Issues
Closes #1207.

## Description
Adds the `#python-help` channel to the list of channels where `.latex` command is allowed.

## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] Read all the comments in this template?
- [x] Ensure there is an issue open, or link relevant discord discussions?
- [x] Read and agree to the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

PR #1231, opened 2023-03-10
Whitelist a thread when its parent is whitelisted

shtlrs:
Currently, we can't execute commands in threads if they're not whitelisted.
The problem is, we can't have a thread id in advance.

This adds the possibility to do that when the parent of the thread is whitelisted (Just like the case of help forum channels)

**Context**:
This is needed in order of achieving #1230

### DIFF
--- a/bot/utils/checks.py
+++ b/bot/utils/checks.py
@@ -3,6 +3,7 @@ import logging
 from collections.abc import Container, Iterable
 from typing import Callable, Optional
 
+from discord import Thread
 from discord.ext.commands import (
     BucketType, CheckFailure, Cog, Command, CommandOnCooldown, Context, Cooldown, CooldownMapping
 )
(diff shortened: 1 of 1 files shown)


bot/utils/checks.py
```diff
@@ -60,6 +61,10 @@ def in_whitelist_check(
         # categories, it's probably not wise to rely on its category in any case.
         channels = tuple(channels) + (redirect,)
 
+    # If it's a thread, and its parent is whitelisted, we whitelist the thread as well
+    if channels and isinstance(ctx.channel, Thread) and ctx.channel.parent_id in channels:
```
mbaruh, line 65: Rather than adding the thread to the whitelisted channels, it makes more sense to check the parent channel instead of ctx.channel in this case. This also keeps the `in channels` check in one place.

mbaruh: fwiw I usually do a hasattr check whenever I have to do something like this, and save the extra import, but that's more a matter of style.

shtlrs: I just wanted to avoid multiple ifs within each other, but i can do that if you want

mbaruh (approved):
Thanks

Ibrahim2750mi (approved):

PR #1260, opened 2023-05-01
Use pydantic for configuration management

shtlrs:
To align better with the [recent change](https://github.com/python-discord/bot/pull/2408) that [bot](https://github.com/python-discord/bot) has witnessed, we've decided to also migrate the rest of the bot to use the same schema/package.

> Note: Some classes are not configurable, so they were left as is, but that can be easily changed if we want.

The following is a list of all env variables we'd need to update @python-discord/devops 

```text
CHANNEL_DEVLOG -> channels_devlog
CHANNEL_COMMUNITY_BOT_COMMANDS -> channels_sir_lancebot_playground
CHANNEL_REDDIT -> channels_reddit
PREFIX -> client_prefix
FILE_LOGS -> logging_file_logs
BOT_TRACE_LOGGERS -> logging_trace_loggers
BOT_ADMIN_ROLE_ID -> roles_admin
ROLE_HELPERS -> roles_helpers


GIPHY_TOKEN -> tokens_giphy
youtube -> tokens_youtube
TMDB_API_KEY -> TMDB_API_KEY
NASA_API_KEY -> tokens_nasa
IGDB_CLIENT_ID -> tokens_igdb_client_id
IGDB_CLIENT_SECRET -> tokens_igdb_client_secret
GITHUB_TOKEN -> tokens_github
UNSPLASH_KEY -> tokens_unsplash
WOLFRAM_API_KEY -> wolfram_key


USE_FAKEREDIS -> redis_use_fakeredis
```


HassanAbouelela:
Chiming in for devoops. Aside from the tokens, all the other constants should be given defaults within the app itself.

shtlrs:
> Chiming in for devoops. Aside from the tokens, all the other constants should be given defaults within the app itself.

Which constants are lacking defaults here ? Aside from the tokens, the `Reddit`, `Redis`, `Logging` are the ones that have attributes without defaults, and they're all secrets, except for the `Logging` one.

HassanAbouelela:
I’m going off the content of the PR message. If those already have constants added, this PR doesn’t need to do anything further.

shtlrs:
> I’m going off the content of the PR message. If those already have constants added, this PR doesn’t need to do anything further.

Ah, those need to be updated in our pods/cluster. (I want to assist when it's going to be done)

HassanAbouelela:
I see, you're welcome to take this on. The production environment only contains tokens and other secret values, no channels or roles.

bot/constants.py
```diff
@@ -36,62 +36,96 @@
...
     name = "Sir Lancebot"
-    guild = int(environ.get("BOT_GUILD", 267624335836053506))
-    prefix = environ.get("PREFIX", ".")
-    token = environ.get("BOT_TOKEN")
-    debug = environ.get("BOT_DEBUG", "true").lower() == "true"
-    in_ci = environ.get("IN_CI", "false").lower() == "true"
-    github_bot_repo = "https://github.com/python-discord/sir-lancebot"
+    guild = 267624335836053506
+    prefix = "."
+    token = ""
```
ChrisLovering, line 109:
```suggestion
    token: str
```
Make this required to be passed in

bot/constants.py
```diff
@@ -236,39 +266,61 @@ def __str__(self) -> str:
...
+    aoc_session_cookie = ""
+    omdb = ""
+    youtube = ""
+    tmdb = ""
+    nasa = ""
+    igdb_client_id = ""
+    igdb_client_secret = ""
+    github = ""
+    unsplash = ""
+    wolfram = ""
```
ChrisLovering, line 299: Should these be `foo: str | None` instead? Maybe even pydantic secret strs?

bot/constants.py
```diff
@@ -236,39 +266,61 @@ def __str__(self) -> str:
...
+
+Wolfram = _Wolfram()
+
+
+class _Redis(EnvConfig):
+    EnvConfig.Config.env_prefix = "redis_"
+
+    host = "redis.default.svc.cluster.local"
+    port = 6379
+    password = ""
```
ChrisLovering, line 319: Should this be a pydantic secret str?

bot/constants.py
```diff
@@ -280,13 +332,17 @@ class RedirectOutput:
...
+class _Reddit(EnvConfig):
+    EnvConfig.Config.env_prefix = "reddit_"
+
     subreddits = ["r/Python"]
 
-    client_id = environ.get("REDDIT_CLIENT_ID")
-    secret = environ.get("REDDIT_SECRET")
-    webhook = int(environ.get("REDDIT_WEBHOOK", 635408384794951680))
+    client_id = ""
+    secret = ""
```
ChrisLovering, line 341: Should these be `foo: str | None` instead? Maybe even pydantic secret strs?

ChrisLovering (changes requested):
More general question, do we actually need `dotenv` as a package too?

docker compose loads from the .env files already, and so does `poetry run ...`

shtlrs:
> More general question, do we actually need `dotenv` as a package too?
> 
> docker compose loads from the .env files already, and so does `poetry run ...`

Hum, good question. Since docker loads them as env variables, there shouldn't be a need to include it. We'll test it out tonight.

bot/constants.py
shtlrs: We could make them a pydantic's `SecretStr`, but it will require calling `get_secret_value` for all of the secrets that we have.
I don't mind having that, it's just that it might not be that obvious to call that method whenever we use secrets, but we can enforce it during reviews i guess.

bot/constants.py
shtlrs: I'll make them nullable.
As for the `SecretStr`, see my comment [here](https://github.com/python-discord/sir-lancebot/pull/1260#discussion_r1185941910)

HassanAbouelela:
It’s still valuable for anyone running outside poetry, is there a downside to including it?

ChrisLovering:
> It’s still valuable for anyone running outside poetry, is there a downside to including it?

Dear Scaleios,

Thank you for reaching out to us regarding running our bot. Please be advised that we do not support running our bot outside of either Poetry or Docker.

Poetry is a tool for dependency management and packaging in Python that we use to manage our bot's dependencies. Docker, on the other hand, is a containerization platform that allows us to package our bot and all its dependencies into a single container for easy deployment.

While we understand that there may be other ways to run our bot, we strongly recommend using either Poetry or Docker to ensure that all dependencies are properly managed and to minimize any potential issues that may arise from running the bot outside of these environments.

If you have any further questions or concerns, please don't hesitate to let us know.

bot/exts/events/hacktoberfest/hacktober-issue-finder.py
```diff
@@ -18,7 +18,7 @@
     "User-Agent": "Python Discord Hacktoberbot",
     "Accept": "application / vnd.github.v3 + json"
 }
-if GITHUB_TOKEN := Tokens.github:
+if GITHUB_TOKEN := Tokens.github.get_secret_value():
     REQUEST_HEADERS["Authorization"] = f"token {GITHUB_TOKEN}"
```
ChrisLovering, line 22: Won't this error if `Tokens.github` is None? We should probably do something like

```suggestion
if Tokens.github:
    REQUEST_HEADERS["Authorization"] = f"token {Tokens.github.get_secret_value()}"
```

This also applies for all the other places with a similar pattern.

bot/constants.py
```diff
@@ -236,39 +269,59 @@ def __str__(self) -> str:
...
-    unsplash_access_key = environ.get("UNSPLASH_KEY")
+    giphy: SecretStr | None
+    youtube: SecretStr | None
+    tmdb: SecretStr | None
+    nasa: SecretStr | None
+    igdb_client_id: SecretStr | None
+    igdb_client_secret: SecretStr | None
+    github: SecretStr | None
+    unsplash: SecretStr | None
+    wolfram: SecretStr | None
```
ChrisLovering, line 300: Why has wolfram been moved here if we have a specific `_Wolfram` class?

ChrisLovering (changes requested):
Just two more comments.

We should also prepare an announcement with instructions on what contribs need to change in their .env files before merging this.

ChrisLovering (approved):

PR #1278, opened 2023-05-17
Add support for latex rendering server >400 errors

### LINKED ISSUE #1146: Handle Latex API Errors
Sentry Issue: [SIR-LANCEBOT-90](https://sentry.io/organizations/python-discord/issues/3733301843/?referrer=github_integration)

We should better handle the case when the API itself is having internal issues (class 5xx of responses), and return an appropriate response to the user telling them to try again later. Here is the error handling for this function, it simply needs to catch and handle the 5xx type exceptions:
https://github.com/python-discord/sir-lancebot/blob/462844b874099cb4403e8a331add3631a5cce5a3/bot/exts/fun/latex.py#L82-L83

```
ClientResponseError: 502, message='Bad Gateway', url=URL('https://rtex.probablyaweb.site/api/v2')
(1 additional frame(s) were not displayed)
...
  File "bot/exts/fun/latex.py", line 119, in latex
    await self._generate_image(TEMPLATE.substitute(text=query), out_file)
  File "bot/exts/fun/latex.py", line 80, in _generate_image
    async with self.bot.http_session.post(LATEX_API_URL, data=payload, raise_for_status=True) as response:
  File "aiohttp/client.py", line 1138, in __aenter__
    self._resp = await self._coro
  File "aiohttp/client.py", line 640, in _request
    resp.raise_for_status()
  File "aiohttp/client_reqrep.py", line 1004, in raise_for_status
    raise ClientResponseError(

Unhandled command error: 502, message='Bad Gateway', url=URL('https://rtex.probablyaweb.site/api/v2')
```

MedAliMarz:
## Relevant Issues & Changes
Closes #1146 

I've added a catch for `ClientResponseError`, raised a `LatexServerError` and handled the later to return a user message, with a tiny refactoring of the code that creates the embed (in case of error)

## Did you:


bot/exts/fun/latex.py
```diff
@@ -81,8 +88,11 @@ def __init__(self, bot: Bot):
     async def _generate_image(self, query: str, out_file: BinaryIO) -> None:
         """Make an API request and save the generated image to cache."""
         payload = {"code": query, "format": "png"}
-        async with self.bot.http_session.post(LATEX_API_URL, data=payload, raise_for_status=True) as response:
-            response_json = await response.json()
+        try:
```
shtlrs, line 91: I'm wondering whether we should keep `raise_for_status` here since in this particular case, we lose access to the `response.json()` call, which contains the logs (per the docs)
image

But i don't think it would be necessary since it's a server-side error and we can't make much out of it, I was just thinking out loud.

bot/exts/fun/latex.py
```diff
@@ -105,6 +115,16 @@ async def _upload_to_pastebin(self, text: str) -> str | None:
         except web.HTTPClientError as e:
             log.info("Error when uploading latex output to pastebin. %s", e)
 
+    async def _prepare_error_embed(self, title: str, err: Exception | None) -> discord.Embed:
+        embed = discord.Embed(title=title)
+        embed.description = "No logs available."
+        if err is not None and err.logs is not None:
```
shtlrs, line 121: You could simplify this to
```suggestion
        if hasattr(err, "logs")
```

MedAliMarz: Thanks for the suggestion :), I gave it a try, yet this will only check if the attr exist and not its nullability
So now I use getattr (which is used internally by hasattr, https://docs.python.org/3/library/functions.html#hasattr)

bot/exts/fun/latex.py
MedAliMarz: based on the server code here https://github.com/DXsmiley/rtex/blob/master/src/jobs.py#L38
I think all potential rendering issue are handled, therefore no logs will be provided in the case of 5xx responses; so keeping `raise_for_status` seems logical in this case

What do you think ?

shtlrs: Yeah, i shouldn't have spoken out loud in the first place :P

The thing is, even if we have server logs, there's nothing we can do about it since we're only the client. (Unless you want to make a PR to the other repo to fix it :P)

So yes, let's keep it as is, since there's not much value to us here

shtlrs (changes requested):

bot/exts/fun/latex.py
```diff
@@ -121,15 +143,12 @@ async def latex(self, ctx: commands.Context, *, query: str) -> None:
                     with open(image_path, "wb") as out_file:
                         await self._generate_image(TEMPLATE.substitute(text=query), out_file)
                 except InvalidLatexError as err:
```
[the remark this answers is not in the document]
MedAliMarz: Both logical and lovely comment, applied with using tuple exceptions (pre-commit wasn't happy :') )

bot/exts/fun/latex.py
```diff
@@ -105,6 +116,17 @@ async def _upload_to_pastebin(self, text: str) -> str | None:
         except web.HTTPClientError as e:
             log.info("Error when uploading latex output to pastebin. %s", e)
 
+    async def _prepare_error_embed(self, title: str, err: Exception | None) -> discord.Embed:
+        embed = discord.Embed(title=title)
+        embed.description = "No logs available."
+        logs = getattr(err, "logs", None)
+        if logs:
+            logs_paste_url = await self._upload_to_pastebin(err.logs)
```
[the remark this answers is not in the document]
MedAliMarz: Fair enough, applied!

bot/exts/fun/latex.py
```diff
@@ -72,6 +72,14 @@ def __init__(self, logs: str | None):
         self.logs = logs
 
 
+class LatexServerError(Exception):
+    """Represents an error raised from Latex rendering server."""
+
+    def __init__(self, logs: str | None):
```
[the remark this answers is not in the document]
MedAliMarz: Fair comment, it indeed unneeded param, removed

bot/exts/fun/latex.py
```diff
@@ -105,6 +113,21 @@ async def _upload_to_pastebin(self, text: str) -> str | None:
         except web.HTTPClientError as e:
             log.info("Error when uploading latex output to pastebin. %s", e)
 
+    async def _prepare_error_embed(self, err: Exception | None) -> discord.Embed:
```
shtlrs, line 116: The type hinting on this one isn't quite valid since you're only catching the two custom latex exceptions. 

```suggestion
    async def _prepare_error_embed(self, err: InvalidLatexError | LatexServerError |  None) -> discord.Embed:
```

shtlrs (changes requested):
Looks great ! Just one last comment.

bot/exts/fun/latex.py
MedAliMarz: Indeed, my bad :')

shtlrs (approved):
LGTM, thank you !

Xithrius (approved):
LGTM as well. Thanks for the PR!

PR #1296, opened 2023-06-12
Disable the r/python auto poster

ChrisLovering:
Otherwise this will error every morning while r/python isn't accessible

## Did you:

- [ ] Join the [**Python Discord Community**](https://discord.gg/python)?
- [ ] Read all the comments in this template?
- [ ] Ensure there is an issue open, or link relevant discord discussions?
- [ ] Read and agree to the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

### DIFF
--- a/bot/exts/utilities/reddit.py
+++ b/bot/exts/utilities/reddit.py
@@ -37,7 +37,7 @@ class Reddit(Cog):
         self.access_token = None
         self.client_auth = BasicAuth(RedditConfig.client_id.get_secret_value(), RedditConfig.secret.get_secret_value())
 
-        self.auto_poster_loop.start()
+        # self.auto_poster_loop.start()
 
     async def cog_unload(self) -> None:
         """Stop the loop task and revoke the access token when the cog is unloaded."""


bot/exts/utilities/reddit.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, bot: Bot):
         self.access_token = None
         self.client_auth = BasicAuth(RedditConfig.client_id.get_secret_value(), RedditConfig.secret.get_secret_value())
 
-        self.auto_poster_loop.start()
+        # self.auto_poster_loop.start()
```
TizzySaurus, line 40: Since we're not starting the loop, do we also need to comment out the `self.auto_poster_loop.cancel()` call in `cog_unload()` (and anywhere else)?

ChrisLovering: No, cancel doesn't error if the task isn't running.

TizzySaurus (approved):

vivekashok1221 (approved):

PR #1302, opened 2023-06-20
Use the same pride facts for each year

ChrisLovering:
The command args have been changed to accept a number between 1 & 20, rather than a date, to make it easier to select which day of the week you want. This also updates the output of the command so that the selected day, or random is included in the title.


Xithrius (approved):
I see no issues in this implementation. Very cool, thank you Chris.

bot/exts/holidays/pride/pride_facts.py
```diff
@@ -28,65 +27,45 @@ def __init__(self, bot: Bot):
...
+          - If a option is a number (1-30), the fact for that given day of June is returned.
         """
         if not option:
-            await self.send_select_fact(ctx, datetime.now(tz=UTC))
+            await self.send_select_fact(ctx, datetime.now(tz=UTC).day)
         elif option.lower().startswith("rand"):
-            await self.send_random_fact(ctx)
-        else:
+            await ctx.send(embed=self.get_fact_embed())
+        elif isinstance(option, int):
```
wookie184, line 53: This needs to be before the previous check, otherwise `.lower()` will be called on the integer and cause an error.

wookie184 (changes requested):

bot/exts/holidays/pride/pride_facts.py
ChrisLovering: Good point, fixed in [forcepush](https://github.com/python-discord/sir-lancebot/compare/b0dc052e9a426511296f83b9f12871fbd4bd4ac4..ab551fa4efbf1001b4e5a14326577dc7c4ee962e)

wookie184 (approved):

PR #1321, opened 2023-07-06
Update to pydantic 2.0

wookie184:
Currently this doesn't work due to https://github.com/pydantic/pydantic-settings/issues/106. A PR fixing that was merged so just waiting for a release.

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -2,7 +2,8 @@ import enum
 from os import environ
 from types import MappingProxyType
 
-from pydantic import BaseSettings, SecretStr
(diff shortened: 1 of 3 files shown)


wookie184:
I'll also change it to use the class kwarg to set the `env_prefix` now that will be supported, since it looks pretty neat https://github.com/pydantic/pydantic-settings/issues/110

ChrisLovering (approved):
Looks good, one thing that we _could_ change, but my approval stands either way

Xithrius (approved):

PR #1343, opened 2023-08-10
Add a `.rfc` utility command

### LINKED ISSUE #1342: Add a `.rfc` command
## Description

Add a `.rfc` command with an RFC ID parameter that would respond with some information regarding the relevant RFC.

## Reasoning

This command would be useful in a situation where a relevant RFC would help contribute to the current conversation.

## Proposed Implementation

A possible implementation of this would get the plaintext representation of the RFC, from `https://www.rfc-editor.org/rfc/rfc{id}.txt`, and some basic information about the RFC, such as the category, name, and first few sentences of the linked RFC. An example implementation of this is shown in the below picture.
Screenshot 2023-08-07 at 12 43 57

SomeHybrid:
## Relevant Issues

#1342 

Closes #1342 

## Description

Added a `.rfc` utility command by using the JSON interface for RFCs at `https://datatracker.ietf.org/doc/rfc{id}/doc.json`, where `{id}` is a placeholder for the RFC ID.

## Did you:

### DIFF
--- a/bot/exts/utilities/rfc.py
+++ b/bot/exts/utilities/rfc.py
@@ -1,0 +1,65 @@
+import logging
+import datetime
+
+from discord.ext import commands
+from discord import Embed
+
+from bot.bot import Bot
+from bot.constants import Colours
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://datatracker.ietf.org/doc/rfc{query}/doc.json"
+
+
+class Rfc(commands.Cog):
+    """Retrieve RFCs from their ID."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @commands.cooldown(1, 10, commands.BucketType.user)
+    @commands.command()
+    async def rfc(self, ctx: commands.Context, query: int) -> None:
+        """Sends the corresponding RFC with the given ID."""
+        async with self.bot.http_session.get(BASE_URL.format(query=query)) as resp:
+            if resp.status != 200:
+                error = Embed(
+                    title="RFC not found",
+                    description=f"RFC {query} was not found.",
+                    colour=Colours.soft_red,
+                )
+
+                await ctx.send(embed=error)
+                return
+
+            data = await resp.json()
+
+            description = (
+                data["abstract"]
+                or f"[Link](https://datatracker.ietf.org/doc/rfc{query})"
+            )
+            title = data["title"]
+
+            embed = Embed(
+                title=f"RFC {query} - {title}",
+                description=description,
+                colour=Colours.gold,
+            )
+
+            embed.add_field(
+                name="Current Revision",
+                value=data["rev"] or len(data["rev_history"]),
+            )
+
+            created = data["rev_history"][0]["published"]
+            created = datetime.datetime.strptime(created, "%Y-%m-%dT%H:%M:%S%z")
+
+            embed.add_field(name="Created", value=created.strftime("%Y-%m-%d"))
+
+            await ctx.send(embed=embed)
+
+
+async def setup(bot: Bot):
+    """Load the Rfc cog."""
+    await bot.add_cog(Rfc(bot))
--- a/bot/exts/utilities/rfc.py
+++ b/bot/exts/utilities/rfc.py
@@ -1,6 +1,7 @@
 import datetime
 import logging
 
+import pydantic
 from discord import Embed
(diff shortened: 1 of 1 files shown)


bot/exts/utilities/rfc.py
```diff
@@ -0,0 +1,65 @@
...
+from bot.bot import Bot
+from bot.constants import Colours
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://datatracker.ietf.org/doc/rfc{query}/doc.json"
+
+
+class Rfc(commands.Cog):
+    """Retrieve RFCs from their ID."""
```
shtlrs, line 16:
```suggestion
    """Retrieves RFCs by their ID."""
```

bot/exts/utilities/rfc.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    @commands.cooldown(1, 10, commands.BucketType.user)
+    @commands.command()
+    async def rfc(self, ctx: commands.Context, query: int) -> None:
```
shtlrs, line 23:
```suggestion
    async def rfc(self, ctx: commands.Context, rfc_id: int) -> None:
```
This will keep it more explicit, instead of guessing what a query is, or reading down to understand that a query is the id of an rfc

The API offers to search by document name, rfc id, etc, but we're only allowing search by id (maybe it could be a good idea to search by name ?), that's why I made the suggestion.

bot/exts/utilities/rfc.py
```diff
@@ -0,0 +1,65 @@
...
+
+            data = await resp.json()
+
+            description = (
+                data["abstract"]
+                or f"[Link](https://datatracker.ietf.org/doc/rfc{query})"
+            )
+            title = data["title"]
+
+            embed = Embed(
```
shtlrs, line 44: You could factor this out into its own method (that remains private to the cog) which does the formatting of the data.

shtlrs (changes requested):
I didn't test this yet, I wast just passing by :P

It looks good overall, thanks.

It would be better to cache the data of an RFC into a dict, so that people who make the same request again would hit the cache instead of sending out the request.

You could make a simple pydantic model if you'd like to give your needed data a better structure, but it's not necessary.

SomeHybrid:
> It would be better to cache the data of an RFC into a dict, so that people who make the same request again would hit the cache instead of sending out the request.
> 
> You could make a simple pydantic model if you'd like to give your needed data a better structure, but it's not necessary.

Should the cache just store everything until the bot restarts, or have a mechanism of "flushing" the queue?

shtlrs:
I'd say just store in memory until the bot restarts as I don't think this command will have a high throughput. 

It's better that we only store the data we need instead of the entire response's body, which is why I suggested grouping them in a pydantic model.

Adding some logs would help us collect data just in case.

SomeHybrid:
> Adding some logs would help us collect data just in case.

Should the logs be sent every time a new item is added to the cache? or when it hits a certain point?

shtlrs:
I'd say just when fetching an RFC.

So something like

```py

log.info(''Fetching RFC %s'', rfc_id)
```

This will allow us to see how much it is used. I don't think it'd matter to find out which rfcs are queried the most, and even if it would, a simple log parsing would do.

bot/exts/utilities/rfc.py
```diff
@@ -0,0 +1,102 @@
...
+            return self.cache[rfc_id]
+
+        async with self.bot.http_session.get(BASE_URL.format(rfc_id=rfc_id)) as resp:
+            if resp.status != 200:
+                return None
+
+            data = await resp.json()
+
+        description = (
+            data["abstract"] or f"[Link](https://datatracker.ietf.org/doc/rfc{rfc_id})"
```
wookie184, line 44: I think we should link to the rfc even if the abstract is included, in case someone wants to read further.

wookie184 (changes requested):
This looks great, I have one suggestion.

SomeHybrid:
Demo screenshot here :)
Screenshot 2023-08-18 at 20 51 40

wookie184 (changes requested):

wookie184 (approved):
Looks good, thanks!

shtlrs (approved):
Looks good & works well, thank you !

PR #1427, opened 2024-01-09
Correctly type revisions as an int

ChrisLovering:
Solves #1426 Soves SIR-LANCEBOT-BA

This was causing a pydantic validation error, was it was being passed a string.

The only place this is used internaly casts it to a string anyway, so no further changes were needed.

## Did you:

- [ ] Join the [**Python Discord Community**](https://discord.gg/python)?
- [ ] Read all the comments in this template?
- [ ] Ensure there is an issue open, or link relevant discord discussions?
- [ ] Read and agree to the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

### DIFF
--- a/bot/exts/utilities/rfc.py
+++ b/bot/exts/utilities/rfc.py
@@ -20,7 +20,7 @@ class RfcDocument(pydantic.BaseModel):
 
     title: str
     description: str
-    revisions: str
+    revisions: int
     created: datetime.datetime
 
 


shtlrs (approved):

hedyhli (approved):
I thought it would send an error embed on occasions like this rather than no output at all... I probably misremembered.

PR #1444, opened 2024-02-01
Use paginator from botcore

shtlrs:
The `LinePaginator` has been migrated to `pydis_core`

The changes here are simply to reuse the paginator from the core lib.


hedyhli (approved):
Wasn't able to test `ImagePaginator`, but otherwise LGTM 👍

PR #1458, opened 2024-02-18
Use botcore command error manager

shtlrs:
This deflates the `CommandErrorHandler` from all the conditionals in place when it comes to error types.
Most importantly, it allows us to centralize error handling of both `text` and `slash` commands in one place, and reuse any necessary logic.

### DIFF
--- a/bot/command_error_handlers/check_failure.py
+++ b/bot/command_error_handlers/check_failure.py
@@ -1,0 +1,40 @@
+from typing import NoReturn
+
+from discord import Embed, Interaction
+from discord.ext.commands import CheckFailure, Context, NoPrivateMessage
+from pydis_core.utils.error_handling.commands import AbstractCommandErrorHandler
+
+from bot.constants import Channels, NEGATIVE_REPLIES
+from bot.utils.decorators import InChannelCheckFailure, InMonthCheckFailure
+
+from ._utils import create_error_embed
+
+
+class CheckFailureErrorHandler(AbstractCommandErrorHandler):
+    """An handler for the CheckFailure error."""
+
+    async def should_handle_error(self, error: Exception) -> bool:
+        """A predicate that determines whether the error should be handled or not."""
+        return isinstance(error, CheckFailure)
+
+    async def handle_text_command_error(self, context: Context, error: Exception) -> NoReturn:
+        """Handle error raised in the context of text commands."""
+        error_embed = self._get_error_embed(error)
+        await context.send(embed=error_embed, delete_after=7.5)
+        return
+
+    async def handle_app_command_error(self, interaction: Interaction, error: Exception) -> NoReturn:
+        """Handle error raised in the context of app commands."""
+        await interaction.response.send_message(embed=self._get_error_embed(error))
+
+    @staticmethod
+    def _get_error_embed(error: Exception) -> Embed:
+        if isinstance(error, InChannelCheckFailure | InMonthCheckFailure):
+            return create_error_embed(str(error), NEGATIVE_REPLIES)
+        if isinstance(error, NoPrivateMessage):
+            return create_error_embed(
+                "This command can only be used in the server. "
+                f"Go to <#{Channels.sir_lancebot_playground}> instead!",
+                NEGATIVE_REPLIES
+            )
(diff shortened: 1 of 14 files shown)


shtlrs:
I'll squash the implementation commits along with the fixup one once the PR is ready to be merged.

wookie184:
Honestly, I don't understand the benefit of this error dispatching system. I feel like this introduces extra complexity and makes the logic much harder to follow and understand.

Before, following the logic was as simple as:
1) `on_command_error` is called
2) `isinstance` is used to dispatch the error to the necessary handling logic

Now, to follow the logic it's:
1) `on_command_error` is called
2) It gets `self.bot.command_error_manager`
    
    - ...which is set by `register_command_error_manager` in bot-core
    - ...which is called by the return value of `bootstrap_command_error_manager` in `bot/__main__.py`
    - ...which is defined in `bot/command_error_handlers/__init__.py`, and constructs the `CommandErrorManager`
3) It calls `handle_error` on `self.bot.command_error_manager`.
    
    - This runs through `self._handlers` (which was set by calls to `register_handler` back in `bot/command_error_handlers/__init__.py`) and calls `should_handle_error` on each handler
    - This actual `should_handle_error` function is split over a lot of files.
    - `isinstance` is used to determine whether the function should be run.
    - `_get_callback` is called and the correct function is selected
    - The function is called.

I'm not sure what benefit these extra steps have given us over the current code.

shtlrs:
The idea initially came to make the exception handling logic reusable for both prefix command & slash command errors.

For the context, we have some prefix/text commands that also have their slash version sibling, and these commands use the same logic/throw the same errors.

The question was (and I'd be interested to have your point of view), how do we reuse all the logic that existed in the error handling cog in case we have errors that came from these slash commands ? 

Discord py dispatches these prefix commands through the `on_command_error` callback, and the ones coming from slash commands through the `CommandTree`'s `on_error` callback, which both have different signatures since one relies on `Context` and the other relies on `Interaction`.  So trying to relay errors from the `on_error` to the `on_command_error` one needs some sort of conversion.

I just thought of unifying both under the same interface and hiding the conversion details for the same type of error.

I understand your point of view and agree that it's slightly more complex than the setup we currently have but I don't necessarily agree that it's makes it much harder to follow nor that complex, but that could just be the fact that I'm biased since I'm the one who wrote, which makes the understanding easier.

bot/command_error_handlers/check_failure.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class CheckFailureErrorHandler(AbstractCommandErrorHandler):
+    """An handler for the CheckFailure error."""
+
+    async def should_handle_error(self, error: Exception) -> bool:
+        """A predicate that determines whether the error should be handled or not."""
+        return isinstance(error, CheckFailure)
+
+    async def handle_text_command_error(self, context: Context, error: Exception) -> NoReturn:
```
MarkKoz, line 20: The default value of a function is `None`. If a function can return, but has no explicit return value, then this annotation should be `None`. 

`NoReturn`  means it does not return i.e. control flow of the program is interrupted before it returns. I've only seen this used for functions that always throw an exception, like an abstract method that raises NotImplemented. Also, this was superceded (?) in 3.11 by `Never`.

bot/command_error_handlers/command_not_found.py
```diff
@@ -0,0 +1,64 @@
...
+QUESTION_MARK_ICON = "https://cdn.discordapp.com/emojis/512367613339369475.png"
+
+
+class CommandNotFoundErrorHandler(AbstractCommandErrorHandler):
+    """A handler for all CommandNotFound exceptions."""
+
+    def __init__(self, bot: Bot):
+        self.bot = bot
+
+    async def should_handle_error(self, error: errors.DiscordException) -> bool:
```
MarkKoz, line 23: Thoughts about using `@typing.override` for these? It makes it more explicit that this is an override, and it helps type checkers. But if we don't use a type checker then it could be argued that this would just clutter the code.

MarkKoz (commented):
A few preliminary comments - still need to do a deeper review.

shtlrs:
Closing this as we haven't reached a consensus within the core dev team.

PR #1500, opened 2024-04-02
Trivia: Fix misleading category description embed

hedyhli:
The current formatting for the `.quiz help` command is misleading, where the spaces causes the wrong categories descriptions to be matched.

imageimage


wookie184 (approved):
Looks good, though do you think it would be neater to have the description on the same line as the category name, since most of the category descriptions are pretty short? I've not tried it so i'm not sure though.

hedyhli:
I experimented with that format but decided on this one because that way makes the list quite spaced out. This way is easier to scan from top to bottom.

I can definitely change it if there's a very good reason you see why putting it on the same line is better.

hedyhli:
The other two formats, for reference:
image
image

Xithrius:
I like the second one more

hedyhli:
I have made it follow the more concise format (second screenshot) as per request from @Xithrius.

hedyhli:
Diff of the force-pushed changes.
```diff
diff --git a/bot/exts/fun/trivia_quiz.py b/bot/exts/fun/trivia_quiz.py
index 6041d031..d624e0fe 100644
--- a/bot/exts/fun/trivia_quiz.py
+++ b/bot/exts/fun/trivia_quiz.py
@@ -625,7 +625,7 @@ class TriviaQuiz(commands.Cog):
 
         for cat, description in self.categories.items():
             embed.description += (
-                f"- **{cat.capitalize()}**\n{description.capitalize()}\n\n"
+                f"- **{cat.capitalize()}**: {description.capitalize()}\n"
             )
 
         return embed
```

wookie184 (approved):
👍

Xithrius (approved):
Thank

PR #1505, opened 2024-04-02
Trivia quiz: Fix question about surrogate codepoints

### LINKED ISSUE #1501: Trivia question 423 is wrong
## Description

Trivia question 423 is "What is the name for unicode codepoints that do not fit into 16 bits?", with the supposedly correct answer being "surrogates". This is wrong.

At some point in the past what is now UTF-16 was occasionally called "Unicode", especially in the Windows and Java world. To encode codepoints outside the [Basic Multilingual Plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) UTF-16 uses surrogate characters, which are part of Unicode, but are _below_ 16 bits (otherwise they couldn't be used in UTF-16).

## Steps to Reproduce
Play `.quiz cs` for a long time until you get this question, then get confused.

## Possible Solutions
Remove or drastically rephrase this question.

(either is fine with me)

L3viathan:
Closes #1501.

I decided to just rephrase the question, but I can also scrap it. I just didn't feel like renumbering a bajillion IDs.

### DIFF
--- a/bot/resources/fun/trivia_quiz.json
+++ b/bot/resources/fun/trivia_quiz.json
@@ -723,8 +723,8 @@
     },
     {
       "id": 423,
-      "question": "What is the name for unicode codepoints that do not fit into 16 bits?",
-      "answer": "surrogates"
+      "question": "What kind of characters are needed in UTF-16 to encode unicode codepoints outside the BMP?",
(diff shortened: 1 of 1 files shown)


wookie184 (approved):
The phrasing looks good, thanks

PR #1626, opened 2024-11-09
Typst command implementation

### LINKED ISSUE #1623: Add Typst support
## Description
Add a `.typst` command, working similarly to `.latex` but for rendering [Typst](<https://github.com/typst/typst>).

## Reasoning
The bot's `.latex` command has been notoriously unstable in the past, but nowadays is mostly working. However, it's pretty hard to consistently write correct LaTeX snippets on the first try, and the error logs are not especially helpful because LaTeX logs rarely are. It would be nice to have a simpler alternative for quickly writing some simple math, and Typst would fill that niche - Typst source code is much more readable and doesn't require an explicit preamble. The following is an (entire!) typst document:
```typst
$lim_(h -> 0) (f(x + h) - f(x))/h$
```
which renders to:
Image

## Proposed Implementation
Typst has a [Python library](<https://pypi.org/project/typst/>), so the implementation should be straightforward. Here's a proof of concept:
```py
import typst
from tempfile import NamedTemporaryFile
from PIL import Image

s = """$lim_(h -> 0) (f(x + h) - f(x))/h$"""
with NamedTemporaryFile("w", encoding="utf-8") as inp, NamedTemporaryFile("rb", suffix=".png") as out:
    inp.write(s)
    inp.flush()
    typst.compile(inp.name, out.name)
    img = Image.open(out)
    img.load()
img.show()
```
An interesting detail is the handling of third-party packages. The Latex cog is currently implemented as having [a specific set of whitelisted packages](<https://github.com/python-discord/sir-lancebot/blob/1cf62552f77955c89fd9dd1631e0efa6391ad74c/bot/resources/fun/latex_template.txt#L2>) available. Typst, meanwhile, by default supports importing arbitrary packages at runtime to be downloaded during the compilation process - which in theory is safe because typst plugins are [required to be pure functions that get compiled to WASM](<https://typst.app/docs/reference/foundations/plugin>). Nevertheless, probably at first it'd be best to only allow a shortlist of popular packages, but in the future we could consider removing that limit.

RundownRhino:
## Relevant Issues
Closes #1623.

## Description
- The overall structure is based on the Latex cog, with about the same features (caching, support for code blocks, etc)
- Refactored some utility functions (such as the codeblock regex) out of the Latex cog to reuse them here.
- Initially I was going to use the `typst` PyPI package, but while it works, it had a weird issue where (only in the bot's container and not on my computer) it needed ~1GB RAM to render the simplest things. I failed to figure out what caused this, and as a result I switched to using the CLI, which is less hungry (~100MB for sane inputs) and supports some nice features the python bindings don't (such as taking the input from stdin with no files involved).
- The typst executable is not included, but instead downloaded (with hash verification) at cog load time, from a link that can be altered with envvars, and then stored in the cache.
- If the global typst packages directory is missing, a default set of packages is installed and the directory then `chmod`ed to not allow writes.
- To render, the `typst` executable gets called in a subprocess, with a time limit (and gets killed if the limit is exceeded) and a memory usage rlimit.
- If the rendered image is too big, this is considered an error (to hopefully make it a bit harder to maliciously post big outputs).
- The rendered image also gets manually cropped to only the content and not the white background (typst's own formatting doesn't seem to have an option for it, and this is the more resilient choice anyway).

Security-wise, the typst invocation sets `--root` to an empty temporary directory, which in theory should prevent access to any files outside of it. As for the packages, even though arbitrary typst packages are supposed to be safe, I still chose to prevent installation of them at runtime by locking the packages directory from writes. Nevertheless, I am pretty worried about the security of this and want someone to take a second look at whether it's exploitable. And of course, if there's ever a new vulnerability in `typst` itself, that'd potentially make Lancebot's environment vulnerable.

## Did you:

### DIFF
--- a/bot/exts/fun/typst.py
+++ b/bot/exts/fun/typst.py
@@ -4,6 +4,7 @@ import string
 import sys
 from io import BytesIO
 from pathlib import Path
+from subprocess import CalledProcessError
 from tempfile import TemporaryDirectory
 
 import discord
@@ -112,7 +113,9 @@ class Typst(commands.Cog):
             stderr=asyncio.subprocess.PIPE,
         )
         stdout, _ = await proc.communicate()
-        log.info(f"Typst executable reports itself as {stdout.decode('utf-8').strip()!r}.")
+        log.info(
+            f"Typst executable reports itself as {stdout.decode('utf-8').strip()!r}."
+        )
 
     async def _download_typst_executable(self) -> None:
         if not Config.typst_archive_url:
@@ -200,7 +203,6 @@ class Typst(commands.Cog):
                     )
                     return
                 except TypstWorkerCrashedError:
-                    log.exception("typst worker crashed")
                     await ctx.send(
                         "Worker process crashed. "
                         f"Perhaps the memory limit of {TYPST_MEMORY_LIMIT/1024**2:.1f}MB was exceeded?"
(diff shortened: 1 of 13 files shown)


bot/exts/fun/typst.py
```diff
@@ -0,0 +1,309 @@
...
+            )
+        log.info("Retrieved Typst archive, unpacking")
+        typst_executable = archive_retrieve_file(
+            arc_data,
+            filename="typst.exe" if sys.platform == "win32" else "typst",
+        )
+        log.info("Storing the typst executable on disk")
+        exe_path = Path(Config.typst_path)
+        exe_path.parent.mkdir(exist_ok=True)
+        exe_path.write_bytes(typst_executable)
```
ChrisLovering, line 143: This isn't going to work in our infrastructure. We use read-only file systems.

I am also very uneasy with the fact we're downloading and executing a binary from the internet like this into the bot's pod.

Is there a typst API similar to what we use for latex? Either that, or we can look into having a typst deployment in our infra, where it won't have access to the bot's secrets.

PR #1656, opened 2025-03-07
feat: add .yoda command

marcuserlandsson:
## Relevant Issues
Closes #494

## Description
Implemented the .yoda command, which converts input text to Yoda-like speech according to the description in issue #494. Also includes unit tests to verify the functionality of the file.

## Did you:

- [x] Join the [**Python Discord Community**](https://discord.gg/python)?
- [x] Read all the comments in this template?
- [x] Ensure there is an issue open, or link relevant discord discussions?
- [x] Read and agree to the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

PR #1657, opened 2025-03-07
Bemyvalentine patch

### LINKED ISSUE #1477: Option to send public valentines privately through DMs
## Description

The `bemyvalentine` feature allows for valentines to either be secret/anonymous (sends the message to the selected user through DMs, without mentioning who it's from), or public by default (sends the message to `#sir-lancebot-playground` for everyone to see).

I would like for the non-secret mode to send valentines _privately_ to the selected user in DMs, just like the secret mode.

The only difference would be that the valentine embed will include who the message is from.

Perhaps this can be made into an option for the user to choose whether to send it through DMs or to a public channel.

## Proposed Implementation

So for the "normal" `.bemyvalentine` command, these two options:

1. **Public**
Send the valentine message embed **into the same channel as the command is invoked in**. Currently, it forwards all valentine messages to `#sir-lancebot-playground` instead, which is away from where the `.bemyvalentine` command is invoked. I believe that if the embed is sent directly in that current channel conversation (whether that be serious or not) can remain in the same place, rather than having the user on the receiving end check for a ping in some other channel and refer back to where `.bemyvalentine` is invoked.

2. **Private**
Send the message **through DMs** to the user. Same as `.bemyvalentine secret`, but reveals who the message is from in the embed. This allows people to make use of Sir Lancebot to send compliments/letters to another member without being anonymous, whilst having the whole ordeal remain private.

## Reasoning
I love to send compliments to other members during the month of February as valentines. The person I send it to should want to know that it was I who sent it. (Actually, I just realized that if they're a mod, they can see the `.bemyvalentine secret` message Sir Lancebot deletes, right? right?! Ah, but I degress.)

Now sometimes, it isn't ideal for everyone to see those valentines messages in `#sir-lancebot-playground` if I choose to have the matter remain private among me and the person on the receiving end. Hence, it would be amazing if we implemented such an option for the user to choose whether the message would be sent publicly or private through DMs, for nonanonymous valentines.

---
*Anyone can implement this feature*

willeStjerna:
## Relevant Issues
Closes #1477 

## Description
Implements the desired changed to .bemyvalentine. Also adds the feature to let the sender choose if they want to send it anonymously or not.

## Did you:
- [X ] Read and agree to the [contributing guidelines](https://pythondiscord.com/pages/contributing/contributing-guidelines/)?

### DIFF
--- a/bot/constants.py
+++ b/bot/constants.py
@@ -47,6 +47,7 @@ class EnvConfig(
 
 
 class _Channels(EnvConfig, env_prefix="channels_"):
+    general: int = 123123123123
     algos_and_data_structs: int = 650401909852864553
     bot_commands: int = 267659945086812160
     community_meta: int = 267659945086812160
@@ -312,6 +313,7 @@ STAFF_ROLES = {Roles.helpers, Roles.moderation_team, Roles.admins, Roles.owners}
 
 # Whitelisted channels
 WHITELISTED_CHANNELS = (
+    Channels.general,
     Channels.bot_commands,
     Channels.sir_lancebot_playground,
     Channels.off_topic_0,
--- a/bot/exts/holidays/valentines/be_my_valentine.py
+++ b/bot/exts/holidays/valentines/be_my_valentine.py
@@ -92,7 +92,7 @@ class BeMyValentine(commands.Cog):
         # COMMENTED FOR TESTING PURPOSES
 
         if user == ctx.author:
-        #     # Well a user can't valentine himself/herself.
+        # Well a user can't valentine himself/herself.
             raise commands.UserInputError("Come on, you can't send a valentine to yourself :expressionless:")
 
         emoji_1, emoji_2 = self.random_emoji()
--- a/bot/exts/holidays/valentines/be_my_valentine.py
+++ b/bot/exts/holidays/valentines/be_my_valentine.py
@@ -89,10 +89,9 @@ class BeMyValentine(commands.Cog):
                 f"Specify if you want the message to be sent privately or publicly!"
             )
         
-        # COMMENTED FOR TESTING PURPOSES
 
         if user == ctx.author:
-        # Well a user can't valentine himself/herself.
+            # Well a user can't valentine himself/herself.
             raise commands.UserInputError("Come on, you can't send a valentine to yourself :expressionless:")
 
         emoji_1, emoji_2 = self.random_emoji()
--- a/bot/exts/holidays/valentines/be_my_valentine.py
+++ b/bot/exts/holidays/valentines/be_my_valentine.py
@@ -44,7 +44,7 @@ class BeMyValentine(commands.Cog):
     # @commands.cooldown(1, 1800, commands.BucketType.user)
     @commands.group(name="bemyvalentine", invoke_without_command=True)
     async def send_valentine(
-        self, ctx: commands.Context, user: discord.Member, privacy_type: str | None = None, valentine_type: str | None = None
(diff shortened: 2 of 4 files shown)


bot/exts/holidays/valentines/be_my_valentine.py
```diff
@@ -41,21 +41,39 @@ async def lovefest_role(self, ctx: commands.Context) -> None:
         """
         raise MovedCommandError(MOVED_COMMAND)
 
-    @commands.cooldown(1, 1800, commands.BucketType.user)
+    # @commands.cooldown(1, 1800, commands.BucketType.user)
     @commands.group(name="bemyvalentine", invoke_without_command=True)
     async def send_valentine(
-        self, ctx: commands.Context, user: discord.Member, *, valentine_type: str | None = None
+        self, ctx: commands.Context, user: discord.Member, privacy_type: str | None = None, anon: str | None = None, valentine_type: str | None = None
```
wookie184, line 47: `typying.Literal` can be used to make discord.py check if the argument is valid https://discordpy.readthedocs.io/en/latest/ext/commands/commands.html#typing-literal

This would simplify the logic later on.

bot/constants.py
```diff
@@ -47,6 +47,7 @@ class EnvConfig(
 
 
 class _Channels(EnvConfig, env_prefix="channels_"):
+    general: int = 123123123123
```
wookie184, line 50: Anything changes made just for testing should be reverted (though I'm not really sure why this would be needed, when you can just use a channel that is already configured as whitelisted)

bot/exts/holidays/valentines/be_my_valentine.py
```diff
@@ -41,21 +41,39 @@ async def lovefest_role(self, ctx: commands.Context) -> None:
         """
         raise MovedCommandError(MOVED_COMMAND)
 
-    @commands.cooldown(1, 1800, commands.BucketType.user)
+    # @commands.cooldown(1, 1800, commands.BucketType.user)
```
wookie184, line 44: I assume this was done while testing, but should be reverted now. To keep git history clean, avoid committing temporary changes.

bot/exts/holidays/valentines/be_my_valentine.py
```diff
@@ -73,12 +98,33 @@ async def send_valentine(
...
+                description=f"{valentine} \n **{emoji_2}From an anonymous admirer{emoji_1}**",
+                color=Colours.pink
+            )
+        
+        else:
+            embed = discord.Embed(
+                title=f"{emoji_1} {title} {user.display_name} {emoji_2}",
+                description=f"{valentine} \n **{emoji_2}From {ctx.author}{emoji_1}**",
+                color=Colours.pink
+            )
```
wookie184, line 113: There is unnecessary duplication here, you could just have something like
```python
from_user = "an anonymous admirer" if anon.lower() == "anon" else ctx.author
```
and format that into a single embed call.

tests/exts/holidays/valentines/test_bemyvalentine.py
```diff
@@ -0,0 +1,323 @@
+
```
wookie184, line 1: See my comment on another PR - https://github.com/python-discord/sir-lancebot/pull/1656#discussion_r2017828822

We don't have existing tests on this repo so I'm not sure we want these.

wookie184 (changes requested):
Thanks for your contribution.

I have not tested or looked into the functionality yet as I notice that CI is failing, but I have a few general comments.

PR #1688, opened 2025-08-03
Madlibs - change zip parameter "strict" to False

DMFriends:
## Relevant Issues
Link to discussion: https://discord.com/channels/267624335836053506/635950537262759947/1401476352758251520

## Description

Changed one line of the Madlibs code which was causing a "ValueError: zip() argument 2 is shorter than argument 1" and, as a result, not displaying the final Madlib. All I had to do is change the "strict" parameter in the zip() function call to False and that fixed the issue.

## Did you:


hedyhli (approved):
Thanks!

PR #1743, opened 2026-04-16
Added additional trivia questions

### LINKED ISSUE #1491: Add more trivia questions
## Description

Add more trivia questions.

Using a format i can copy and paste into a python script to generate the json.

`list[tuple[str, list[str], Optional[str]]] = [(question, [answer1, answer2], hint)]`
### Python
```py
[
("What common operation does this function implement?\n```py\nmystery = lambda *args: map(lambda *x: x, *args)\n```", ["zip"], None),
("What common operation does this function implement?\n```py\ndef f(a):\n    b = []\n    for b[len(b):] in a:3\n    return b\n```", ["flatten"], None),
("What algorithm does this function implement?\n```py\nmystery = lambda xs: xs and [*mystery([x for x in xs[1:] if x <= xs[0]]), xs[0], *mystery([x for x in xs[1:] if x > xs[0]])]\n```", ["quicksort"], "It is a sorting algorithm."),
("In what year did Python 2 reach End of Life?", ["2020"], None),
]
```
### CS
A question I thought would be fun is generating a random directed graph and asking people to find the shortest path between two randomly selected nodes. Variations could be solving TSP, and so on. Would need to send images though, which would require some modifications to a lot of code. Additionally, we would need to bring in another dependency like networkx or graphviz. Maybe this would be turned into another issue/PR? I'm not sure what the protocol for that is, would like thoughts on that from maintainer people.

Another question I thought would be interesting is: "give a string matching this regular expression". And randomly generate regexes.

```py
[
("What lemma is commonly used to prove non-regularity of languages?", ["pumping lemma", "pumping"], None),
("TCP congestion and flow control throttle the sender by adjusting the?", ["window size"], None),
("What algorithm was famously created at a cafe in 20 minutes?", ["dijkstra"], None),
("What is the minimal time complexity for a comparison-based sorting algorithm?", ["O(n log n)", "O(nlogn)", "n log n", "nlogn"], None),
("How many valid code points are in the Unicode codespace?", ["1112064", "2^16-2^11+2^20"], None),
("Deterministic finite automata are equivalent to nondeterministic finite automata (true/false).", ["true"], None),
("Deterministic pushdown automata are equivalent to context free grammars (true/false).", ["false"], None),
("The family of algorithms for choosing a simple random sample without replacement from a stream of unknown size in a single pass is called?", ["reservoir sampling"], None),
("A hash function with no hash collisions is known as?", ["perfect", "perfect hash function"], None),
]
```

## Reasoning

More trivia questions good.

## Proposed Implementation
Modify the json file for static questions. For the dynamic question we would implement this as a function that generates a graph and associated image.

asterized:
## Relevant Issues

Closes #1491 

## Description

Added additional questions to `bot/resources/fun/trivia_quiz.json`. Some questions were omitted from the original issue. 

## Did you:

### DIFF
--- a/bot/resources/fun/trivia_quiz.json
+++ b/bot/resources/fun/trivia_quiz.json
@@ -830,6 +830,73 @@
       "id": 444,
       "question": "On Linux systems, the `fork` system call returns what value in the parent process?",
       "answer": ["PID", "child PID"]
+    },
+    {
+      "id": 445,
+      "question": "What lemma is commonly used to prove non-regularity of languages?",
+      "answer": ["pumping lemma", "pumping"]
+    },
+    {
+      "id": 446,
+      "question": "TCP congestion and flow control throttle the sender by adjusting the _________",
+      "answer": "window size"
+    },
+    {
+      "id": 447,
+      "question": "What algorithm was famously created at a cafe in 20 minutes?",
+      "answer": ["Dijkstra's", "Dijkstra", "Dijkstra's algorithm"],
+      "info": "Dijkstra's algorithm is an algorithm to find the shortest path between two nodes in a weighted graph."
+    },
+    {
+      "id": 448,
+      "question": "What is the minimal time complexity for a comparison-based sorting algorithm?",
+      "answer": ["O(n log n)", "O(nlogn)", "nlogn", "n log n"]
+    },
+    {
+      "id": 450,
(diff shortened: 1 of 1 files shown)


L3viathan (approved):
LGTM

jchristgit (approved):

jchristgit:
Thanks!

